PR #3, opened 2020-09-30
Spotless

jplaisted:
## Checklist


.github/ISSUE_TEMPLATE/--bug-report.md
```diff
@@ -3,30 +3,26 @@ name: "\U0001F41EBug report"
...
 title: A short description of the bug
 labels: bug
-assignees: ''
-
+assignees: ""
 ---
 
-**Describe the bug**
-A clear and concise description of what the bug is.
+**Describe the bug** A clear and concise description of what the bug is.
```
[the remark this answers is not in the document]
mars-lan: Yeah it's a bit out of our control. Is it possible to tune the format so that it works in both cases?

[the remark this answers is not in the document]
jplaisted: I don't think so. This is probably GitHub's fault for having two different renderers, one of which is non standard.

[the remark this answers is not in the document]
jplaisted: The best solution is to probably just make these section headers instead. They don't have to be level one, we can use something like 3rd or 4th level headings `####` to make them a bit smaller.

# first
## second
### third
#### fourth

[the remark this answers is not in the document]
mars-lan: SGTM

.github/ISSUE_TEMPLATE/--bug-report.md
```diff
@@ -3,30 +3,32 @@ name: "\U0001F41EBug report"
 about: Create a report to help us improve
 title: A short description of the bug
 labels: bug
-assignees: ''
-
+assignees: ""
```
mars-lan, line 6: Is this a spotless change as well?

build.gradle
```diff
@@ -8,6 +8,7 @@ buildscript {
     classpath 'com.github.node-gradle:gradle-node-plugin:2.2.4'
     classpath 'com.commercehub.gradle.plugin:gradle-avro-plugin:0.8.1'
     classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.1.2.RELEASE'
+    classpath 'com.diffplug.spotless:spotless-plugin-gradle:5.6.1'
```
mars-lan, line 11: order plugins alphabetically?

docs/what/entity.md
```diff
@@ -67,48 +87,52 @@ record DatasetEntity includes BaseEntity {
...
-| 1 July | Cancer | Larkspur  
-| ... | ... |
-| 31 December | Capricorn | Poinsettia
+A lot of time it may not be obvious if something should be modeled as an entity, a [metadata aspect](aspect.md), or even
+a field in a metadata aspect. One way to think of it is using the concept of
+[Third Normal Form](https://en.wikipedia.org/wiki/Third_normal_form). We'll use the example from the Wikipedia entry to
+illustrate the idea.
 
-# How to delete an entity?
+**Tournament Table** | Tournament | Year | Winner | | ------- | --------- | --------- | | Indiana Invitational | 1998 |
```
mars-lan, line 95: This makes table super hard to read?

docs/rfc/templates/000-template.md
```diff
@@ -66,7 +66,7 @@
 
 ## Drawbacks
 
-> Why should we *not* do this? Please consider the impact on teaching DataHub, on the integration of this feature with
+> Why should we _not_ do this? Please consider the impact on teaching DataHub, on the integration of this feature with
```
mars-lan, line 69: Does spotless prefer italic over bold?

mars-lan (changes requested):

.github/ISSUE_TEMPLATE/--bug-report.md
jplaisted: Yes

mars-lan: How come it didn't touch the other single quotes below?

docs/rfc/templates/000-template.md
jplaisted: `*` and `_` are both italic, so it standardizes on `_` for consistency. `**` is bold.

*star*
_underscore_
**double star**

docs/what/entity.md
jplaisted: Same thing as before. This wasn't an actual heading and had no whitespace so spotless collapsed it. Fixed.

docs/rfc/templates/000-template.md
mars-lan: Got it.

docs/what/entity.md
```diff
@@ -67,48 +87,62 @@ record DatasetEntity includes BaseEntity {
...
 
-# How to delete an entity?
+In other words, when you start asking yourself "Should I normalize this thing so it doesn't end up repeated data that
+can potentially become inconsistent?", chances are that "thing" should probably be made an entity. For example, if we
+need to associate a specific birthday with additional facts, such as the corresponding star sign or
+[birth flower](https://en.wikipedia.org/wiki/Birth_flower), we'll end up introducing the `Birthday` table & entity and
+further denormalize the `Winner` table.
 
-We purposely made all [metadata aspects](aspect.md) immutable, i.e. each edit results in a new version created with no easy way to remove a specific version. However, since the existance of an entity is determined by the existance of its associated metadata aspects, it seems that there's no easy way to delete an entity. In fact, this is echoed by the fact that [GMS](gms.md) doesn't actually provide any `DELETE` API!
+**Winner Table** | Winner | Birthday | Birth year | | ------- | --------- | --------- | | Chip Masterson | 14 March |
```
mars-lan, line 123: More tables to fix?

jplaisted: Done, thanks. Verified these are the only tables by checking out master and running

```
find . -name '*.md' | xargs grep -l '^|' | grep -v build/
```

.github/ISSUE_TEMPLATE/--bug-report.md
jplaisted: Those are in actual markdown. I'm not sure what the `---` section is, but seems to be some metadata that spotless is aware of.

jplaisted: It's YAML

mars-lan: Interesting way to mix YAML inside of MD :)

docs/rfc/templates/000-template.md
mars-lan: Actually seems like underscore is discouraged? It also looks less natural than `*` IMHO: https://www.markdownguide.org/basic-syntax/#italic

jplaisted: Prettier is opinionated enough where I don't think they have an option for this.

https://prettier.io/docs/en/option-philosophy.html

Part of the point of prettier/spotless/_any formatter_ is to avoid this kind of debate, imo.

mars-lan: Seems like the wrong opinion to have given that `_` may not get rendered correctly in all cases. Never mind.

mars-lan (approved):

docs/rfc/templates/000-template.md
jplaisted: It seems like they target common mark?

https://github.com/prettier/prettier/issues/3728

Issue #4, opened 2020-10-01
feat: Start publishing jars


jplaisted:
We need to start publishing jars do DataHub can use them.

jplaisted:
Before we start doing this, if we don't touch the code here are the jars we'll make today:

- ebean-dao
- elasticsearch-dao
- li-utils
- metadata-dao
- metadata-restli-resource
- metadata-test-models
- metadata-test-utils
- metadata-utils
- metadata-validators
- neo4j-dao

@mars-lan thoughts on this? I'd propose the following changes:

1. Drop the "metadata" prefix from packages.
2. Rename `li-utils` to `urn`
3. Delete `metadata-test-utils` entirely.
    a. Neo4j and ES shouldn't be used in the same module to avoid the dependency. If we _need_ these, they can move to their own modules.
    b. `TestUtils` doesn't do much and can easily be embedded.
    c. `AuditStamps`, while nice also isn't doing a ton. I don't think it justifies the existence of a "utils" module. We could move it to `test-models`. However, this file may be more widely used by DataHub and within LinkedIn. At the very least maybe we can just drop the "utils" from the module name; it it'd just be `testing`.
4. Move `metadata-utils` back into datahub. Code is used mostly by jobs and can move back when the "job libraries" move.

So after that we'd be left with:

- ebean-dao
- elasticsearch-dao
- urn
- dao
- restli-resource
- test-models
- testing
- validators
- neo4j-dao

jplaisted:
Could also rename `metadata-dao` to `dao-api` instead.

mars-lan:
Do we really want to rename everything? That means a lot of changes internally as well.

jplaisted:
We already need internal changes when we switch from internal to external jars.

We don't have to rename anything, this is just a good time to do it if we want to :)

mars-lan:
Agree. Some of these can use a better name :)

jplaisted:
Any specific ideas or do I have carte blanche here? I'd rather discuss what we want upfront than in a PR.

mars-lan:
1. `dao` sounds a bit too generic and confusing. Maybe rename it to `dao-api` or something?
2. Maybe rename `restli-resource` to `restli-resources` so it's consistent with the others?
3. `li-utils` contains more than just URN, unless you plan to move non-URN code/models out of the module?
4. Agree that `metadata-utils` should be moved back to datahub and maybe split into `factories` and `elasticsearch-support` or something.

[...]

PR #12, opened 2020-10-06
Continously release versions.

jplaisted:
This is done via a custom GitHub workflow / action that uses shipkit github release and auto version plugins. This means we will publish a release on GitHub and bintray for every commit on a branch named `release/*`.

For now autopublish to bintray is still disabled, we can enable it once we're sure we're happy with this pipeline.

#4 
## Checklist


jplaisted:
Also, I acknowledge we should have some discussion here as to how we want to do development / releases.

Right now, this based on mockito. There is no "master" branch in this model. Instead, there are release branches (e.g. `release/0.1.x`). We can them make the highest version the default branch and work off that. We can keep old release branches around to port fixes back, if needed, or clean them up if we're no longer maintaining them.

docs/how/releases.md
```diff
@@ -0,0 +1,31 @@
+# Releases
+
+We create [GitHub releases](https://github.com/linkedin/datahub-gma/releases) and publish our artifacts to
+[Bintray](https://bintray.com/linkedin/maven/datahub-gma).
+
+We use shipkit to help create our releases and auto-increment our versions.
```
mars-lan, line 6: Link for shipkit?

docs/how/releases.md
```diff
@@ -0,0 +1,31 @@
...
+We automatically create releases for every commit on a branch that starts with `release/%semver%`, where semver is the
+wildcard semantic version of that branch (e.g. `release/1.0.x`).
+
+## Versioning
+
+As stated above, we use shipkit for automatically creating version tags. See
+[shipkit-auto-version for details](https://github.com/shipkit/shipkit-auto-version).
+
+We follow the [semantic versioning 2.0.0 convention](https://semver.org/). When bumping the major or minor version,
+please make a new `release` branch for it.
```
mars-lan, line 17: Understandable for major version, but creating a new branch for minor versions seems excessive?

docs/how/releases.md
jplaisted: We aren't using the shipkit plugin itself but these smaller ones that I've linked below. Clarified.

jplaisted: (there is a "shipkit" plugin)

docs/how/releases.md
jplaisted: /shrug

Up to you. I think it is important to remind ourselves that the minor version exists and we should be using it when we need to. Which is hopefully rare, but could happen. So I think in most cases we won't bump the minor anyway, so it is a bit moot. But when it does come up, might be nice to have?

jplaisted: Also while we're on `0` the minor version has more meaning and will actually be bumped more often.

jplaisted: Switched to trunk development. PTAL.

build.gradle
```diff
@@ -155,4 +141,51 @@ spotless {
     target '**/*.md'
     prettier().config([printWidth: 120, proseWrap: 'always'])
   }
+}
+
+task getBranchName() {
+  doLast {
+    new ByteArrayOutputStream().withStream { os ->
+      def result = exec {
+        commandLine 'git', 'branch', '--show-current'
```
mars-lan, line 150: Seems a bit hacky to me? Hard to believe there's no existing lib/plugin that does this?

docs/how/releases/continuous-releases.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Automatic Changelog
+
+We use shipkit changelog to create the release notes of each version. See
+[shipkit-changelog](https://github.com/shipkit/shipkit-changelog) for more details.
+
+## Automatic Publishing
+
+Automation is done with a [GitHub workflow](../../.github/workflows/gh-version.yml) that will create the GitHub release
+and then publish to bintray for every commit on a `release` branch.
```
mars-lan, line 32: main branch?

docs/how/releases/versioning.md
```diff
@@ -0,0 +1,13 @@
+# GMA Versioning
```
mars-lan, line 1: Seems redundant info?

mars-lan (approved):
Great work

PR #14, opened 2020-10-07
Refactor in preparation of publishing jars.

jplaisted:
- Delete some files that can stay in Datahub (Configuration, Pair).
- Ensure data template jars do not contain duplicate class definitions.
- Ensure we do publish data template jars.

#4 

## Checklist


test-models/build.gradle
```diff
@@ -7,4 +7,11 @@ dependencies {
   compile spec.product.pegasus.restliCommon
 }
 
-sourceSets.mainGeneratedDataTemplate.java.srcDirs('src/main/java/com/linkedin/testing/urn')
```
mars-lan, line 10: What's wrong with this old approach?

jplaisted: 1) Duplicate class definitions. You can't just copy classes from one jar to another.
2) Mixing schema definitions and java in the first place was probably not a great idea; should've separated into different modules, imo (other schema libraries don't need all your java baggage!). Note there's an exception where the PDL _needs_ the Java (e.g. `@java.class`); I moved the Java sources out of the main jar. So really there's only the data template jar for these. In fact, let me not publish the normal jars here...
3) Gradle will always end up using the `-data-template` jar intra-project, so we can't get rid of it or rename it (I'd love to rename it in most cases to be the "normal" jar).

Basically, if we do things right (from my perspective), each module produces one jar. It may or may not have a `-data-template` tag. What its called is a bit moot, the main idea behind that statement is achieved (one dependency)

mars-lan (approved):

PR #15, opened 2020-10-19
rfc: Elasticsearch Integration Testing

jplaisted:
Brief: Design an easy to use test framework for integration testing Elasticsearch with GMA.

#17 

## Checklist


jplaisted:
Rendered at https://github.com/jplaisted/datahub-gma/blob/integRfc/docs/rfc/active/elasticsearch-integration-testing.md

camelliazhang:
A general comment: do you want to mention/try some existing integration plugins? 
https://sharing.luminis.eu/blog/elasticsearch-instances-for-integration-testing/

docs/rfc/active/elasticsearch-integration-testing.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+## Motivation
+
+To increase test coverage and reliability of GMA's search stack.
+
+## Requirements
+
+- Must be able to easily start an Elasticsearch instance for testing at test start up, and then clean it up at test tear
```
mars-lan, line 71: Will the startup/teardown be at the per-test or per-class granularity?

docs/rfc/active/elasticsearch-integration-testing.md
```diff
@@ -0,0 +1,232 @@
...
+
+As stated above, we will:
+
+- Use JUnit 5.
+- Use assertj.
+- Provide a _default_ implementation that uses the Testcontainers framework.
+
+We can add two new modules to the project:
+
+1. `dao-impl/elasticsearch-dao-integ-testing`: Contains JUnit 5 extension and interfaces, as well as assertj assertions.
```
mars-lan, line 105: Maybe create a top-level folders to hold all test-related modules, e.g. test models, integ test harness, test utils etc?

docs/rfc/active/elasticsearch-integration-testing.md
```diff
@@ -0,0 +1,232 @@
...
+
+## How we teach this
+
+We can add an extensive readme to the new modules, as well as create robust java doc documentation on the classes.
+
+## Drawbacks
+
+- Works only with JUnit 5.
+- Assertions provided only with assertj.
+- **Works only with Elasticsearch 5.** This is the biggest drawback here. This design has not considered an
```
mars-lan, line 212: If we expect the lib to take a hard dependency on ES 5, maybe we should name the module as such for clarity?

mars-lan (changes requested):
Some minor comments. Otherwise LGTM

jplaisted:
> A general comment: do you want to mention/try some existing integration plugins?
> https://sharing.luminis.eu/blog/elasticsearch-instances-for-integration-testing/

See the RFC. It does mention why we're building something of a custom framework here.

A) Not everyone is going to be able to use the same solution (e.g. we can't use docker at LinkedIn). Some abstraction here is needed.
B) There's more to the setup than just starting/stopping ES. There's also connecting to it, making an index, setting settings/mappings, making DAOs, etc.

The RFC also mentions it will use Testcontainers under the hood by default.

docs/rfc/active/elasticsearch-integration-testing.md
jplaisted: I think we can have both. Easily specified by making a `SearchIndex` static or non-static :)

docs/rfc/active/elasticsearch-integration-testing.md
jplaisted: SG. So

```
testing/
  test-models/
  elasticsearch-dao-integ-testing/
  elasticsearch-dao-integ-testing-docker/
```

? Moving `test-models`?

docs/rfc/active/elasticsearch-integration-testing.md
jplaisted: We didn't with our other libraries. Probably better to be consistent. So lack of number means 5, in the future we'll probably have modules with 7 in their name...

docs/rfc/active/elasticsearch-integration-testing.md
mars-lan: I'm more concerned about the overhead of starting/stopping ES instances if it's done at a per-test level.

docs/rfc/active/elasticsearch-integration-testing.md
mars-lan: LGTM

docs/rfc/active/elasticsearch-integration-testing.md
mars-lan: The bigger question is can we somehow make the test harness ES version insensitive?

docs/rfc/active/elasticsearch-integration-testing.md
jplaisted: Oh, ah. I meant index life cycle can be per test or class by setting the field(s) as static/instance.

The cluster itself should have a long lived life cycle, ideally. Across all tests. Yes.

docs/rfc/active/elasticsearch-integration-testing.md
jywadhwani: +1 our tests should (ideally) be independent of the ES version. Atleast the generic tests (ES version insensitive) should reside in one place?

jplaisted: Changing the version that is started with docker/test containers is super simple, just involves changing a string.

The issue is the test framework here is also going to be connecting to it, and making DAOs, both of which link against the ES client (directly or transitively). So we have a tight coupling, like it or not. Even if we didn't have a coupling directly, we've have it transitively because of how the DAO is written today.

We'll be encountering the same issues when we migrate the DAO to 7.

docs/rfc/active/elasticsearch-integration-testing.md
mars-lan: SGTM

docs/rfc.md
```diff
@@ -1,4 +1,4 @@
-# DataHub RFC
+# GMA RFC
```
mars-lan, line 1: Move this to a separate PR? No review needed.

mars-lan: And all the cleanup of existing RFCs

mars-lan (changes requested):
Still got some extraneous files?

jplaisted:
> Still got some extraneous files?

Done

docs/rfc/active/15-elasticsearch-integration-testing/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+As stated above, we will:
+
+- Use JUnit 5.
+- Use assertj.
+- Provide a _default_ implementation that uses the Testcontainers framework.
+
+We can add two new modules to the project:
+
+1. `dao-impl/elasticsearch-dao-integ-testing`: Contains JUnit 5 extension and interfaces, as well as assertj assertions.
```
mars-lan, line 106: Please update the module paths

mars-lan (approved):

docs/rfc/active/15-elasticsearch-integration-testing/README.md
```diff
@@ -0,0 +1,233 @@
...
+   * should also write more specific tests to verify their settings / mappings / queries / rankings, etc.
+   */
+}
+```
+
+## Motivation
+
+To increase test coverage and reliability of GMA's search stack.
+
+## Requirements
```
jywadhwani, line 69: I believe testing `browse` feature which is also dependent on search stack is not in the scope of this doc?

docs/rfc/active/15-elasticsearch-integration-testing/README.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+With helper class stubs:
+
+```java
+/**
+ * Keeps track of all bulk requests made to Elasticsearch while testing.
+ */
+class BulkRequestsContainer {
+  Colleciton<BulkRequest> getAllRequests();
```
jywadhwani, line 147: nit: you mean `Collection`? same for the line below.

docs/rfc/active/15-elasticsearch-integration-testing/README.md
```diff
@@ -0,0 +1,233 @@
...
+be clearer to just query the DAO and assert the results in the test. We should debate this before finalizing the list of
+assertions we wish to provide.
+
+```java
+@Test
+public void withAssertionLibrary() {
+  // This might actually be less clear than without assertions below? At least code wise. We could provide a much
+  // clearer error message in the event of a failure, however.
+  assertThat(searchIndex)
+    .usingConfig(config)
```
jywadhwani, line 182: I assume the search query template will be pulled from the search config and this assertion will test that feature. How about autocomplete? Also we provide facet fields in the search config primarily used for aggregations and those fields require certain types/params (https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html). Do we provide assertion helpers for them as well?

docs/rfc/active/15-elasticsearch-integration-testing/README.md
jplaisted: Correct. We can always add on later. For now, I'm looking just to test `ESSearchDAO`. We can consider browse DAO later.

docs/rfc/active/15-elasticsearch-integration-testing/README.md
```diff
@@ -0,0 +1,233 @@
...
+class ElasticsearchConnection {
+  RestHighLevelClient getRestHighLevelClient();
+  TransportClient getTransportClient();
+}
+```
+
+### Common Assertions
+
+Commonly used assertions that we should provide helpers for will include:
+
```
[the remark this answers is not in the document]
jplaisted: That's just naturally part of test set up / tear down, I guess? Maybe adding it to the "sanity" tests later would be a good idea.

docs/rfc/active/15-elasticsearch-integration-testing/README.md
jplaisted: Presumably, this isn't mean to be exhaustive.

PR #25, opened 2020-10-27
Start adding ES integration test framework.

jplaisted:
This adds the initial module, as well as some basic annotations and interfaces.

## Checklist


settings.gradle
```diff
@@ -5,5 +5,6 @@ include 'dao-impl:elasticsearch-dao'
 include 'dao-impl:neo4j-dao'
 include 'restli-resources'
 include 'testing:core-models-testing'
+include 'testing:elasticsearch-dao-integ-testing'
```
mars-lan, line 8: Is the intention to have this module as ES5-only, or the base module from which ES5 & ES7 implementations will be built?

jplaisted: ES5 only. There will be direct references to ES and the DAO in this module, so it will depend on 5 directly.

mars-lan: Should we name it as such? Or do you plan to rename this and DAO in one big swoop later when ES7 support is added?

jplaisted: I want everything to be consistent with everything else.

imo we shouldn't care that much about naming. it's fine if 7 has 7 in the name, imo, because eventually there will be an 8...

jplaisted:
Ping :)

mars-lan:
Might want to resolve conflict first?

mars-lan (commented):
LGTM for overall. Will let @jywadhwani comment on the actual tests.

jplaisted:
> LGTM for overall. Will let @jywadhwani comment on the actual tests.

What tests? This is just basic interfaces for now :p

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchConnection.java
```diff
@@ -0,0 +1,30 @@
...
+
+import javax.annotation.Nonnull;
+import org.elasticsearch.client.RestHighLevelClient;
+import org.elasticsearch.client.transport.TransportClient;
+
+
+/**
+ * POJO to hold Elasticsearch client objects.
+ */
+public final class ElasticsearchConnection {
```
jywadhwani, line 11: why not use the builder pattern here?

testing/elasticsearch-dao-integ-testing/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'java'
+
+dependencies {
+  compile project(':dao-impl:elasticsearch-dao')
+
+  compile externalDependency.junitJupiterApi
+  compile externalDependency.junitJupiterParams
+  compile externalDependency.assertJ
```
jywadhwani, line 8: nit: order external dependencies alphabetically?

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchConnection.java
jplaisted: Small, unlikely to expand, requires all parameters, not going to be constructed in many places.

jplaisted: (re all params: builder would require validation logic to validate required fields).

testing/elasticsearch-dao-integ-testing/README.md
```diff
@@ -0,0 +1,54 @@
...
+import com.linkedin.metadata.testing.SearchIndex;
+import com.linkedin.metadata.testing.annotations.SearchIndexMappings;
+import com.linkedin.metadata.testing.annotations.SearchIndexSettings;
+import com.linkedin.metadata.testing.annotations.SearchIndexType;
+import org.junit.jupiter.api.Test;
+
+@ElasticsearchIntegrationTest // 2
+public class ExampleTest {
+  @SearchIndexType(MySearchDocument.class) // 3
+  @SearchIndexSettings("/settings.json") // 4
```
jywadhwani, line 29: why do we have separate interfaces for index settings and mappings? In our current implementation, we provide both mappings and settings in one "config" file (https://github.com/linkedin/datahub/tree/master/docker/elasticsearch-setup)

jplaisted: The Java API provides no way to set it via one json object, at least for 5.

We _could_ accept one file, but then it'd be on us to parse it so we could split out the settings and mappings and then make the separate calls.

I'd be fine with that, up to you.

Also, worth noting, but those files are not part of GMA / any API we currently have. Those are in docker, and nothing besides that little docker script cares about them. They could easily be split up.

jywadhwani: Let's keep it separate in that case. The docker script should also use individual settings and mapping file to keep everything consistent.

jplaisted: ack

jywadhwani (approved):

PR #27, opened 2020-10-28
[Breaking] Update definition of URN to match LI's internal definition.

jplaisted:
This should help us avoid more breakages when bringing in changes / pushing out changes. We're less likely to hit compiler errors due to bad references to Urn.

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #28, opened 2020-10-28
refactor: method to validate entity type from raw string form of urn

jywadhwani:
## Checklist

### DIFF
--- a/core-models/src/main/javaPegasus/com/linkedin/common/urn/Urn.java
+++ b/core-models/src/main/javaPegasus/com/linkedin/common/urn/Urn.java
@@ -99,6 +99,11 @@ public class Urn {
     }
   }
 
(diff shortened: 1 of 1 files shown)


jywadhwani:
primarily to bring this change back https://github.com/linkedin/datahub/pull/1958

keremsahin1:
How does this aling with https://github.com/linkedin/datahub-gma/pull/27 ?

jplaisted:
Please do not change `Urn` ever! At this point let's keep it totally in sync with the definition we have internally. I added a `Urns` class we can add to instead.

jplaisted:
Also, why do we need this? Is writing `validateUrn(new Urn(str, type))` that frequent?

jywadhwani:
intended to call `validateUrn` in urn definitions of each entity. Looks like we are not going to be using a util function anymore to validate the same (as per https://github.com/linkedin/datahub/pull/1977/files) - in favor of code gen.

Will discard this PR.

Issue #29, opened 2020-10-29
Enable WError


jplaisted:
We should promote warnings to errors, and clean up said errors. There are a few places in our code that are a bit sloppy due to lack of these warnings being enforced (e.g. `rawtype` and `unchecked` warnings)

PR #33, opened 2020-11-04
fix: enable werror for dao-api.

jplaisted:
#29 

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,8 +70,10 @@ apply from: "./gradle/release.gradle"
 // than in one big change.
 def wErrorProjects = [
     project(':core-models'),
-    // project(':dao-api'),
-    // project(':dao-api-impl'),
+    project(':dao-api'),
+    // project(':dao-impl:ebean-dao'),
+    // project(':dao-impl:elasticsearch-dao'),
+    // project(':dao-impl:neo4j-dao'),
     // project(':restli-resources'),
     project(':testing'),
     project(':validators')
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -74,15 +74,15 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
   private static final int DEFAULT_MAX_TRANSACTION_RETRY = 3;
 
-  protected final BaseMetadataEventProducer _producer;
+  protected final BaseMetadataEventProducer<?, ASPECT_UNION, URN> _producer;
   protected final LocalDAOStorageConfig _storageConfig;
 
   // Maps an aspect class to the corresponding retention policy
   private final Map<Class<? extends RecordTemplate>, Retention> _aspectRetentionMap = new HashMap<>();
 
   // Maps as aspect class to a list of post-update hooks
-  private final Map<Class<? extends RecordTemplate>, List<BiConsumer<Urn, RecordTemplate>>> _aspectPostUpdateHooksMap =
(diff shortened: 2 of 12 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -74,15 +74,15 @@
 
   private static final int DEFAULT_MAX_TRANSACTION_RETRY = 3;
 
-  protected final BaseMetadataEventProducer _producer;
+  protected final BaseMetadataEventProducer<?, ASPECT_UNION, URN> _producer;
```
keremsahin1, line 77: Any reason why it's not `? extends RecordTemplate` ? This is repeated in multiple lines.

dao-api/src/main/java/com/linkedin/metadata/dao/equality/AlwaysFalseEqualityTester.java
```diff
@@ -1,23 +1,25 @@
...
-import com.linkedin.data.template.DataTemplate;
+import com.linkedin.data.template.RecordTemplate;
 import javax.annotation.Nonnull;
 
 
 /**
  * A {@link EqualityTester} that always returns false.
  */
-public class AlwaysFalseEqualityTester<T extends DataTemplate> implements EqualityTester<T> {
+public class AlwaysFalseEqualityTester<T extends RecordTemplate> implements EqualityTester<T> {
```
keremsahin1, line 10: Why are we switching from DataTemplate to RecordTemplate?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jplaisted: `BaseMetadataEventProducer` already has the bounds. No need to respecify it.

dao-api/src/main/java/com/linkedin/metadata/dao/equality/AlwaysFalseEqualityTester.java
jplaisted: DataTemplate was not useful. It is _too_ high level. All the usages (setEqualityTester on BaseLocalDAO) were `RecordTempatae`. I cannot pass a `EqualityTester<?>` to setEqualityTester if the bound is `DataTemplate`, which is basically what happens when you use inference by inlining with `dao.setEqualityTester(AspectFoo.class, AlwaysFalseEqualityTester.instance());`

jplaisted:
ping

build.gradle
```diff
@@ -66,8 +66,10 @@ apply plugin: 'org.shipkit.shipkit-auto-version'
 // than in one big change.
 def wErrorProjects = [
     project(':core-models'),
-    // project(':dao-api'),
-    // project(':dao-api-impl'),
+    project(':dao-api'),
+    // project(':dao-impl:ebean-dao'),
+    // project(':dao-impl:elasticsearch-dao'),
+    // project(':dao-impl:neo4j-dao'),
```
camelliazhang, line 72: What are these commented out lines for?

camelliazhang (commented):
Otherwise, LGTM.

build.gradle
jplaisted: See the TODO 4 lines above :)

It can be nice to comment out stuff like this when rolling out; makes it clear what is left.

jplaisted:
Also, since this is a breaking change, and the others will be too, how do we feel about adding them to a separate branch, and then once they're all in merging to master without review? i.e. review small PRs to the other branch, then we can have one big breaking change all at once that was "already reviewed" in chunks?

Issue #17, opened 2020-10-20
Elasticsearch Integration Tests


jplaisted:
We should add integration tests for Elasticsearch, or at least a framework for it in GMA so it is easy to write integration tests in DataHub.

PR #38, opened 2020-11-06
feat: add more classes for testing elasticsearch.

jplaisted:
This adds a few search configs for testing, as well as the "SearchIndex" class which is a container object for most other things you'd want to use while testing (e.g. daos, bulk requests, etc).

#17 

## Checklist

### DIFF
--- a/testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/BulkRequestsContainer.java
+++ b/testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/BulkRequestsContainer.java
@@ -1,0 +1,69 @@
+package com.linkedin.metadata.testing;
+
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import org.elasticsearch.action.bulk.BulkProcessor;
+import org.elasticsearch.action.bulk.BulkRequest;
+import org.elasticsearch.action.bulk.BulkResponse;
+
+
+/**
+ * Records the results of all requests and also has the capability to flush and then block for all results via the
+ * {@link #flushAndSettle()} method.
+ *
+ * <p>Implementation wise, this just a super light wrapper around {@link BulkProcessor} and {@link TestBulkListener} so
+ * that limited functionality is exposed to tests.
+ */
+public final class BulkRequestsContainer {
+  private final BulkProcessor _bulkProcessor;
+  private final TestBulkListener _testBulkListener;
+
+  BulkRequestsContainer(@Nonnull BulkProcessor bulkProcessor, @Nonnull TestBulkListener testBulkListener) {
+    _bulkProcessor = bulkProcessor;
+    _testBulkListener = testBulkListener;
+  }
+
+  /**
+   * Resets the state of this container, clearing the recorded requests and responses.
+   */
+  public void reset() {
+    _testBulkListener.reset();
+  }
+
+  /**
+   * Flushes any queued requests and then waits for them to settle.
+   */
+  public void flushAndSettle() throws InterruptedException {
+    _bulkProcessor.flush();
+    _testBulkListener.settle();
+  }
+
+  /**
+   * All requests (executing, successful, and errored) that this bulk processor saw.
+   */
+  public Set<BulkRequest> getAllRequests() {
+    return _testBulkListener.getAllRequests();
+  }
+
+  /**
+   * The set of currently executing requests.
+   */
+  public Set<BulkRequest> getExecutingRequests() {
+    return _testBulkListener.getExecutingRequests();
+  }
+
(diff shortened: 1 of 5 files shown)


testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/SearchIndex.java
```diff
@@ -0,0 +1,138 @@
...
+  private final ElasticsearchConnection _connection;
+  private final String _name;
+  private final ESBulkWriterDAO<DOCUMENT> _bulkWriterDAO;
+  private final BulkRequestsContainer _requestContainer;
+
+  public SearchIndex(@Nonnull Class<DOCUMENT> documentClass, @Nonnull ElasticsearchConnection connection,
+      @Nonnull String name) {
+    _documentClass = documentClass;
+    _connection = connection;
+    _name = name;
```
jywadhwani, line 33: isn't the search index name essentially search document class name (normalized)?

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/SearchIndex.java
```diff
@@ -0,0 +1,138 @@
...
+        .setSettings(settingsJson, XContentType.JSON)
+        .get();
+  }
+
+  /**
+   * Sets the mappings of this index.
+   *
+   * @param mappingsJson the new Elasticsearch mappings as a JSON string
+   */
+  public void setMappings(@Nonnull String mappingsJson) {
```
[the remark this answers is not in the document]
jplaisted: Yeah, out of scope. That is easily a separate set of utilities, and doesn't even require an elasticsearch dependency. ie should just involve parsing json and can be done as a unit test.

Good idea for the future though.

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/SearchIndex.java
jplaisted: I don't like that we forced that tbh. It was a nice convention, but is awkwardly enforced, imo.

Anyway for tests it is going to be _really_ nice to make sure they all have different names. My plan is to have them be named something like `testMethod_testClass_timestamp` so they're unique per test run. No chance of anything interfering with each other by mistake.

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/BulkRequestsContainer.java
```diff
@@ -0,0 +1,69 @@
...
+  public void flushAndSettle() throws InterruptedException {
+    _bulkProcessor.flush();
+    _testBulkListener.settle();
+  }
+
+  /**
+   * All requests (executing, successful, and errored) that this bulk processor saw.
+   */
+  public Set<BulkRequest> getAllRequests() {
+    return _testBulkListener.getAllRequests();
```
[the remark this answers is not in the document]
jplaisted: That's a good question! My initial thought was that we'd need to do object equality because we may not always get the same reference from Elasticsearch when it starts and ends. However, it does look like `BulkRequest` does not implement any object equality.

Let me switch this to a map from execution ID to request? Then we should be safe.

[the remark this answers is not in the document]
jplaisted: I think I had tested it earlier and references were fine, but let's be safe :)

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/SearchIndex.java
jywadhwani: It might be worth adding what you mean by search index here? Are we talking about the alias or physical search index? 
I agree that we should not enforce any rules on the `physical` search index name. However, we are enforcing the alias to be search document class name and it is being used in the search index job https://github.com/linkedin/datahub/blob/master/metadata-jobs/mae-consumer-job/src/main/java/com/linkedin/metadata/kafka/MetadataAuditEventsProcessor.java#L124

jplaisted: > Are we talking about the alias or physical search index?

Shouldn't matter. That's the whole point of aliases.

In this case the literal index name will be something like `testMethod_testClass_timestamp` so literally ever test method has its own index and the test methods cannot accidentally interfere with each other via the index state.

jywadhwani (approved):

PR #46, opened 2020-11-12
fix: fix the lifecycle of the elasticsearch instances.

jplaisted:
The store is torn down after each class; we want the instance to live the life of the JVM.

## Checklist

### DIFF
--- a/testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
+++ b/testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
@@ -33,10 +33,28 @@ final class ElasticsearchIntegrationTestExtension
   private static final ExtensionContext.Namespace NAMESPACE =
       ExtensionContext.Namespace.create(ElasticsearchIntegrationTestExtension.class);
 
-  private final static String CONTAINER_FACTORY = "containerFactory";
-  private final static String CONNECTION = "connection";
   private final static String STATIC_INDICIES = "staticIndicies";
   private final static String INDICIES = "indicies";
+  private static ElasticsearchConnection _connection = null;
+
+  private ElasticsearchConnection startElasticSearch() throws Exception {
+    if (_connection != null) {
+      return _connection;
+    }
+
+    final ElasticsearchContainerFactory factory = getContainerFactory();
+    _connection = factory.start();
+
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+      try {
+        factory.close();
+      } catch (Throwable throwable) {
+        throwable.printStackTrace();
+      }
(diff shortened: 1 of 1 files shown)


testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
```diff
@@ -33,10 +33,28 @@
...
 
-  private final static String CONTAINER_FACTORY = "containerFactory";
-  private final static String CONNECTION = "connection";
   private final static String STATIC_INDICIES = "staticIndicies";
   private final static String INDICIES = "indicies";
+  private static ElasticsearchConnection _connection = null;
+
+  private ElasticsearchConnection startElasticSearch() throws Exception {
+    if (_connection != null) {
+      return _connection;
```
RealChrisL, line 42: Not exactly sure about the lifecycle, if the _**connection != null** and the **startElasticSearch()** is called, should we close then re-initiate the connection?

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
```diff
@@ -46,11 +64,7 @@ public void beforeAll(ExtensionContext context) throws Exception {
 
     final ExtensionContext.Store store = context.getStore(NAMESPACE);
 
-    final ElasticsearchContainerFactory factory = getContainerFactory();
-    final ElasticsearchConnection connection = factory.start();
-
-    store.put(CONTAINER_FACTORY, factory);
-    store.put(CONNECTION, connection);
+    final ElasticsearchConnection connection = startElasticSearch();
```
jywadhwani, line 67: why does `startElasticSearch` need to return a connection? can't we just call `startElasticSearch` method and use the `_connection` variable?

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
jplaisted: No.

The entire point is the variables are all static and only initialized once, ever. Meaning Elasticsearch is only started (and shutdown) once, since it is expensive.

testing/elasticsearch-dao-integ-testing/src/main/java/com/linkedin/metadata/testing/ElasticsearchIntegrationTestExtension.java
jplaisted: True, doesn't matter either way.

jywadhwani (approved):

PR #49, opened 2020-11-18
fix: do not split value of a filter by comma for an urn

jywadhwani:
Currently clients are using `comma` to provide multiple values a key can take while issuing filtered search query. Elasticsearch DAO splits the value by comma delimiter to populate the `terms` query. This does not work well when the value is an `Urn` with comma as some of the characters.
This change ensure we do not split the value by comma when the value is an Urn. Ideally we should be using the `OR` criterion to provide multiple values a key can take. But this will need changes on the client side before we stop supporting the `comma` feature to specify multiple values on the DAO side.

## Checklist

### DIFF
--- a/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
+++ b/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
@@ -47,17 +47,20 @@ public class SearchUtils {
   /**
    * Builds search query given a {@link Criterion}, containing field, value and association/condition between the two.
    *
-   * <p> If the condition between a field and value (specified in {@link Criterion}) is EQUAL, we construct a Terms query.
+   * <p>If the condition between a field and value (specified in {@link Criterion}) is EQUAL, we construct a Terms query.
    * In this case, a field can take multiple values, specified using comma as a delimiter - this method will split
    * tokens accordingly. This is done because currently there is no support of associating two different {@link Criterion}
    * in a {@link Filter} with an OR operator - default operator is AND.
+   *
+   * <p>This approach of supporting multiple values using comma as delimiter, prevents us from specifying a value that has comma
+   * as one of it's characters. This is particularly true when one of the values is an urn e.g. "urn:li:example:(1,2,3)".
+   * Hence we do not split the value (using comma as delimiter) if the value starts with "urn:li:".
    * TODO(https://github.com/linkedin/datahub-gma/issues/51): support multiple values a field can take without using delimiters like comma.
    *
-   * <p> If the condition between a field and value is not the same as EQUAL, a Range query is constructed. This
+   * <p>If the condition between a field and value is not the same as EQUAL, a Range query is constructed. This
    * condition does not support multiple values for the same field.
    *
    * @param criterion {@link Criterion} single criterion which contains field, value and a comparison operator
-   * @return QueryBuilder
    */
   @Nonnull
   public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
--- a/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
+++ b/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
@@ -1,12 +1,10 @@
 package com.linkedin.metadata.dao.utils;
 
-import com.linkedin.common.urn.Urn;
 import com.linkedin.metadata.query.Condition;
 import com.linkedin.metadata.query.Criterion;
 import com.linkedin.metadata.query.Filter;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -56,12 +54,10 @@ public class SearchUtils {
   public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
     final Condition condition = criterion.getCondition();
     if (condition == Condition.EQUAL) {
-      try {
-        Urn.createFromString(criterion.getValue());
+      if (criterion.getValue().startsWith("urn:li:")) {
         return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim());
-      } catch (URISyntaxException e) {
-        return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim().split("\\s*,\\s*"));
       }
+      return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim().split("\\s*,\\s*"));
     } else if (condition == Condition.GREATER_THAN) {
       return QueryBuilders.rangeQuery(criterion.getField()).gt(criterion.getValue().trim());
     } else if (condition == Condition.GREATER_THAN_OR_EQUAL_TO) {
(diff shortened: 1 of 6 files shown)


dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -54,7 +56,12 @@ private SearchUtils() {
   public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
     final Condition condition = criterion.getCondition();
     if (condition == Condition.EQUAL) {
-      return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim().split("\\s*,\\s*"));
+      try {
+        Urn.createFromString(criterion.getValue());
```
jplaisted, line 60: This is still just a heuristic.

The better thing to do would be to change `Criteron` to have a list of values rather than a comma delimited value that we then have to parse. You get the same issue here if there's some legitimate string that has a comma in it that isn't an URN.

I don't think this work around is worth it. Instead, we should work to fix clients ASAP, and then remove this splitting behavior.

jplaisted (changes requested):

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
jplaisted: Whoops, sorry, the better thing to do* would be what you suggested in the OP (fix clients and then remove behavior).

camelliazhang: Agree with @jplaisted . Split with comma was a hack solution already. There are a few options here, @keremsahin1 had proposed search filter redesign, let's discuss more on this internally. 
Meanwhile, for search 'Filter', have we explored the advanced search? That way we can leverage ES's out of box query parser. https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_field_names

camelliazhang: Also, we need to think/discuss is whether we need to keep both 5 & 7 DAO. Quite some duplicated codes here.

keremsahin1: Agreed with @jplaisted and @camelliazhang . Some time ago I shared a proposal to redesign filter to make it cover different filtering use cases - without hacks. We can iterate more on that and prioritize the work if necessary.

I would say let's not try to add more hacks here and opt for a long term, more robust solution.

jplaisted: Since this apparently blocking another team, let's considering adding this.

HOWEVER, and this applies to ALL workarounds like this, please document (in the code, in comments) why this code is the way it is, and add a TODO that links to a ticket to clean it up.

jywadhwani: @camelliazhang  we will need to support both DAOs in `datahub-gma` till we remove support for ES5 in `datahub`.
I have created an issue for tracking purpose https://github.com/linkedin/datahub-gma/issues/51

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -56,12 +54,10 @@ private SearchUtils() {
   public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
     final Condition condition = criterion.getCondition();
     if (condition == Condition.EQUAL) {
-      try {
-        Urn.createFromString(criterion.getValue());
+      if (criterion.getValue().startsWith("urn:li:")) {
```
jplaisted, line 57: Please a comment which states:

- What this is doing
- Why we're doing it
- Add a `TODO(github issue url): explanation of todo` for the long term solution

jplaisted (changes requested):

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
jywadhwani: Added, hope it looks good.

dao-impl/elasticsearch-dao-7/src/test/resources/UrnFilterQuery.json
```diff
@@ -0,0 +1,36 @@
...
+              "value1",
+              "value2"
+            ],
+            "boost" : 1.0
+          }
+        },
+        {
+          "terms" : {
+            "key2" : [
+              "urn:li:entity:(foo,bar,baz)"
```
camelliazhang, line 19: Add a test where a value is sth like "a,b,c"?

jywadhwani: We already have one where value is `"value1, value2 "` which should cover this scenario?

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -45,15 +45,30 @@ private SearchUtils() {
...
+   * condition does not support multiple values for the same field.
    *
    * @param criterion {@link Criterion} single criterion which contains field, value and a comparison operator
-   * @return QueryBuilder
    */
   @Nonnull
   public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
     final Condition condition = criterion.getCondition();
     if (condition == Condition.EQUAL) {
+      if (criterion.getValue().startsWith("urn:li:")) {
```
camelliazhang, line 69: This regex works internally, but what about open source? urn:li seems be specific to LinkedIn?

jywadhwani: The definition of `Urn` is similar to what we have internally https://github.com/linkedin/datahub-gma/blob/master/core-models/src/main/javaPegasus/com/linkedin/common/urn/Urn.java

jywadhwani: @jplaisted can you please confirm?

camelliazhang: Looks like the breaking change was indeed made. Will approve this PR to un-block customers first.

camelliazhang (approved):
Thanks for the prompt fix!

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
jplaisted: In the future, this stuff really should just go right next to the actual line, not in java doc. Java doc should be reserved for documenting the API, and any hacks/workarounds shouldn't go in there :p

jywadhwani: Thanks!

PR #50, opened 2020-11-19
feat: produce apsect specific event only if there is a diff by default

RealChrisL:
In this PR, we align the aspect-specific MAE with the old union MAE behavior.
For those aspects not being migrated/onboarded to the aspect-specific events, the producer will validate before emitting events.

- Start to produce the aspect specific event from GMS by default only if there is a diff.
- Bump up version due to naming refactor. 

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -95,6 +95,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
   private boolean _emitAspectSpecificAuditEvent = false;
 
+  private boolean _alwaysEmitAspectSpecificAuditEvent = false;
+
   // Flag for enabling reads and writes to local secondary index
   private boolean _enableLocalSecondaryIndex = false;
 
@@ -208,10 +210,17 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   }
 
   /**
-   * Sets if aspect specific MAE should be always emitted after each update even if there's no actual value change.
+   * Sets if aspect specific MAE should be enabled.
    */
   public void setEmitAspectSpecificAuditEvent(boolean emitAspectSpecificAuditEvent) {
     _emitAspectSpecificAuditEvent = emitAspectSpecificAuditEvent;
+  }
+
+  /**
+   * Sets if aspect specific MAE should be always emitted after each update even if there's no actual value change.
+   */
+  public void setAlwaysEmitAspectSpecificAuditEvent(boolean alwaysEmitAspectSpecificAuditEvent) {
+    _alwaysEmitAspectSpecificAuditEvent = alwaysEmitAspectSpecificAuditEvent;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -299,8 +299,8 @@ public boolean isLocalSecondaryIndexEnabled() {
     }
 
     // TODO: Replace step 6 with step 6.1 with diff option after pipeline is fully migrated to aspect specific events.
-    // 6.1. Produce aspect specific MAE after a successful update
-    if (_emitAspectSpecificAuditEvent) {
+    // 6.1 Produce aspect specific MAE after a successful update
+    if (_alwaysEmitAspectSpecificAuditEvent || oldValue != newValue) {
```
jplaisted, line 303: Is this going to break any existing pipelines at LinkedIn? What about externally? Not everyone is onboarded to v5, what happens if I don't have a v5 kafka topic and this is called?

Do we need two flags? Both the old flag (to enable emitting any events) and the new flag (to enable/disable "no change" events?)?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -211,8 +211,8 @@ public void setAlwaysEmitAuditEvent(boolean alwaysEmitAuditEvent) {
   /**
    * Sets if aspect specific MAE should be always emitted after each update even if there's no actual value change.
    */
-  public void setEmitAspectSpecificAuditEvent(boolean emitAspectSpecificAuditEvent) {
-    _emitAspectSpecificAuditEvent = emitAspectSpecificAuditEvent;
+  public void setAlwaysEmitAspectSpecificAuditEvent(boolean alwaysEmitAspectSpecificAuditEvent) {
```
jplaisted, line 214: I don't think this renaming is worth breaking the API over. Just use the existing name and update the documentation.

Also, just remove the comment on the private variable, and only have the documentation on the setter. Let's just have it in one place, and let's pick the public one :)

jplaisted (changes requested):
Please add more rationale to the PR description.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
RealChrisL: SGTM, thanks for the suggestion.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
RealChrisL: No, for the internal, we were by default emitting all the MAEs via setter in the DAO factory, which will need to config each DAO one by one, in this PR, we align the aspect-specific MAE with the old MAE behavior. 

For those aspects not being migrated/onboarded to the aspect-specific events, the producer will [validate](https://github.com/linkedin/datahub/blob/dc9c877984111c036a4120677c242b1ed4b12b8e/metadata-dao-impl/kafka-producer/src/main/java/com/linkedin/metadata/dao/producer/KafkaMetadataEventProducer.java#L157) before emitting events.

The flags were intended for the backfill scenario, in which only one kind of the union event (old) or the aspect-specific event (new) should suffice (depends on your consumer subscribes to).

jplaisted: You're going to cause log spam on the failed validations.

imo worth having two flags. One to enable MAE v5 at all, and then the other for the backfill events.

The existing flag can stay as is (enables / disables MAE v5; defaults to false). Then we can have the enable/disable backfill events flag as a new flag.

RealChrisL: The logs are in the warn level, which should be able to filter out.
But yes, we can keep two flags and migrate the legacy models per entity.

jplaisted (approved):

PR #79, opened 2021-02-24
fix: use the getEntityType() to get the URN string.

RealChrisL:
## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -520,7 +520,8 @@ public static String getEntityTypeFromUrnClass(@Nonnull Class<? extends Urn> urn
   @Nonnull
   public static <URN extends Urn, ASPECT extends RecordTemplate> String getAspectSpecificMAETopicName(@Nonnull URN urn,
       @Nonnull ASPECT newValue) {
-    final String urnStr = urn.getClass().getSimpleName().toUpperCase();
+    String urnStr = urn.getClass().getSimpleName().toUpperCase();
+    urnStr = urnStr.startsWith("INTERNAL") ? urnStr.substring("INTERNAL".length()) : urnStr;
```
camelliazhang, line 524: Add unit test?

RealChrisL: Yes, create an internal urn to validate the topic name.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -520,7 +520,8 @@ public static String getEntityTypeFromUrnClass(@Nonnull Class<? extends Urn> urn
   @Nonnull
   public static <URN extends Urn, ASPECT extends RecordTemplate> String getAspectSpecificMAETopicName(@Nonnull URN urn,
       @Nonnull ASPECT newValue) {
-    final String urnStr = urn.getClass().getSimpleName().toUpperCase();
+    String urnStr = urn.getClass().getSimpleName().toUpperCase();
```
jplaisted, line 523: Document why this is here.

RealChrisL: Add the java comment.

jplaisted: That didn't really clarify things. That just states what the code is doing, not why.

```
// LinkedIn has some urns marked as internal urns, that start with Internal. We want to strip that off.
```

Keep in mind that this now impacts everyone. If someone else, at LI or externally, wanted to name some urn as InternalStuff, we'd always expect the topic name to be Stuff (or apparently STUFF? why are we doing all upper case? our topic names aren't actually all upper case?). There's no way to change this.

Would it be a reasonable alternative to say that users at LI should just include "Internal" in their topic name instead? Going further, I'm not sure it even makes sense to have "internal" urns in GMA since, reading up on it, they are meant to only be used by a superblock. GMA isn't part of a superblock? We shouldn't be taking other's data that isn't meant to be shared...

keremsahin1: I agree with John here. Internal urn is a quite Linkedin-ey concept (well you could say the same for urns but whatever, it's too late for that :)) and we shouldn't carry this handling logic down to datahub-gma. And as John mentioned there are already some disadvantages it brings.

I would suggest to handle this check at a a higher level and call this method after converting internal urn to actual urn. Does that make sense @RealChrisL ?

RealChrisL: I agree this is the LI-specific changes, let's move it internally.

dao-api/src/test/java/com/linkedin/metadata/dao/utils/ModelUtilsTest.java
```diff
@@ -273,7 +273,7 @@ public void testGetMAETopicName() throws URISyntaxException {
     FooUrn urn = new FooUrn(1);
     AspectFoo foo = new AspectFoo().setValue("foo");
 
-    assertEquals(ModelUtils.getAspectSpecificMAETopicName(urn, foo), "METADATA_AUDIT_EVENT_FOO_ASPECTFOO");
+    assertEquals(ModelUtils.getAspectSpecificMAETopicName(urn, foo), "METADATA_AUDIT_EVENT_ENTITYFOO_ASPECTFOO");
```
jplaisted, line 276: This should be `FOO` not `ENTITYFOO` right?

keremsahin1: Depends on how entity type is defined for FooUrn. It actually is EntityFoo :)
https://github.com/linkedin/datahub-gma/blob/master/testing/test-models/src/main/javaPegasus/com/linkedin/testing/urn/FooUrn.java#L9

RealChrisL: Yes, this is due to the definition of the [URN](https://github.com/linkedin/datahub-gma/blob/master/testing/test-models/src/main/javaPegasus/com/linkedin/testing/urn/FooUrn.java#L9), let me do the refactor of the test models and related tests.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -520,8 +520,7 @@ public static String getEntityTypeFromUrnClass(@Nonnull Class<? extends Urn> urn
   @Nonnull
   public static <URN extends Urn, ASPECT extends RecordTemplate> String getAspectSpecificMAETopicName(@Nonnull URN urn,
       @Nonnull ASPECT newValue) {
-    final String urnStr = urn.getClass().getSimpleName().toUpperCase();
-    return String.format("%s_%s_%s", METADATA_AUDIT_EVENT_PREFIX, urnStr.substring(0, urnStr.length() - "Urn".length()),
+    return String.format("%s_%s_%s", METADATA_AUDIT_EVENT_PREFIX, urn.getEntityType().toUpperCase(),
```
jplaisted, line 523: Is there a test for this? Can we add one?

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
jplaisted: Turns out `entityType` in the urn is what we want and won't have the internal prefix.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
RealChrisL: Add another test case after the test model refactor.

jplaisted (approved):

Issue #77, opened 2021-02-10
Bintray deprecation


jplaisted:
Bintray is being deprecated. Awaiting guidance from LinkedIn for a migration path forward.

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Deadline seems to be May 1st.

PR #80, opened 2021-02-24
Setup publishing to LI artifactory

jplaisted:
Bintray is going to be deprecated soon, prepare to switch to our artifactory instance.

## Checklist

### DIFF
--- a/docs/how/releases/continuous-releases.md
+++ b/docs/how/releases/continuous-releases.md
@@ -1,7 +1,7 @@
 # GMA Continuous Releases
 
 We create [GitHub releases](https://github.com/linkedin/datahub-gma/releases) and publish our artifacts to
-[Bintray](https://bintray.com/linkedin/maven/datahub-gma).
+[LinkedIn's open source artifactory](https://linkedin.jfrog.io/artifactory).
 
 We use [shipkit-auto-version](https://github.com/shipkit/shipkit-auto-version),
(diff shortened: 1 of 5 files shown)


docs/how/releases/continuous-releases.md
```diff
@@ -29,8 +29,8 @@ We use shipkit changelog to create the release notes of each version. See
 ## Automatic Publishing
 
 Automation is done with a [GitHub workflow](../../.github/workflows/gh-version.yml) that will create the GitHub release
-and then publish to bintray for every push of the `main` branch.
+and then publish to artifactory for every push of the `main` branch.
```
camelliazhang, line 32: Seems that we will stay in dual publishing mode for a while ? I see both bintrayUploadAll and  artifactoryPublishAll. Does it make sense to keep both mentioned here as well?

jplaisted: So far as I'm concerned "awhile" can be just this PR's version lol. I don't think we need to dual publish for very long, so its not worth having the documentation reflect that, imo.

jplaisted: Even if we do want to bake it awhile, I'm fine pointing new people just to the new area.

camelliazhang: Prefer consistency, but agree it's fine for this particular case.

camelliazhang (approved):

jplaisted:
https://github.com/linkedin/datahub-gma/issues/77

PR #81, opened 2021-02-25
Add a topic convention class for MXE v5, which determines how to name…

jplaisted:
… MXE v5 event topics and get their schemas.

We can then use this to dynamically get topic names (and schemas) from Urn/RecordTemplates, rather than hard code everything, which we have done internally.

This ships with a default implementation that allows for some customization, but without any will do %event type%_%entity%_%aspect%.

This is being split off from the one in DataHub, which was both v4 and v5 (but v5 was not implemented there; it is here, we can delete it there in a follow up).

https://github.com/linkedin/datahub/tree/125ae288f1601bef2c42f44db6d8eb58b380a304/metadata-events/mxe-registration/src/main/java/com/linkedin/mxe

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5.java
+++ b/dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5.java
@@ -1,0 +1,71 @@
+package com.linkedin.mxe;
+
+import com.linkedin.common.urn.Urn;
+import com.linkedin.data.template.RecordTemplate;
+import javax.annotation.Nonnull;
+
+
+/**
+ * The convention for naming event topics, meant to be used in conjunction with a {@link
+ * com.linkedin.metadata.dao.producer.BaseMetadataEventProducer}.
+ *
+ * <p>Different companies may have different naming conventions or styles for their kafka topics. Namely, companies
+ * should pick _ or . as a delimiter, but not both, as they collide in metric names.
+ *
+ * <p>This convention applies to MXE v5, not to MXE v4 or below. In MXE v5, every entity/aspect pair can have its
+ * own topic, where as in v4 (and below) topics were monolithic. There is a topic convention for v4 as well, but it is
+ * defined in the DataHub repo due to tight coupling with MXE v4.
+ */
+public interface TopicConventionV5 {
+  /**
+   * Returns the name of the metadata change event topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  @Nonnull
+  String getMetadataChangeEventTopicName(@Nonnull Urn urn, @Nonnull RecordTemplate aspect);
+
+  /**
+   * Returns the class that defines the MCE schema for a given topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  Class<?> getMetadataChangeEventSchema(@Nonnull Urn urn, @Nonnull RecordTemplate aspect) throws ClassNotFoundException;
+
+  /**
+   * Returns the name of the metadata audit event topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  @Nonnull
+  String getMetadataAuditEventTopicName(@Nonnull Urn urn, @Nonnull RecordTemplate aspect);
+
+  /**
+   * Returns the class that defines the MAE schema for a given topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  Class<?> getMetadataAuditEventSchema(@Nonnull Urn urn, @Nonnull RecordTemplate aspect) throws ClassNotFoundException;
+
+  /**
+   * Returns the name of the failed metadata change event topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  @Nonnull
+  String getFailedMetadataChangeEventTopicName(@Nonnull Urn urn, @Nonnull RecordTemplate aspect);
+
+  /**
+   * Returns the class that defines the FMCE schema for a given topic.
+   *
+   * @param urn the urn of the entity being updated
+   * @param aspect the aspect name being updated
+   */
+  Class<?> getFailedMetadataChangeEventSchema(@Nonnull Urn urn, @Nonnull RecordTemplate aspect)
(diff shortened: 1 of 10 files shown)


dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5Impl.java
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Default implementation of a {@link TopicConventionV5}, which is fully customizable for event names.
+ *
+ * <p>The newer aspect-entity specific event names are based on a pattern that can also be configured. The pattern is a
+ * string, which can use {@link #EVENT_TYPE_PLACEHOLDER}, {@link #ENTITY_PLACEHOLDER}, and {@link #ASPECT_PLACEHOLDER}
+ * as placeholders for the event type (MCE, MAE, FMCE, etc), entity name, and aspect name, respectively.
+ *
+ * <p>The default pattern is {@code %EVENT%_%ENTITY%_%ASPECT%}. So, for example, given an URN like
+ * {@code urn:li:pizza:0} and an aspect like {@code PizzaInfo}, you would ge the following topic names by default:
```
jjoyce0510, line 21: s/ge/get

dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5Impl.java
```diff
@@ -0,0 +1,234 @@
...
+ * </ul>
+ */
+public final class TopicConventionV5Impl implements TopicConventionV5 {
+  // Placeholders
+  public static final String EVENT_TYPE_PLACEHOLDER = "%EVENT%";
+  public static final String ENTITY_PLACEHOLDER = "%ENTITY%";
+  public static final String ASPECT_PLACEHOLDER = "%ASPECT%";
+
+  // v5 defaults
+  public static final String DEFAULT_EVENT_PATTERN = "%EVENT%_%ENTITY%_%ASPECT%";
```
jjoyce0510, line 36: Thoughts on DEFAULT_TOPIC_NAME_PATTERN?

dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5Impl.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  public TopicConventionV5Impl withOverride(@Nonnull Class<?> schema, @Nonnull String topicName) {
+    return withSchemaOverrides(new HashMap<Class<?>, String>() {{
+      put(schema, topicName);
+    }});
+  }
+
+  @Nonnull
+  private String getDefaultEventName(@Nonnull Class<?> eventSchema) {
```
jjoyce0510, line 118: nit: getDefaultTopicName?

dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5Impl.java
```diff
@@ -0,0 +1,234 @@
...
+    } else {
+      throw new IllegalArgumentException(String.format("Unrecognized MXE class name: %s", className));
+    }
+
+    return buildEventName(eventType, toUpperCamelCase(matcher.group("entity")),
+        toUpperCamelCase(matcher.group("aspect")));
+  }
+
+  @Nonnull
+  private String buildEventName(@Nonnull String eventType, @Nonnull String entityName, @Nonnull String aspectName) {
```
jjoyce0510, line 148: nit: buildTopicName?

dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5Impl.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    if (str.length() == 1) {
+      return str.toLowerCase();
+    }
+
+    return Character.toLowerCase(str.charAt(0)) + str.substring(1);
+  }
+
+  private String buildEventName(@Nonnull String eventType, @Nonnull Urn urn, @Nonnull RecordTemplate aspect) {
```
jjoyce0510, line 182: same nit

jjoyce0510 (commented):
Going forward, is the idea that v5 is the default, and folks will have to change code to opt into v4? Vice versa?

jplaisted:
> Going forward, is the idea that v5 is the default, and folks will have to change code to opt into v4? Vice versa?

At some point, not right after this PR. Need to still figure out a rollout externally. I'm 100% punting on that. We have been for awhile. We'll have to discuss how to do this carefully together :)

For clarity, we're on the verge of fully moving to v5 internally. We're not done, but we're close, and have many kafka topics and things using v5. We wanted to roll it out internally before thinking about an external rollout to ensure everything works ok.

The motivation for this is in the OP; you can see topics are hardcoded (and also there's a "Configs.java" file right next to it that hardcodes schemas). We have 200+ topics internally; these files are forked internally and are massive and a maintenance burden. It also creates a coupling between the emitter and all the models, which is part of what MXE v5 is meant to stop. Dynamically building topic names breaks that coupling. Customization as to how those names are built are just a cherry on top, so to speak (though we've had requests for this from open source users in the past, so I'm maintaining that behavior).

woodyzhou (approved):
LGTM

dao-api/src/main/java/com/linkedin/mxe/TopicConventionV5.java
```diff
@@ -0,0 +1,71 @@
...
+ * com.linkedin.metadata.dao.producer.BaseMetadataEventProducer}.
+ *
+ * <p>Different companies may have different naming conventions or styles for their kafka topics. Namely, companies
+ * should pick _ or . as a delimiter, but not both, as they collide in metric names.
+ *
+ * <p>This convention applies to MXE v5, not to MXE v4 or below. In MXE v5, every entity/aspect pair can have its
+ * own topic, where as in v4 (and below) topics were monolithic. There is a topic convention for v4 as well, but it is
+ * defined in the DataHub repo due to tight coupling with MXE v4.
+ */
+public interface TopicConventionV5 {
```
camelliazhang, line 19: This may sound more like the 5th version of topic convention logic :) 
How about MXEV5TopicConvention Or TopicConventionForMXEV5? WDYT

jplaisted: (fyi this was merged)

I don't think most people know what MXE is, and I don't want to add that additional confusion.

We should've had conventions from the beginning, in which case this would've evolved along with MXE :)

PR #94, opened 2021-04-13
feat(neo4j): Extend neo4j writer dao to write to a given db

keremsahin1:
Neo4j 4+ brings multiple DB support and we're planning to leverage this new feature. Writer DAO was only writing to the default DB. With this change, writer DAO could be configured to write to a specific graph DB by using `USE` keyword.
https://neo4j.com/docs/cypher-manual/4.1/clauses/use/

Unfortunately, we can't really write unit tests for at the moment. This needs bumping Neo4j test harness to 4+ which needs Java 11. And Java 11 build is blocked by ES7 migration.

### DIFF
--- a/dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
+++ b/dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
@@ -16,6 +16,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
+
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
@@ -23,6 +24,7 @@ import org.apache.commons.lang.time.StopWatch;
 import org.neo4j.driver.Driver;
 import org.neo4j.driver.Record;
 import org.neo4j.driver.Session;
+import org.neo4j.driver.SessionConfig;
 import org.neo4j.driver.exceptions.Neo4jException;
 
 import static com.linkedin.metadata.dao.Neo4jUtil.*;
@@ -123,17 +125,33 @@ public class Neo4jGraphWriterDAO extends BaseGraphWriterDAO {
 
   private static final int MAX_TRANSACTION_RETRY = 3;
   private final Driver _driver;
+  private SessionConfig _sessionConfig;
   private static Map<String, String> _urnToEntityMap = null;
   private DelegateMetricListener _metricListener = new DelegateMetricListener();
 
   public Neo4jGraphWriterDAO(@Nonnull Driver driver) {
-    this._driver = driver;
-    buildUrnToEntityMap(getAllEntities());
+    this(driver, SessionConfig.defaultConfig());
   }
 
-  /* Should only be sed for testing */
-  public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull Set<Class<? extends RecordTemplate>> allEntities) {
+  /**
+   * WARNING: Do NOT use this! This is not tested yet.
+   * Multi-DB support comes with Neo4j 4+.
+   * Although DAO works with Neo4j 4+, we can't bump Neo4j test harness to 4+ to test this because it needs Java 11
+   * And Java 11 build is blocked by ES7 migration.
+   */
+  public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull String databaseName) {
+    this(driver, SessionConfig.forDatabase(databaseName));
+  }
+
+  public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull SessionConfig sessionConfig) {
+    this(driver, sessionConfig, getAllEntities());
+  }
+
+  /* Should only be used for testing */
+  public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull SessionConfig sessionConfig,
+                             @Nonnull Set<Class<? extends RecordTemplate>> allEntities) {
     this._driver = driver;
+    this._sessionConfig = sessionConfig;
     buildUrnToEntityMap(allEntities);
   }
 
@@ -222,7 +240,7 @@ public class Neo4jGraphWriterDAO extends BaseGraphWriterDAO {
     final StopWatch stopWatch = new StopWatch();
     stopWatch.start();
     Exception lastException;
-    try (final Session session = _driver.session()) {
+    try (final Session session = _driver.session(_sessionConfig)) {
       do {
         try {
           session.writeTransaction(tx -> {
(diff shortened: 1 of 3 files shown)


jplaisted:
Sorry, why do we need this? Your doc seemed to indicate that we'll always be writing to the default DB, except in offline which will use the admin import tool.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
```diff
@@ -131,7 +132,19 @@ public Neo4jGraphWriterDAO(@Nonnull Driver driver) {
...
 
-  /* Should only be sed for testing */
+  /**
+   * WARNING: Do NOT use this! This is not tested yet.
+   * Multi-DB support comes with Neo4j 4+.
+   * Although DAO works with Neo4j 4+, we can't bump Neo4j test harness to 4+ to test this because it needs Java 11
+   * And Java 11 build is blocked by ES7 migration.
+   */
+  public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull String databaseName) {
+    this._databaseName = databaseName;
```
jplaisted, line 142: Please try to consolidate the constructors by having the other constructor call this constructor like

```
this(driver, null);
```

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
```diff
@@ -483,8 +496,12 @@ private Statement getOrInsertNode(@Nonnull Urn urn) {
     return buildStatement(statement, params);
   }
 
+  // visible for testing
   @Nonnull
-  private Statement buildStatement(@Nonnull String queryTemplate, @Nonnull Map<String, Object> params) {
+  Statement buildStatement(@Nonnull String queryTemplate, @Nonnull Map<String, Object> params) {
+    if (_databaseName != null) {
+      queryTemplate = String.format("USE %s ", _databaseName) + queryTemplate;
```
jplaisted, line 503: afaik the recommended way to set the database would be via the Java API. Any time there's a Java API equivalent for something it is probably preferred.

```
Session session = driver.session(SessionConfig.forDatabase(db))
```

It is not clear if that accepts `null` or not. This might be cleaner:

```
private Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull SessionConfig sessionConfig) {
  _driver = driver;
  _sessionConfig = sessionConfig;
  buildUrnToEntityMap(getAllEntities());
}

public Neo4jGraphWriterDAO(@Nonnull Driver driver) {
  this(driver, Session.defaultConfig());
}

public Neo4jGraphWriterDAO(@Nonnull Driver driver, @Nonnull String databaseName) {
  this(driver, Session.forDatabase(databaseName));
}

// ...
try (Session session = _driver.session(_sessionConfig))
```

jplaisted (changes requested):

keremsahin1:
> Sorry, why do we need this? Your doc seemed to indicate that we'll always be writing to the default DB, except in offline which will use the admin import tool.

Nope. Readers will always read from the default DB. There will be two Samza instances writing to different DBs. So, we need this for writer DAO.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
keremsahin1: Tried that but Java can't really resolve which constructor you really mean. The other constructor used for testing has a similar constructor which takes two params.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
keremsahin1: Oh that looks better, didn't know if we can do that via the driver. Let me check that.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
jplaisted: Ah good point :)

jplaisted:
> There will be two Samza instances writing to different DBs. So, we need this for writer DAO.

Good point!

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
jplaisted: How about

```
this(driver, null, getAllEntities());
```

And then have the "testing" constructor take 3 arguments?

keremsahin1: Yes, we can do that. This is actually a backward incompatible change to the testing constructor (which is `public` unfortunately) but I guess we can assume there is no user of it.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
keremsahin1: Alright done, thanks for the suggestion.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
jplaisted: Yeah, breaking changes can be made _judiciously_ :)

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
```diff
@@ -483,8 +501,9 @@ private Statement getOrInsertNode(@Nonnull Urn urn) {
     return buildStatement(statement, params);
   }
 
+  // visible for testing
   @Nonnull
-  private Statement buildStatement(@Nonnull String queryTemplate, @Nonnull Map<String, Object> params) {
+  Statement buildStatement(@Nonnull String queryTemplate, @Nonnull Map<String, Object> params) {
```
jplaisted, line 506: This can be made private again.

jplaisted (changes requested):

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/internal/Neo4jGraphWriterDAO.java
keremsahin1: I wanted to keep the test actually

jplaisted (approved):

PR #101, opened 2021-06-09
feat: add sorting by fields to SCSI

kaliang1:
This change adds sorting by fields to SCSI. A new parameter called IndexSortCriterion was added to the filter method, which specifies the aspect, path, value column, and sorting order.

In BaseLocalDAO, the signature of the abstract method `listUrns` was modified to include the sort criterion parameter.

## Checklist


jywadhwani (changes requested):

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexSortCriterion.pdl
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Sort order along with the aspect and path to sort it on, to be applied to the results.
+ */
+record IndexSortCriterion {
+
+  /**
+   * FQCN of the aspect class in the index table that this criterion refers to e.g. com.linkedin.common.Status
+   */
+  aspect: string
```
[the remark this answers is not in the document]
kaliang1: I think the `aspect` and `path` are needed to specify the field that the user wants to sort on. For example, if a user wanted to sort by the aspect AspectFoo and the path “/id” with the column stringval, then the query would be:

`SELECT DISTINCT(t0.urn), tsort.stringval FROM metadata_index t0 INNER JOIN metadata_index tsort ON t0.urn = tsort.urn WHERE tsort.aspect = “AspectFoo” and tsort.field = “/id” order by tsort.stringval ASC;`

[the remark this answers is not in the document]
jywadhwani: Actually you could use the `IndexCriterion` itself to specify the filter and then sort by the column. The current limitation is that `IndexCriterion` requires you to specify a value along with path.

If I make that change, i.e. provide (aspect, path) in the `IndexCriterion` without providing the `value`, do you think we should be fine with just providing the sort column and sort order?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -418,13 +425,27 @@ private IndexFilter getDefaultIndexFilter() {
...
+        return filterAspects(aspectClasses, filter, indexSortCriterion, lastUrn, pagingContext);
       }
     });
   }
 
+  /**
+   * Similar to {@link #filter(IndexFilter, IndexSortCriterion, String[], String, PagingContext)} but
+   * uses null sorting criterion.
+   */
+  @Finder(FINDER_FILTER)
```
jywadhwani, line 439: Do you think we can mark this finder method as deprecated? `FINDER_FILTER_SORT` has sort params as optional so we can leverage that method directly.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -400,15 +405,17 @@ private IndexFilter getDefaultIndexFilter() {
...
    */
-  @Finder(FINDER_FILTER)
+  @Finder(FINDER_FILTER_SORT)
   @Nonnull
   public Task<List<VALUE>> filter(
       @QueryParam(PARAM_FILTER) @Optional @Nullable IndexFilter indexFilter,
+      @QueryParam(PARAM_SORT) @Optional @Nullable IndexSortCriterion indexSortCriterion,
       @QueryParam(PARAM_ASPECTS) @Optional @Nullable String[] aspectNames,
       @QueryParam(PARAM_URN) @Optional @Nullable String lastUrn,
       @PagingContextParam @Nonnull PagingContext pagingContext) {
```
jywadhwani, line 421: Since we are creating a new finder method, I will use this opportunity to remove `pagingContext` and replace it with `count` since we are using cursor based pagination here.

jywadhwani (changes requested):

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexSortCriterion.pdl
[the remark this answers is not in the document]
kaliang1: That should work!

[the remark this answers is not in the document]
keremsahin1: When you get multiple aspects, how does sorting work? Do you only sort AspectFoo in this case?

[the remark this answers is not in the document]
kaliang1: Yes, it only sorts by the aspect provided so it will only sort AspectFoo.

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexSortCriterion.pdl
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Corresponding path column of the index table that this criterion refers to e.g. /removed (corresponding to field "removed" of com.linkedin.common.Status aspect)
+   */
+  path: string
+
+  /**
+   * Corresponding value column that this criterion refers to e.g. STRING_COLUMN
+   */
+  column: IndexColumn
```
jywadhwani, line 21: This is a storage specific information i.e. how we store the values. I wouldn't assume the client should know this information - also this mapping of value type to column could change.
If you know the aspect and path spec, you should be able to get the value type? On the server side you can then map it to the relevant SQL column?

jywadhwani (changes requested):

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexSortCriterion.pdl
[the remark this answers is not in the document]
jywadhwani: Sounds good.
I guess you missed a `WHERE` clause on `t0` in your above query. But I get what your requirement is.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +883,46 @@ private static String getStringForOperator(@Nonnull Condition condition) {
     return CONDITION_STRING_MAP.get(condition);
   }
 
+  @Nonnull
+  private String getSortingColumn(@Nonnull IndexSortCriterion indexSortCriterion) {
+    final EbeanMetadataIndex record = _server.find(EbeanMetadataIndex.class)
```
jywadhwani, line 888: querying the DB to get the value type/column is not optimal approach, will slow down the read path. Ideally we should be able to obtain the value type given an aspect and path. Can we explore feasibility of that approach?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +887,98 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+
+  @Nonnull
+  private <ASPECT extends RecordTemplate> String getSortingColumn(@Nonnull IndexSortCriterion indexSortCriterion) {
+    // split sort criterion path into an array
+    final Pattern leadingSpaceslashPattern = Pattern.compile("^[/ ]+");
+    final Pattern trailingSpaceslashPattern = Pattern.compile("[/ ]+$");
+    final Pattern slashPattern = Pattern.compile("/");
+
+    String pathSpecAsString = indexSortCriterion.getPath();
+    pathSpecAsString = leadingSpaceslashPattern.matcher(pathSpecAsString).replaceAll("");
```
jywadhwani, line 927: again there seems to be repeated code here as well https://github.com/linkedin/datahub-gma/blob/master/dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java#L409
Can we consolidate this logic in a util method that both can use? You could put this method in `RecordUtils` only.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +887,98 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+    final DataSchema.Type type = dataSchema.getDereferencedType();
+
+    if (type == DataSchema.Type.INT || type == DataSchema.Type.LONG) {
+      return EbeanMetadataIndex.LONG_COLUMN;
+    } else if (type == DataSchema.Type.BOOLEAN || type == DataSchema.Type.STRING) {
+      return EbeanMetadataIndex.STRING_COLUMN;
+    } else if (type == DataSchema.Type.DOUBLE || type == DataSchema.Type.FLOAT) {
+      return EbeanMetadataIndex.DOUBLE_COLUMN;
+    } else {
+      throw new IllegalArgumentException("Invalid sort criterion " + indexSortCriterion);
```
jywadhwani, line 960: Same here, let's add more details on why the criterion is invalid.

jywadhwani (changes requested):
This looks good, thanks for the change.
Added some more comments.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +885,73 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+      aspectClass = (Class<ASPECT>) Class.forName(aspectClassString);
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException("Exception occurred while trying to get the aspect class. ", e);
+    }
+
+    return RecordUtils.getAspectFromClass(aspectClass);
+  }
+
+  @Nonnull
+  private <ASPECT extends RecordTemplate> String getSortingColumn(@Nonnull IndexSortCriterion indexSortCriterion) {
```
jywadhwani, line 901: Can we add some tests for this method? We should be able to test for different fields such as an array of primitives, array of records, enum, union, etc.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +885,73 @@ private static String getStringForOperator(@Nonnull Condition condition) {
     return CONDITION_STRING_MAP.get(condition);
   }
 
+  @Nonnull
+  private <ASPECT extends RecordTemplate> ASPECT getAspectFromString(@Nonnull String aspectClassString) {
```
jywadhwani, line 889: possible to move this method as well in `RecordUtils`?

jywadhwani (changes requested):

dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java
```diff
@@ -400,17 +411,30 @@ private static Object invokeMethod(@Nonnull RecordTemplate record, @Nonnull Stri
...
   }
 
+  /**
+   * Returns a PathSpec as an array representation from a string representation.
+   *
+   * @param pathSpecAsString string representation of a path spec
+   * @return array representation of a path spec
+   */
+  @Nonnull
+  public static String[] getPathSpecAsArray(@Nonnull String pathSpecAsString) {
```
keremsahin1, line 421: Please make sure to add tests for any public util methods

dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java
```diff
@@ -400,17 +411,30 @@ private static Object invokeMethod(@Nonnull RecordTemplate record, @Nonnull Stri
...
+
   /**
    * Similar to {@link #getFieldValue(RecordTemplate, PathSpec)} but takes string representation of Pegasus PathSpec as
    * input.
    */
   @Nonnull
   public static Optional<Object> getFieldValue(@Nonnull RecordTemplate recordTemplate, @Nonnull String pathSpecAsString) {
-    pathSpecAsString = LEADING_SPACESLASH_PATTERN.matcher(pathSpecAsString).replaceAll("");
-    pathSpecAsString = TRAILING_SPACESLASH_PATTERN.matcher(pathSpecAsString).replaceAll("");
+    String[] pathSpecAsArray = getPathSpecAsArray(pathSpecAsString);
```
keremsahin1, line 434: nit: final

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -844,6 +846,146 @@ public void testStartsWith() {
...
+    String sortingColumn5 = EbeanLocalDAO.getSortingColumn(indexSortCriterion5);
+    assertEquals(sortingColumn5, EbeanMetadataIndex.DOUBLE_COLUMN);
+
+    // 6. float corresponds to double column
+    IndexSortCriterion indexSortCriterion6 = new IndexSortCriterion().setAspect(AspectBaz.class.getCanonicalName())
+        .setPath("/floatField").setOrder(SortOrder.DESCENDING);
+    String sortingColumn6 = EbeanLocalDAO.getSortingColumn(indexSortCriterion6);
+    assertEquals(sortingColumn6, EbeanMetadataIndex.DOUBLE_COLUMN);
+
+    // 7. array corresponds to string column
```
jywadhwani, line 887: So if this field is an array of N integers (say), what does sorting by ascending order of this array mean here?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -844,6 +846,146 @@ public void testStartsWith() {
...
+    IndexSortCriterion indexSortCriterion7 = new IndexSortCriterion().setAspect(AspectBaz.class.getCanonicalName())
+        .setPath("/arrayField").setOrder(SortOrder.DESCENDING);
+    String sortingColumn7 = EbeanLocalDAO.getSortingColumn(indexSortCriterion7);
+    assertEquals(sortingColumn7, EbeanMetadataIndex.STRING_COLUMN);
+
+    // 8. union corresponds to string column
```
jywadhwani, line 893: Here `unionField` is a union of two record templates. What does `/unionField` path imply? which field do we wish to sort on?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +885,59 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+   * @param sortColumn column the sort criterion is applied on which is used to construct the SQL query
    * @return String representation of SQL query
    */
   @Nonnull
-  private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriterionArray) {
-    String selectClause = "SELECT DISTINCT(t0.urn) FROM metadata_index t0";
+  private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriterionArray,
+      @Nullable IndexSortCriterion indexSortCriterion, @Nonnull String sortColumn) {
+    String selectClause = "SELECT DISTINCT(t0.urn)";
+    if (indexSortCriterion != null) {
```
jywadhwani, line 936: `sortColumn` should also be a non-empty string here for this block to execute?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -906,7 +958,16 @@ private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriter
             .append("?");
       }
     });
-    final String orderByClause = "ORDER BY urn ASC";
+    final String orderByClause;
+    if (indexSortCriterion != null) {
```
jywadhwani, line 962: same here, `sortColumn` should be non-empty string here?

jywadhwani (changes requested):

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
kaliang1: I think it would sort by the string literal of the array. It doesn't make too much sense to sort urns based on an aspect that is an array or union type so I will remove this case.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java
jywadhwani: Looks like this has been added.

dao-api/src/test/java/com/linkedin/metadata/dao/utils/RecordUtilsTest.java
```diff
@@ -165,6 +165,39 @@ public void testGetInvalidMetadataSnapshotClassFromName() {
     ModelUtils.getMetadataSnapshotClassFromName("com.linkedin.testing.AspectInvalid");
   }
 
+  @Test
+  public void testGetPathSpecAsArray() {
+    String ps1 = "/test/path";
```
jywadhwani, line 170: Let us also add a test for an array of records i.e. `/recordArray/*/<recordField>` ?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +886,71 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+
+    // get nested field
+    ASPECT aspect = RecordUtils.getAspectFromString(indexSortCriterion.getAspect());
+
+    final int pathSize = pathSpecArray.length;
+
+    for (int i = 0; i < pathSize - 1; i++) {
+      final String part = pathSpecArray[i];
+
+      if (part.equals("*")) {
```
jywadhwani, line 901: `*` would mean a path corresponding to an array. We shouldn't be supporting it right?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +886,71 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+      }
+
+      final RecordDataSchema.Field field = aspect.schema().getField(part);
+      final DataSchema dataSchema = field.getType();
+
+      final String nestedAspectName;
+      if (dataSchema.getDereferencedType() == DataSchema.Type.RECORD) {
+        nestedAspectName = ((RecordDataSchema) dataSchema).getBindingName();
+        aspect = RecordUtils.getAspectFromString(nestedAspectName);
+      } else if (dataSchema.getDereferencedType() == DataSchema.Type.ARRAY) {
```
jywadhwani, line 912: this is if the field is of type array? Are we allowing sort over an array of elements?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -844,6 +846,146 @@ public void testStartsWith() {
...
+    String sortingColumn8 = EbeanLocalDAO.getSortingColumn(indexSortCriterion8);
+    assertEquals(sortingColumn8, EbeanMetadataIndex.STRING_COLUMN);
+
+    // 9. invalid type
+    IndexSortCriterion indexSortCriterion9 = new IndexSortCriterion().setAspect(AspectBaz.class.getCanonicalName())
+        .setPath("/arrayField").setOrder(SortOrder.DESCENDING);
+    assertThrows(UnsupportedOperationException.class,
+        () -> EbeanLocalDAO.getSortingColumn(indexSortCriterion9));
+
+    // 10. array of records
```
jywadhwani, line 905: this shouldn't be supported either IMO.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -876,16 +886,71 @@ private static String getStringForOperator(@Nonnull Condition condition) {
...
+   * @param sortColumn column the sort criterion is applied on which is used to construct the SQL query
    * @return String representation of SQL query
    */
   @Nonnull
-  private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriterionArray) {
-    String selectClause = "SELECT DISTINCT(t0.urn) FROM metadata_index t0";
+  private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriterionArray,
+      @Nullable IndexSortCriterion indexSortCriterion, @Nonnull String sortColumn) {
+    String selectClause = "SELECT DISTINCT(t0.urn)";
+    if (indexSortCriterion != null && !sortColumn.isEmpty()) {
```
jywadhwani, line 949: Better way is to only take `IndexSortCriterion` as input in function `constructSQLQuery` and compute sort column within this method.
Then you can only check if sort column is non empty in this condition.

jywadhwani: Or you could check if `indexSortCriterion != null` - that should work too. I suggested having both conditions in my previous review, but it looks like `indexSortCriterion != null` implies `!sortColumn.isEmpty()`.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani: Sorry for the confusion here. Didn't realize `indexSortCriterion != null` implies `!sortColumn.isEmpty()`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
kaliang1: Sorry for the confusion, we should not be sorting over an array, I will fix this.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -400,31 +405,49 @@ private IndexFilter getDefaultIndexFilter() {
...
-  @Finder(FINDER_FILTER)
+  @Finder(FINDER_FILTER_SORT)
   @Nonnull
   public Task<List<VALUE>> filter(
       @QueryParam(PARAM_FILTER) @Optional @Nullable IndexFilter indexFilter,
+      @QueryParam(PARAM_SORT) @Optional @Nullable IndexSortCriterion indexSortCriterion,
       @QueryParam(PARAM_ASPECTS) @Optional @Nullable String[] aspectNames,
       @QueryParam(PARAM_URN) @Optional @Nullable String lastUrn,
-      @PagingContextParam @Nonnull PagingContext pagingContext) {
+      @QueryParam(PARAM_COUNT) int count) {
```
jywadhwani, line 421: Can we use a default value of 10 for the count? you can do this by specifying `@Optional("10")`.

jywadhwani:
> I don't think they can be merged because it is used by clients, but I deprecated it [here](https://github.com/linkedin/datahub-gma/pull/101/commits/4bec13eeebf69d29b6d74611273f234cd81fd02f).

That is correct. But more than internal, we need to be careful about our open source users. Making breaking changes can impact our (invisible) open source adopters.

jywadhwani (approved):

restli-resources/src/main/java/com/linkedin/metadata/restli/RestliConstants.java
```diff
@@ -5,6 +5,7 @@ private RestliConstants() { }
 
   public static final String FINDER_SEARCH = "search";
   public static final String FINDER_FILTER = "filter";
+  public static final String FINDER_FILTER_SORT = "filterAndSort";
```
keremsahin1, line 8: I don't think we need to use a different finder name here. Search also does sorting but we don't have a `searchAndSort` filter per se. Sorting param is already optional and it's obvious that this new finder is able to sort if that param is set. Also, other filter finder also does sorting but with a default order :)

And you already deprecate the other finder, so, it should be okay to use same finder name for both. Maybe, you can also put a note such that user should only implement one of the filter finders in their resource implementation.

keremsahin1 (approved):
Great job!

PR #103, opened 2021-06-11
fix: use same urn class in both read and write path in SCSI

jywadhwani:
## Checklist


keremsahin1 (commented):
You also need to change default entity type filter logic, I guess?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -371,8 +372,14 @@ private void updateUrnInLocalIndex(@Nonnull URN urn) {
     }
 
     final Map<String, Object> pathValueMap = _urnPathExtractor.extractPaths(urn);
+    final String entityType;
```
keremsahin1, line 375: Hmmm, you don't need to do this here (every time you index and urn into the index). Better to do this as part of the constructor?

keremsahin1 (approved):
Thanks for a much simpler fix

jywadhwani:
Looks like build is failing. I will revert this change and try to submit it again.

PR #107, opened 2021-06-16
feat: Use optimistic locking for db updates in ebean dao

keremsahin1:
Start using optimistic locking to prevent [lost updates](https://www.avaje.org/occ.html) in Ebean dao. The feature is flag guarded for now. After testing it in production, the plan is to make it the default behavior going forward.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -178,9 +178,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   }
 
   /**
-   * Determines whether we should use optimistic locking in updates.
-   *
-   * <p>The column to execute optimistic locking is controlled by {@link #DEFAULT_COLUMN_FOR_OPTIMISTIC_LOCK}
+   * Determines whether we should use optimistic locking in updates. Optimistic locking is done on {@code createdOn} column
    *
    * <p>DO NOT USE THIS FLAG! This is for LinkedIn use to help us test this feature without a rollback. Once we've
    * vetted this in production we will be removing this flag and making the the default behavior. So if you set this
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -30,6 +30,7 @@ import io.ebean.EbeanServerFactory;
 import io.ebean.ExpressionList;
 import io.ebean.PagedList;
 import io.ebean.Query;
+import io.ebean.SqlUpdate;
 import io.ebean.Transaction;
 import io.ebean.config.ServerConfig;
 import io.ebean.datasource.DataSourceConfig;
@@ -50,6 +51,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.persistence.OptimisticLockException;
 import javax.persistence.RollbackException;
 import javax.persistence.Table;
 import lombok.Value;
@@ -72,8 +74,9 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   private UrnPathExtractor<URN> _urnPathExtractor;
   private int _queryKeysCount = 0; // 0 means no pagination on keys
 
-  // TODO feature flag, remove when vetted.
+  // TODO feature flags, remove when vetted.
   private boolean _useUnionForBatch = false;
+  private boolean _useOptimisticLocking = false;
 
   @Value
   static class GMAIndexPair {
@@ -174,6 +177,18 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     _useUnionForBatch = useUnionForBatch;
   }
 
+  /**
+   * Determines whether we should use optimistic locking in updates. Optimistic locking is done on {@code createdOn} column
+   *
+   * <p>DO NOT USE THIS FLAG! This is for LinkedIn use to help us test this feature without a rollback. Once we've
+   * vetted this in production we will be removing this flag and making the the default behavior. So if you set this
+   * to true by calling this method, your code will break when we remove this method. Just wait a bit for us to turn
+   * it on by default!
+   */
+  public void setUseOptimisticLocking(boolean useOptimisticLocking) {
+    _useOptimisticLocking = useOptimisticLocking;
+  }
+
   @Nonnull
   private static EbeanServer createServer(@Nonnull ServerConfig serverConfig) {
     // Make sure that the serverConfig includes the package that contains DAO's Ebean model.
@@ -252,7 +267,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
         transaction.commit();
         lastException = null;
         break;
-      } catch (RollbackException | DuplicateKeyException exception) {
+      } catch (RollbackException | DuplicateKeyException | OptimisticLockException exception) {
         lastException = exception;
       }
     } while (++retryCount <= maxTransactionRetry);
@@ -276,7 +291,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     }
 
     // Save newValue as the latest version (v0)
-    save(urn, newValue, newAuditStamp, LATEST_VERSION, oldValue == null);
+    save(urn, newValue, newAuditStamp, LATEST_VERSION, oldValue == null, oldAuditStamp);
     return largestVersion;
   }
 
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -308,30 +323,56 @@ public static ServerConfig createTestingH2ServerConfig() {
...
     } else {
-      _server.update(aspect);
+      if (_useOptimisticLocking && oldAuditStamp != null) {
+
+        // Build manual SQL update query to enable optimistic locking on createdOn column
+        final String updateQuery = "UPDATE metadata_aspect "
+                        + "SET urn = :urn, aspect = :aspect, version = :version, metadata = :metadata, createdOn = :createdOn, createdBy = :createdBy "
+                        + "WHERE urn = :urn and aspect = :aspect and version = :version and createdOn = :oldCreatedOn";
+
+        SqlUpdate update = _server.createSqlUpdate(updateQuery);
```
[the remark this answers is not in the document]
keremsahin1: Thanks, done

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -1849,6 +1857,42 @@ public void testPageSizeGreaterThanResultsSize() {
...
+
+    ThrowingRunnable runnable = () -> dao.save(fooUrn, fooAspect, makeAuditStamp("fooActor", 789),
+            0, false, makeAuditStamp("fooActor", 123));
+    if (_useOptimisticLocking) {
+      assertThrows(OptimisticLockException.class, runnable);
+    } else {
+      try {
+        runnable.run();
+      } catch (Throwable t) {
+        fail("This shouldn't have failed when optimistic locking is disabled");
```
jywadhwani, line 1891: it shouldn't have failed if `oldAuditStamp` was `null` as well? can we verify this condition as well?

jywadhwani (changes requested):

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jywadhwani, line 1891: If optimistic locking was _NOT_ enabled, then the aspect should have been updated successfully. Can we verify that as well?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -308,30 +323,56 @@ public static ServerConfig createTestingH2ServerConfig() {
...
+        update.setParameter("urn", aspect.getKey().getUrn());
+        update.setParameter("aspect", aspect.getKey().getAspect());
+        update.setParameter("version", aspect.getKey().getVersion());
+        update.setParameter("metadata", aspect.getMetadata());
+        update.setParameter("createdOn", aspect.getCreatedOn());
+        update.setParameter("createdBy", aspect.getCreatedBy());
+        update.setParameter("oldCreatedOn", new Timestamp(oldAuditStamp.getTime()));
+
+        if (_server.execute(update) != 1) {
+          throw new OptimisticLockException();
```
RealChrisL, line 362: Do you want to put messages/aspects in the OptimisticLockException to provide the details?

keremsahin1: Thanks, done

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
keremsahin1: After some refactoring, I don't think this is relevant any more. Actually, your comment was a trigger for the refactor :)

keremsahin1: This wasn't the main motivation for this test, I only want to test if I got OptimisticLockingException here. Other than that, all tests in the DAO is executed for both optimistic locking is enabled and disabled.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -66,14 +68,16 @@
   private static final int INDEX_QUERY_TIMEOUT_IN_SEC = 5;
   private static final String EBEAN_MODEL_PACKAGE = EbeanMetadataAspect.class.getPackage().getName();
   private static final String EBEAN_INDEX_PACKAGE = EbeanMetadataIndex.class.getPackage().getName();
+  private static final String DEFAULT_COLUMN_FOR_OPTIMISTIC_LOCK = CREATED_ON_COLUMN;
```
jywadhwani, line 71: Curious, could there be a scenario where the column for optimistic locking is different from `created_on` column? If so, should we expose a constructor to provision that?
This might be beyond the scope of this change, but maybe worth addressing in future?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -174,6 +178,20 @@ public void setUseUnionForBatch(boolean useUnionForBatch) {
...
+   * Determines whether we should use optimistic locking in updates.
+   *
+   * <p>The column to execute optimistic locking is controlled by {@link #DEFAULT_COLUMN_FOR_OPTIMISTIC_LOCK}
+   *
+   * <p>DO NOT USE THIS FLAG! This is for LinkedIn use to help us test this feature without a rollback. Once we've
+   * vetted this in production we will be removing this flag and making the the default behavior. So if you set this
+   * to true by calling this method, your code will break when we remove this method. Just wait a bit for us to turn
+   * it on by default!
+   */
+  public void setUseOptimisticLocking(boolean useOptimisticLocking) {
```
jywadhwani, line 191: nit: `setOptimisticLocking` should suffice? don't see `Use` adding any value here.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
keremsahin1: That's a good question. Considering the columns in the DB schema, there is no other column that can be meaningfully used for optimistic locking. `createdOn` is the only one that makes sense.

I guess I wanted to make this as generic as possible if we add a column to the DB schema which is used solely for optimistic locking. But, in that case, I guess using `@version` would be better rather than manually writing our own optimistic locking logic.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
keremsahin1: I guess `setOptimisticLocking` doesn't make sense because this is a setter for the `useOptimisticLocking` class variable. If you also mean class variable should be renamed as `optimisticLocking`, I don't agree with that. But, I can replace `use` with `enable`, if that makes better sense?

jywadhwani: Sure, let's stick with the current naming.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -320,11 +343,61 @@ protected void save(@Nonnull URN urn, @Nonnull RecordTemplate value, @Nonnull Au
...
+
+  // visible for testing
+  protected void saveWithOptimisticLocking(@Nonnull URN urn, @Nonnull RecordTemplate value,
+                                           @Nonnull AuditStamp newAuditStamp,
+                                           long version, boolean insert,
+                                           @Nonnull Object oldTimestamp) {
+
+    final EbeanMetadataAspect aspect = buildMetadataAspectBean(urn, value, newAuditStamp, version);
+
+    if (insert) {
```
jywadhwani, line 362: don't see a case (atleast in this PR) where `saveWithOptimisticLocking` is being called with `true` as `insert` variable. Is that intentional?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -320,11 +343,61 @@ protected void save(@Nonnull URN urn, @Nonnull RecordTemplate value, @Nonnull Au
...
+    //   1. That prevents flag guarding optimistic locking feature
+    //   2. When using @version annotation, Ebean starts to override all updates to that column
+    //      by disregarding any user change.
+    // Ideally, another column for the sake of optimistic locking would be preferred but that means a change to
+    // metadata_aspect schema and we don't take this route here to keep this change backward incompatible.
+    final String updateQuery = String.format("UPDATE metadata_aspect "
+            + "SET urn = :urn, aspect = :aspect, version = :version, metadata = :metadata, createdOn = :createdOn, createdBy = :createdBy "
+            + "WHERE urn = :urn and aspect = :aspect and version = :version and createdOn = :oldTimestamp");
+
+    final SqlUpdate update = _server.createSqlUpdate(updateQuery);
```
jywadhwani, line 379: curious, why did you choose `createSqlUpdate` method over the `update` method of `EbeanServer`?

keremsahin1: `update` takes bean object directly and creates the sql query under the hood but the query it creates is different. It doesn't do optimistic locking. If you want it to do optimistic locking, you need to put `@version` annotation on the `createdOn` column in the bean model. But, we can't use it because of the reasons listed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
keremsahin1: That's true, it's not used right now but this method will be the default going forward after it's vetted. So, it'll be used :)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani: Sure. Using raw SQL query is fine but in certain cases it is better to use the bean object so that the query works for a wide range of databases.

jywadhwani (approved):

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
keremsahin1: Agreed, I also wouldn't prefer raw sql. I don't think there is really a way that I can make ebean generate the sql for me in this case (at least I'm not aware). That being said, this sql query should still be `ansi sql` compliant and I think it's supposedly work for all rdbms.

The only difference compared to the auto-generated query without optimistic locking is the adding of last condition ` and createdOn = :oldTimestamp`.

PR #109, opened 2021-06-23
feat: update derived fields in SCSI in add transaction

kaliang1:
This change allows users to add custom logic to update derived fields in SCSI. The user can pass in a callback function, which returns an array of derived aspects to be updated in the index table.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -254,15 +254,18 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    * monotonically increasing. Older versions of aspect will be purged automatically based on the retention setting. A
    * MetadataAuditEvent is also emitted if there's an actual update.
    *
+   * <p>If SCSI is enabled, the secondary index of the derived aspects will be updated.
+   *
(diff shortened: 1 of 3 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -294,6 +297,12 @@ public boolean isLocalSecondaryIndexEnabled() {
       // 5. Save to local secondary index
       if (_enableLocalSecondaryIndex) {
         updateLocalIndex(urn, newValue, largestVersion);
+
+        // 5.1 Update derived aspects in secondary index
+        List<ASPECT> derivedAspects = getDerivedAspects.apply(urn);
```
jywadhwani, line 302: so this callback function gets called every time an aspect is updated? more specifically does this derived aspect have anything to do with the aspect that is being updated to SCSI?

jywadhwani: I will also encourage you to look at the post-update hook which essentially means that if a particular aspect is updated then some more updates will occur in the DB (following the initial update). It might not happen in the same transaction but in the same method.

Line 326 invokes post-update hook.

kaliang1:
Closing this PR to use the post-update hook method instead.

PR #110, opened 2021-06-24
feat: add offset pagination to SCSI

kaliang1:
This change adds offset pagination to SCSI. A new listUrns function returns a ListResult with pagination information. 

The next step is to add a filter method for offset pagination in BaseEntityResource, which will be done in a separate PR.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -416,7 +416,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    */
   @Nonnull
   public abstract <ASPECT extends RecordTemplate> ListResult<URN> listUrns(@Nonnull IndexFilter indexFilter,
-      @Nullable IndexSortCriterion indexSortCriterion, @Nonnull int start, int pageSize);
+      @Nullable IndexSortCriterion indexSortCriterion, int start, int pageSize);
 
   /**
    * Similar to {@link #listUrns(IndexFilter, Urn, int)}. This is to get all urns with type URN.
@@ -437,7 +437,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    */
   @Nonnull
   private List<UrnAspectEntry<URN>> getUrnAspectEntries(@Nonnull Set<Class<? extends RecordTemplate>> aspectClasses,
-      List<URN> urns) {
+      @Nonnull List<URN> urns) {
     final Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> urnAspectMap =
         get(aspectClasses, new HashSet<>(urns));
 
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -380,8 +380,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    * @param newValue {@link RecordTemplate} of the new value of aspect
    * @param version version of the aspect
    */
-  public abstract <ASPECT extends RecordTemplate> void updateLocalIndex(@Nonnull URN urn,
-      @Nullable ASPECT newValue, long version);
+  public abstract <ASPECT extends RecordTemplate> void updateLocalIndex(@Nonnull URN urn, @Nullable ASPECT newValue,
+      long version);
 
   /**
    * Returns list of urns from local secondary index that satisfy the given filter conditions.
@@ -423,8 +423,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    */
   @Nonnull
   public List<URN> listUrns(@Nonnull Class<URN> urnClazz, @Nullable URN lastUrn, int pageSize) {
-    final IndexFilter indexFilter = new IndexFilter()
-            .setCriteria(new IndexCriterionArray(new IndexCriterion().setAspect(urnClazz.getCanonicalName())));
+    final IndexFilter indexFilter = new IndexFilter().setCriteria(
+        new IndexCriterionArray(new IndexCriterion().setAspect(urnClazz.getCanonicalName())));
     return listUrns(indexFilter, lastUrn, pageSize);
   }
 
@@ -476,7 +476,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    */
   @Nonnull
   public List<UrnAspectEntry<URN>> getAspects(@Nonnull Set<Class<? extends RecordTemplate>> aspectClasses,
-      @Nonnull IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion, @Nullable URN lastUrn, int pageSize) {
+      @Nonnull IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion, @Nullable URN lastUrn,
+      int pageSize) {
 
     final List<URN> urns = listUrns(indexFilter, indexSortCriterion, lastUrn, pageSize);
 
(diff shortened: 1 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1046,10 +1052,23 @@ private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriter
     } else {
       orderByClause = "ORDER BY urn ASC";
     }
-    final String limitClause = "LIMIT ?";
+    final String limitClause = offsetPagination ?  "" : "LIMIT ?";
     return String.join(" ", selectClause, whereClause, orderByClause, limitClause);
   }
 
+  void checkValidIndexCriterionArray(@Nonnull IndexCriterionArray indexCriterionArray) {
+    if (!isLocalSecondaryIndexEnabled()) {
```
jywadhwani, line 1060: this isn't checking the validity on index criterion array. Let's move it out of this method?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1022,7 +1029,6 @@ private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriter
     final StringBuilder whereClause = new StringBuilder("WHERE t0.urn > ?");
```
jywadhwani, line 1029: If we choose to go with offset based pagination then this is not needed? it essentially sets the last urn which is needed in cursor based pagination

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1046,10 +1052,23 @@ private static String constructSQLQuery(@Nonnull IndexCriterionArray indexCriter
     } else {
       orderByClause = "ORDER BY urn ASC";
     }
-    final String limitClause = "LIMIT ?";
+    final String limitClause = offsetPagination ?  "" : "LIMIT ?";
```
jywadhwani, line 1055: not sure about this. whether it is offset or cursor based pagination, we will need to limit the results, right?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -923,16 +923,18 @@ static String getValueFromIndexCriterion(@Nonnull IndexCriterion criterion) {
...
    * @param indexSortCriterion {@link IndexSortCriterion} whose values will be used to set parameters in query
    * @param indexQuery {@link Query} whose ordered parameters need to be set, based on it's position
    * @param lastUrn string representation of the urn whose value is used to set the last urn parameter in index query
    * @param pageSize maximum number of distinct urns to return which is essentially the LIMIT clause of SQL query
+   * @param offsetPagination used to determine whether to used cursor or offset pagination
    */
   private static void setParameters(@Nonnull IndexCriterionArray indexCriterionArray, @Nullable IndexSortCriterion indexSortCriterion,
-      @Nonnull Query<EbeanMetadataIndex> indexQuery, @Nonnull String lastUrn, int pageSize) {
+      @Nonnull Query<EbeanMetadataIndex> indexQuery, @Nonnull String lastUrn, int pageSize, boolean offsetPagination) {
     indexQuery.setParameter(1, lastUrn);
```
jywadhwani, line 938: we shouldn't really be setting the last urn (even if it is empty string) for offset based pagination.

jywadhwani (changes requested):

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
kaliang1: yes, we still need to limit the results. I did it [here](https://github.com/linkedin/datahub-gma/pull/110/files#diff-b5517a7dd054ca33f211c1dddc406b0bebbb4ae5a30cca17c59bff073b221488R1150) because to use `findPagedList()`, the `maxRows` had to be specified.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -407,6 +407,17 @@ public boolean isLocalSecondaryIndexEnabled() {
...
+  /**
+   * Similar to {@link #listUrns(IndexFilter, IndexSortCriterion, Urn, int)} but returns a list result with pagination
+   * information.
+   *
+   * @param start the starting offset of the page
+   * @return a {@link ListResult} containing a list of urns and other pagination information
+   */
+  @Nonnull
+  public abstract <ASPECT extends RecordTemplate> ListResult<URN> listUrns(@Nonnull IndexFilter indexFilter,
+      @Nullable IndexSortCriterion indexSortCriterion, @Nonnull int start, int pageSize);
```
jywadhwani, line 419: you do not need `Nonnull` with int type

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -928,18 +928,25 @@ static String getValueFromIndexCriterion(@Nonnull IndexCriterion criterion) {
...
    * @param lastUrn string representation of the urn whose value is used to set the last urn parameter in index query
    * @param pageSize maximum number of distinct urns to return which is essentially the LIMIT clause of SQL query
+   * @param offsetPagination used to determine whether to used cursor or offset pagination
    */
   private static void setParameters(@Nonnull IndexCriterionArray indexCriterionArray, @Nullable IndexSortCriterion indexSortCriterion,
-      @Nonnull Query<EbeanMetadataIndex> indexQuery, @Nonnull String lastUrn, int pageSize) {
-    indexQuery.setParameter(1, lastUrn);
-    int pos = 2;
+      @Nonnull Query<EbeanMetadataIndex> indexQuery, @Nonnull String lastUrn, int pageSize, boolean offsetPagination) {
+    int pos;
```
jywadhwani, line 943: you could simply initialize pos=1 (as default) and get rid of else?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1104,24 +1131,55 @@ void addEntityTypeFilter(@Nonnull IndexFilter indexFilter) {
...
+    final Query<EbeanMetadataIndex> query = _server
+        .findNative(EbeanMetadataIndex.class, constructSQLQuery(indexCriterionArray, indexSortCriterion, true))
+        .setTimeout(INDEX_QUERY_TIMEOUT_IN_SEC);
+    setParameters(indexCriterionArray, indexSortCriterion, query, "", pageSize, true);
+
+    final PagedList<EbeanMetadataIndex> pagedList = query
+        .setFirstRow(start)
+        .setMaxRows(pageSize)
+        .findPagedList();
+
```
jywadhwani, line 1179: can we load the total row count in background thread like https://ebean.io/docs/query/findPagedList

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1104,24 +1128,57 @@ void addEntityTypeFilter(@Nonnull IndexFilter indexFilter) {
...
+    final List<URN> urns =
+        pagedList.getList().stream().map(entry -> getUrn(entry.getUrn())).collect(Collectors.toList());
+
+    pagedList.loadCount();
```
jywadhwani, line 1180: should we load it before retrieving the list of urns (i.e. before line 1177)?

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -106,6 +106,12 @@ protected void save(FooUrn urn, RecordTemplate value, AuditStamp auditStamp, lon
       return null;
     }
 
+    @Override
+    public ListResult<FooUrn> listUrns(@Nonnull IndexFilter indexFilter,
+        @Nullable IndexSortCriterion indexSortCriterion, int start, int pageSize) {
+      return null;
```
jywadhwani, line 112: isn't the return type `Nonnull` in the base class that we are overriding? can we return an empty list here instead?

jywadhwani (approved):

PR #113, opened 2021-06-29
feat: add in filter criterion

kaliang1:
This change allows users to filter by a string array using the IN operator in SCSI. A user can specify an array and the aspect to filter on.

## Checklist

### DIFF
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexValue.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexValue.pdl
@@ -4,6 +4,7 @@ namespace com.linkedin.metadata.query
  * A union of all supported value types in the index table
  */
 typeref IndexValue = union[
+  array[string],
   boolean,
   double
   float,
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -964,6 +964,16 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   }
 
   @Nonnull
+  private static void validateConditionAndValue(@Nonnull IndexCriterion criterion) {
+    final Condition condition = criterion.getPathParams().getCondition();
+    final IndexValue indexValue = criterion.getPathParams().getValue();
+
+    if (condition == Condition.IN && !indexValue.isArray()) {
+      throw new IllegalArgumentException("Invalid condition " + condition + " for index value " + indexValue);
+    }
+  }
+
+  @Nonnull
   static <ASPECT extends RecordTemplate> String getSortingColumn(@Nonnull IndexSortCriterion indexSortCriterion) {
     final String[] pathSpecArray = RecordUtils.getPathSpecAsArray(indexSortCriterion.getPath());
 
@@ -1041,6 +1051,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
 
       whereClause.append(" AND t").append(i).append(".aspect = ?");
       if (criterion.getPathParams() != null) {
+        validateConditionAndValue(criterion);
         whereClause.append(" AND t")
             .append(i)
             .append(".path = ? AND t")
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -43,6 +43,7 @@ import java.lang.reflect.Method;
 import java.net.URISyntaxException;
 import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -998,6 +1013,17 @@ private static String getStringForOperator(@Nonnull Condition condition) {
     }
   }
 
+  private static String getPlaceholderStringForValue(@Nonnull IndexValue indexValue) {
+    if (indexValue.isArray()) {
```
jywadhwani, line 1017: we could make use of the condition here instead i.e. if the condition is `IN` then follow the steps?

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexValue.pdl
```diff
@@ -4,6 +4,7 @@ namespace com.linkedin.metadata.query
  * A union of all supported value types in the index table
  */
 typeref IndexValue = union[
+  array[string],
```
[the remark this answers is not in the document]
jywadhwani: Looks like you are checking if the field value is present in the given array. We can use this type, in that case.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -907,6 +909,9 @@ static GMAIndexPair getGMAIndexPair(@Nonnull IndexCriterion criterion) {
     } else if (indexValue.isString()) {
       object = getValueFromIndexCriterion(criterion);
       return new GMAIndexPair(EbeanMetadataIndex.STRING_COLUMN, object);
+    } else if (indexValue.isArray()) {
```
jywadhwani, line 912: if in future we start supporting an array of integers (say), then this will break? let us also check if the array elements are of type string as well?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani, line 1017: again, just checking for an array will not suffice, I think.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -958,6 +963,17 @@ private static String getStringForOperator(@Nonnull Condition condition) {
     return CONDITION_STRING_MAP.get(condition);
   }
 
+  @Nonnull
+  private static void validateConditionAndValue(@Nonnull IndexCriterion criterion) {
+    final Condition condition = criterion.getPathParams().getCondition();
+    final IndexValue indexValue = criterion.getPathParams().getValue();
+
+    if (condition == Condition.IN && (!indexValue.isArray() || indexValue.getArray().size() == 0
```
jywadhwani, line 971: I think it should be fine to just check that the IN condition is combined with an array as a value? This might get pretty big (and possibly ugly as we support an array of different types). Would love to hear your thoughts!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -998,6 +1014,17 @@ private static String getStringForOperator(@Nonnull Condition condition) {
     }
   }
 
+  private static String getPlaceholderStringForValue(@Nonnull IndexValue indexValue) {
+    if (indexValue.isArray() && indexValue.getArray().size() > 0 && indexValue.getArray().get(0).getClass() == String.class) {
```
jywadhwani, line 1018: Sorry for the confusion with my previous comments, but if the steps you follow in creating a placeholder string are valid for an array of _any_ primitive type then we don't need to check for String data type. What do you think?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -991,6 +992,44 @@ public void testSorting() {
     assertEquals(urns5, Arrays.asList(urn3, urn2, urn1));
   }
 
+  public void testIn() {
```
jywadhwani, line 995: Let us also test the behavior when _IN_ is combined with a value that is not an array. Also when the array is empty.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani: This should be fine.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
kaliang1: I agree, I think it should be ok to just check those conditions

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
kaliang1: That makes sense!

jywadhwani (approved):

PR #115, opened 2021-07-13
feat: add aggregation count to SCSI

kaliang1:
This change adds a new method to the DAO and resource to count aggregations. The aspect and path to be grouped/aggregated on can be passed in, and the count of the urns for each grouped value is returned. The returned value is a map of the field value to the count.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -24,6 +24,7 @@ import com.linkedin.metadata.query.ExtraInfo;
 import com.linkedin.metadata.query.IndexCriterion;
 import com.linkedin.metadata.query.IndexCriterionArray;
 import com.linkedin.metadata.query.IndexFilter;
+import com.linkedin.metadata.query.IndexGroupByCriterion;
 import com.linkedin.metadata.query.IndexSortCriterion;
 import java.time.Clock;
 import java.util.Collections;
@@ -753,6 +754,16 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   @Nonnull
   public abstract <ASPECT extends RecordTemplate> ListResult<ASPECT> list(@Nonnull Class<ASPECT> aspectClass, int start,
       int pageSize);
+
+  /**
+   *  Gets the count of an aggregation specified by the aspect and field to group on.
+   * @param indexFilter {@link IndexFilter} that defines the filter conditions
+   * @param indexGroupByCriterion {@link IndexGroupByCriterion} that defines the aspect to group by
+   * @return map of the field to the count
+   */
+  @Nonnull
+  public abstract Map<String, Long> countAggregate(@Nonnull IndexFilter indexFilter,
+      @Nonnull IndexGroupByCriterion indexGroupByCriterion);
 
   /**
    * Batch retrieves metadata aspects along with {@link ExtraInfo} using multiple {@link AspectKey}s.
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
@@ -1,0 +1,17 @@
+namespace com.linkedin.metadata.query
+
+/**
+ * Aspect and path to group by, to be applied to the results.
+ */
+record IndexGroupByCriterion {
+  /**
+   * FQCN of the aspect class in the index table that this criterion refers to e.g. com.linkedin.common.Status
+   */
+  aspect: string
+
+  /**
+   * Corresponding path column of the index table that this criterion refers to e.g. /removed (corresponding to field "removed" of com.linkedin.common.Status aspect)
+   */
+  path: string
+}
+
--- a/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
+++ b/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
@@ -132,7 +132,7 @@ public class BaseLocalDAOTest {
 
     @Override
     public Map<String, Long> countAggregate(@Nonnull IndexFilter indexFilter, @Nonnull IndexGroupByCriterion groupCriterion) {
-      return null;
+      return Collections.emptyMap();
     }
 
     @Override
--- a/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
+++ b/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
@@ -7,6 +7,7 @@ import com.linkedin.metadata.dao.retention.TimeBasedRetention;
 import com.linkedin.metadata.dao.retention.VersionBasedRetention;
 import com.linkedin.metadata.query.ExtraInfo;
 import com.linkedin.metadata.query.IndexFilter;
+import com.linkedin.metadata.query.IndexGroupByCriterion;
 import com.linkedin.metadata.query.IndexSortCriterion;
 import com.linkedin.testing.AspectFoo;
 import com.linkedin.testing.EntityAspectUnion;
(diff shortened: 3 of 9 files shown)


dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -129,6 +130,11 @@ protected void save(FooUrn urn, RecordTemplate value, AuditStamp auditStamp, lon
       return null;
     }
 
+    @Override
+    public Map<String, Long> countAggregate(@Nonnull IndexFilter indexFilter, @Nonnull IndexGroupByCriterion groupCriterion) {
```
jywadhwani, line 134: The original method this is overriding seems to be returning `NonNull` value. Let's return an empty map here instead?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1180,4 +1186,103 @@ void addEntityTypeFilter(@Nonnull IndexFilter indexFilter) {
...
+    setCountAggregateParameters(indexCriterionArray, indexGroupByCriterion, query);
+
+    final List<EbeanMetadataIndex> list = query.setDistinct(true).findList();
+    Map<String, Long> result = new HashMap<>();
+    list.stream().map(entry -> {
+      if (entry.getStringVal() != null) {
+        return result.put(entry.getStringVal(), entry.getTotalCount());
+      } else if (entry.getDoubleVal() != null) {
+        return result.put(entry.getDoubleVal().toString(), entry.getTotalCount());
+      } else {
```
jywadhwani, line 1282: Let us explicitly check for the `getLongVal` to be non-null. That way if in future we add new columns then this code does not break.

jywadhwani (changes requested):

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
```diff
@@ -0,0 +1,17 @@
...
+record IndexGroupByCriterion {
+  /**
+   * FQCN of the aspect class in the index table that this criterion refers to e.g. com.linkedin.common.Status
+   */
+  aspect: string
+
+  /**
+   * Corresponding path column of the index table that this criterion refers to e.g. /removed (corresponding to field "removed" of com.linkedin.common.Status aspect)
+   */
+  path: string
```
jywadhwani, line 15: should this be optional? Do you foresee a case where we may want to group by just the aspect name?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -754,6 +755,16 @@ protected abstract void save(@Nonnull URN urn, @Nonnull RecordTemplate value, @N
...
 
+  /**
+   *  Gets the count of an aggregation specified by the aspect and field to group on.
+   * @param indexFilter {@link IndexFilter} that defines the filter conditions
+   * @param indexGroupByCriterion {@link IndexGroupByCriterion} that defines the aspect to group by
+   * @return map of the field to the count
+   */
+  @Nonnull
+  public abstract Map<String, Long> countAggregate(@Nonnull IndexFilter indexFilter,
+      @Nonnull IndexGroupByCriterion indexGroupByCriterion);
```
jywadhwani, line 766: While running SQL query, we can group by multiple columns. This method has provision for grouping only one such column? Do you think we should extend this to multiple columns?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -2078,6 +2071,92 @@ public void testGetWithKeysCount() {
...
+
+    // group by string
+    IndexValue indexValue1 = new IndexValue();
+    indexValue1.setString("val");
+    IndexCriterion criterion1 = new IndexCriterion().setAspect(aspect1)
+        .setPathParams(new IndexPathParams().setPath("/value").setValue(indexValue1).setCondition(Condition.START_WITH));
+
+    IndexCriterionArray indexCriterionArray1 = new IndexCriterionArray(Collections.singletonList(criterion1));
+    final IndexFilter indexFilter1 = new IndexFilter().setCriteria(indexCriterionArray1);
+    IndexGroupByCriterion indexGroupByCriterion1 = new IndexGroupByCriterion().setAspect(AspectFoo.class.getCanonicalName())
```
jywadhwani, line 2110: let us also test the case when `IndexGroupByCriterion` uses an aspect and path that none of the filtered results contain?

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
jywadhwani: Something like: there are 10 filtered results that satisfy `IndexFilter` condition, `aspectA` exists in 8 of them, `aspectB` in 5 of them, etc.

kaliang1: I think this could potentially be a use case in the future? To group by aspects, perhaps the path can be optional and for the aspects string, we could pass in "*" to indicate that we want to group by all of the aspects. What do you think?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kaliang1: Sure, I think this could be a use case in the future.

kaliang1: Actually, I'm not sure about how this would look like. For grouping by multiple columns, would that return the count of columns that have the same value for multiple columns? For example, if we group by X, Y, this would put aspects with the same values for both X and Y in one group and return the count?
Also, the returned value may be a nested map, which could get complicated with multiple group by conditions?

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
jywadhwani: We could as well have a `Nullable` `IndexGroupByCriterion`? where null value would mean what you are suggesting?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -980,11 +981,12 @@ private static void validateConditionAndValue(@Nonnull IndexCriterion criterion)
   }
 
   @Nonnull
-  static <ASPECT extends RecordTemplate> String getSortingColumn(@Nonnull IndexSortCriterion indexSortCriterion) {
-    final String[] pathSpecArray = RecordUtils.getPathSpecAsArray(indexSortCriterion.getPath());
+  static <ASPECT extends RecordTemplate> String getFieldColumn(@Nonnull String stringPath,
```
jywadhwani, line 984: nit: I would just name these variables as `path` and `aspect` (`string` as prefix doesn't add much value, I think)

kaliang1: For https://github.com/linkedin/datahub-gma/pull/115/files#diff-b5517a7dd054ca33f211c1dddc406b0bebbb4ae5a30cca17c59bff073b221488R989, it uses the variable `aspect` and I think `stringAspect` helps to differentiate them?

dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexGroupByCriterion.pdl
jywadhwani: Let's just skip it for now. We can implement it later. This thread will help with tracking.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jywadhwani: Sounds good. Let us skip this functionality for now.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani: how about we use `aspectName` then?

jywadhwani (approved):

PR #116, opened 2021-07-13
feat: add filter for offset pagination

kaliang1:
This change adds a new filter method for offset pagination in BaseEntityResource. It is a follow-up to https://github.com/linkedin/datahub-gma/pull/110. 

## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
@@ -393,6 +393,7 @@ public abstract class BaseEntityResource<
    * takes in a start offset and returns a list result with pagination information.
    *
    * @param start defining the paging start
+   * @param count defining the maximum number of values returned
    * @return a {@link ListResult} containing a list of version numbers and other pagination information
    */
   @Nonnull
@@ -442,6 +443,7 @@ public abstract class BaseEntityResource<
    * takes in a start offset and returns a list result with pagination information.
    *
    * @param start defining the paging start
+   * @param count defining the maximum number of values returned
    * @return a {@link ListResult} containing an ordered list of values of multiple entities and other pagination information
    */
   @Nonnull
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
@@ -8,6 +8,7 @@ import com.linkedin.data.template.UnionTemplate;
 import com.linkedin.metadata.backfill.BackfillMode;
 import com.linkedin.metadata.dao.AspectKey;
 import com.linkedin.metadata.dao.BaseLocalDAO;
+import com.linkedin.metadata.dao.ListResult;
 import com.linkedin.metadata.dao.UrnAspectEntry;
 import com.linkedin.metadata.dao.utils.ModelUtils;
 import com.linkedin.metadata.query.IndexCriterion;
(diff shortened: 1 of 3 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/RestliConstants.java
```diff
@@ -5,6 +5,7 @@ private RestliConstants() { }
 
   public static final String FINDER_SEARCH = "search";
   public static final String FINDER_FILTER = "filter";
+  public static final String FINDER_FILTER_OFFSET_PAGINATION = "filterOffsetPagination";
```
jywadhwani, line 8: `filterOffset` should suffice I think.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -452,6 +519,34 @@ private IndexFilter getDefaultIndexFilter() {
...
 
+  /**
+   * Similar to {@link #filter(IndexFilter, IndexSortCriterion, String[], String, int)} but
+   * returns a list result with pagination information.
+   *
+   * @param pagingContext {@link PagingContext} defines the paging start and count
+   * @return {@link ListResult} containing values along with the associated urns in {@link ListResultMetadata} and
+   *        pagination information
+   */
+  @Finder(FINDER_FILTER_OFFSET)
```
keremsahin1, line 530: I believe I had a similar comment earlier related to another method. Let's keep all filter finders named simply as `filter`?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -397,6 +437,33 @@ private IndexFilter getDefaultIndexFilter() {
...
+  /**
+   * Similar to {@link #filterUrns(IndexFilter, IndexSortCriterion, String, int)} but
+   * takes in a paging context and returns a list result with pagination information.
+   *
+   * @param pagingContext {@link PagingContext} defines the paging start and count
+   * @return a {@link ListResult} containing an ordered list of values of multiple entities and other pagination information
+   */
+  @Nonnull
+  private ListResult<VALUE> filterUrns(@Nonnull IndexFilter filter, @Nullable IndexSortCriterion indexSortCriterion,
+      @Nonnull PagingContext pagingContext) {
```
keremsahin1, line 449: Better to terminate `pagingContext` use at the actual resource method and take `start` and `count` as separate params in this method?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -452,6 +519,34 @@ private IndexFilter getDefaultIndexFilter() {
...
+      @QueryParam(PARAM_SORT) @Optional @Nullable IndexSortCriterion indexSortCriterion,
+      @QueryParam(PARAM_ASPECTS) @Optional @Nullable String[] aspectNames,
+      @PagingContextParam @Nonnull PagingContext pagingContext) {
+
+    final IndexFilter filter = indexFilter == null ? getDefaultIndexFilter() : indexFilter;
+
+    return RestliUtils.toTask(() -> {
+      final Set<Class<? extends RecordTemplate>> aspectClasses = parseAspectsParam(aspectNames);
+      if (aspectClasses.isEmpty()) {
+        return filterUrns(filter, indexSortCriterion, pagingContext);
```
keremsahin1, line 543: In a couple of other places, we treat this optional aspect list param as this:
1. if aspect list is not present => return all aspects
2. if it's empty => return no aspects
3. if there is a list => return that list of aspects as long as they're present

Seems like we're following a different approach here in that case 1 and 2 behaves the same way. This might be intentional but not directly clear from the code. cc @jywadhwani

kaliang1: I think it should be following the same approach here too. In [parseAspectsParam](https://github.com/linkedin/datahub-gma/pull/116/files#diff-ef2014126659a7045b6b1d4c97c7ae54d5fa9d61142c8093ccce3aca8da7c9e3R542), I think it returns all aspects if the aspect list is not present. In the other [finder filter](https://github.com/linkedin/datahub-gma/blob/master/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java#L416), it does that too

keremsahin1: Ah you're definitely right :)

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -390,18 +390,18 @@ private IndexFilter getDefaultIndexFilter() {
 
   /**
    * Similar to {@link #filterAspects(Set, IndexFilter, IndexSortCriterion, String, int)} but
-   * takes in a paging context and returns a list result with pagination information.
+   * takes in a start offset and returns a list result with pagination information.
    *
-   * @param pagingContext {@link PagingContext} defines the paging start and count
+   * @param start defining the paging start
```
keremsahin1, line 395: doc missing for `count`

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -439,17 +439,16 @@ private IndexFilter getDefaultIndexFilter() {
 
   /**
    * Similar to {@link #filterUrns(IndexFilter, IndexSortCriterion, String, int)} but
-   * takes in a paging context and returns a list result with pagination information.
+   * takes in a start offset and returns a list result with pagination information.
    *
-   * @param pagingContext {@link PagingContext} defines the paging start and count
+   * @param start defining the paging start
```
keremsahin1, line 444: doc missing for `count`

keremsahin1 (approved):
LGTM

Issue #114, opened 2021-07-12
Support "CONTAIN" filter condition


jjoyce0510:
#### Describe the bug A clear and concise description of what the bug is.
Currently, the "CONTAIN" filter condition is not supported in Search, because the `getQueryBuilderFromCriterion` in `SearchUtils` class explicitly does not support it.

```  public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion criterion) {
    final Condition condition = criterion.getCondition();
    if (condition == Condition.EQUAL) {
      if (criterion.getValue().startsWith("urn:li:")) {
        return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim());
      }
      return QueryBuilders.termsQuery(criterion.getField(), criterion.getValue().trim().split("\\s*,\\s*"));
    } else if (condition == Condition.GREATER_THAN) {
      return QueryBuilders.rangeQuery(criterion.getField()).gt(criterion.getValue().trim());
    } else if (condition == Condition.GREATER_THAN_OR_EQUAL_TO) {
      return QueryBuilders.rangeQuery(criterion.getField()).gte(criterion.getValue().trim());
    } else if (condition == Condition.LESS_THAN) {
      return QueryBuilders.rangeQuery(criterion.getField()).lt(criterion.getValue().trim());
    } else if (condition == Condition.LESS_THAN_OR_EQUAL_TO) {
      return QueryBuilders.rangeQuery(criterion.getField()).lte(criterion.getValue().trim());
    }

    throw new UnsupportedOperationException("Unsupported condition: " + condition);
  }```

(even though it exists in the Filter Condition enum.


#### To Reproduce

Steps to reproduce the behavior:

1. Deploy DataHub
2. Issue a Search Query with a specific "filter" criteria that has the condition "CONTAIN". You'll see a server error returned. 

![image](https://user-images.githubusercontent.com/17549204/125294210-efdd4080-e2d8-11eb-8a0d-143f87c93453.png)


#### Expected behavior

Contains operator should work for substring of string fields. 

As reported by Lal Rishav at Saxo Bank!

#### Screenshots If applicable, add screenshots to help explain your problem.

#### Desktop (please complete the following information):

- OS: [e.g. iOS]
- Browser [e.g. chrome, safari]
- Version [e.g. 22]

#### Additional context

Add any other context about the problem here.

PR #117, opened 2021-07-14
fix: add 'CONTAIN' support in query builder

yangyangv2:
This fix to address the issue https://github.com/linkedin/datahub-gma/issues/114

## Checklist

### DIFF
--- a/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
+++ b/dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
@@ -60,6 +60,10 @@ public class SearchUtils {
    * <p>If the condition between a field and value is not the same as EQUAL, a Range query is constructed. This
    * condition does not support multiple values for the same field.
    *
+   * <p>When CONTAIN, START_WITH and END_WITH conditions are used, the underlying logic is using wildcard query which is
+   * not performant according to ES. For details, please refer to:
+   * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html#wildcard-query-field-params
+   *
    * @param criterion {@link Criterion} single criterion which contains field, value and a comparison operator
    */
   @Nonnull
@@ -78,6 +82,15 @@ public class SearchUtils {
       return QueryBuilders.rangeQuery(criterion.getField()).lt(criterion.getValue().trim());
     } else if (condition == Condition.LESS_THAN_OR_EQUAL_TO) {
       return QueryBuilders.rangeQuery(criterion.getField()).lte(criterion.getValue().trim());
+    }  else if (condition == Condition.CONTAIN) {
+      return QueryBuilders.wildcardQuery(criterion.getField(),
+          "*" + ESUtils.escapeReservedCharacters(criterion.getValue().trim()) + "*");
+    } else if (condition == Condition.START_WITH) {
+      return QueryBuilders.wildcardQuery(criterion.getField(),
+          ESUtils.escapeReservedCharacters(criterion.getValue().trim()) + "*");
+    } else if (condition == Condition.END_WITH) {
+      return QueryBuilders.wildcardQuery(criterion.getField(),
+          "*" + ESUtils.escapeReservedCharacters(criterion.getValue().trim()));
     }
 
     throw new UnsupportedOperationException("Unsupported condition: " + condition);
--- a/dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
+++ b/dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
@@ -68,6 +68,10 @@ public class SearchUtils {
    * <p>If the condition between a field and value is not the same as EQUAL, a Range query is constructed. This
    * condition does not support multiple values for the same field.
    *
+   * <p>When CONTAIN, START_WITH and END_WITH conditions are used, the underlying logic is using wildcard query which is
+   * not performant according to ES. For details, please refer to:
+   * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html#wildcard-query-field-params
+   *
    * @param criterion {@link Criterion} single criterion which contains field, value and a comparison operator
    */
   @Nonnull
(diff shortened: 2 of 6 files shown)


dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -78,6 +78,8 @@ public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion crite
       return QueryBuilders.rangeQuery(criterion.getField()).lt(criterion.getValue().trim());
     } else if (condition == Condition.LESS_THAN_OR_EQUAL_TO) {
       return QueryBuilders.rangeQuery(criterion.getField()).lte(criterion.getValue().trim());
+    } else if (condition == Condition.CONTAIN) {
+      return QueryBuilders.termQuery(criterion.getField(), criterion.getValue().trim());
```
keremsahin1, line 82: I don't think term query is fitting CONTAIN use case, match query is a better fit.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html

jywadhwani: My thoughts on this:
First let us define what the desired behavior is. Does `CONTAIN` search for a substring of a word, or only a prefix? If the field that the `Criterion` model contains has been analyzed such that it indexes those substring tokens then match query should suffice, but what if the field doesn't?
Here is an example of using prefix based queries in the search request vs. indexing those tokens (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html)
The former approach is slow, but is also agnostic to the mapping that exists.

keremsahin1: I think prefix is another condition `START_WITH`.

yangyangv2: @keremsahin1  it's a good call out. It's should be match query

@jywadhwani, there're also `START_WITH` and `END_WITH` defined in the `Criterion`, which we can fit `prefix query` or `suffix query` respectively. I'm thinking we just treat `CONTAIN` as simple text query as general understanding, and use 'prefix' and 'suffix query' as other Criterion, what do you think? Besides, I'm also confirming this behavior with the issue reporter, still waiting for response.

jywadhwani: Looks like the reporter expects the desired behavior as "Contains operator should work for substring of string fields. " as mentioned in the issue itself.

yangyangv2: Just updated the PR, can you take a look again? @jywadhwani @keremsahin1 . Thanks!

keremsahin1:
There seems to be something off here. Why do I see diff from @kaliang1 in this PR?

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
keremsahin1: I though you changed this to `match` query @yangyangv2 ?

yangyangv2: you're right. there was a duplicated implementation... Just updated the PR.

dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -88,8 +88,9 @@ public static QueryBuilder getQueryBuilderFromCriterion(@Nonnull Criterion crite
       return QueryBuilders.rangeQuery(criterion.getField()).lt(criterion.getValue().trim());
     } else if (condition == Condition.LESS_THAN_OR_EQUAL_TO) {
       return QueryBuilders.rangeQuery(criterion.getField()).lte(criterion.getValue().trim());
+    } else if (condition == Condition.CONTAIN) {
+      return QueryBuilders.matchQuery(criterion.getField(), criterion.getValue().trim());
```
jywadhwani, line 92: should this be a `wildcard` query instead? https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
```diff
@@ -60,6 +60,10 @@ private SearchUtils() {
    * <p>If the condition between a field and value is not the same as EQUAL, a Range query is constructed. This
    * condition does not support multiple values for the same field.
    *
+   * <p>When CONTAIN, START_WITH and END_WITH conditions are used, the underlying logic is using wildcard query which is
+   * not performant according to ES. For details, please refer to:
+   * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html#wildcard-query-field-params
```
camelliazhang, line 65: LGTM in general; Regarding performance, how do we ensure this is not stressing the server (in some extreme cases)?

camelliazhang (commented):
also, the current query template supports advanced query as well, so that user can just issue *text? https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
yangyangv2: it's a known issue that wildcast query is running known. For extreme case consideration, we could add a timeout for wildcard query and include it into the SLA. 

At the moment, this is only requested from OSS, we can just push the update and keep an eye on the feedback how it performs in real-life cases.

dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/utils/SearchUtils.java
yangyangv2: yes, according to this [decision tree](https://www.elastic.co/blog/find-strings-within-strings-faster-with-the-new-elasticsearch-wildcard-field), wildcard is the right choice here.

yangyangv2:
> also, the current query template supports advanced query as well, so that user can just issue *text? https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

Good to know! It could be good for advance users.

This specific ask is for supporting CONTAIN with sbustring query, we prefer to have explicit filter options (CONTAIN, START_WITH, END_WITH) to users and hide ES internal details such as '*'. If users knows they are looking for START_WITH, then using START_WITH will perform better as the underlying query pattern is _wildcard_query(term*)_. 

Exposing '*' to users may lead to complexity and out of controlled use cases such as searching on "*A*B*C*D*"..etc.

yangyangv2:
> also, the current query template supports advanced query as well, so that user can just issue *text? https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

keremsahin1 (approved):

PR #118, opened 2021-07-26
fix: add in filter and use action param for countAggregate

kaliang1:
This PR fixes the following issues with countAggregate: 
- In the SQL query, the array for the IN condition was not being set properly
- The action was using `QueryParam` instead of `ActionParam`

## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1214,6 +1214,7 @@ private static String constructCountAggregateSQLQuery(@Nonnull IndexCriterionArr
       }
       whereClause.append(" t").append(i).append(".aspect = ?");
       if (criterion.getPathParams() != null) {
+        validateConditionAndValue(criterion);
```
jywadhwani, line 1217: Probably missed it earlier, let's move this outside, we can check for `getParams` to be non-null in `validateConditionAndValue` method itself? This way you will fail early.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -560,8 +560,8 @@ private IndexFilter getDefaultIndexFilter() {
   @Action(name = ACTION_COUNT_AGGREGATE)
```
jywadhwani, line 560: I wonder why we didn't use a `Finder` method and return `CollectionResult<V, MetaData>`? I understand we are returning the count along with the field, but can we capture this map in the corresponding metadata?
Would love to hear your thoughts.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
kaliang1: I think the `pathParams` can be null in the index criterion: https://github.com/linkedin/datahub-gma/blob/master/dao-api/src/main/pegasus/com/linkedin/metadata/query/IndexCriterion.pdl#L16

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
kaliang1: That makes a lot of sense! I deprecated the old action and created a new finder.

dao-api/src/main/pegasus/com/linkedin/metadata/query/CountAggregateMetadata.pdl
```diff
@@ -0,0 +1,11 @@
+namespace com.linkedin.metadata.query
+
+/**
+ * The model for count aggregate metadata
+ */
+record CountAggregateMetadata {
```
jywadhwani, line 6: how about we just call this `MapMetadata` - then we can reuse this for other purposes and not just aggregated results?

dao-api/src/main/pegasus/com/linkedin/metadata/query/CountAggregateMetadata.pdl
```diff
@@ -0,0 +1,11 @@
...
+  /**
+   * Map of the field values to the count
+   */
+  countAggregateMap: map[string, long]
```
jywadhwani, line 10: rename this field to something more generic, like `longMap` or something?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jywadhwani: Yes but we can check for non-null value inside `validateConditionAndValue` method itself?

jywadhwani: We can address this separately.

jywadhwani (approved):

PR #119, opened 2021-07-27
feat: add support for sub fields in ES Browse query

shreybatra:
This is add support for filtering on subfield, which may have a different analyzer.
For example - filter `browsePaths.edge_search = value` will now be possible as
`browsePaths` is available in model but not `browsePaths.edge_search`.

## Checklist

### DIFF
--- a/dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
+++ b/dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
@@ -122,8 +122,9 @@ public class ESBrowseDAO extends BaseBrowseDAO {
    * @param isGroupQuery true if it's group query false otherwise
    * @return {@link QueryBuilder}
    */
+  @VisibleForTesting
   @Nonnull
-  private QueryBuilder buildQueryString(@Nonnull String path, @Nonnull Map<String, String> requestMap,
+   QueryBuilder buildQueryString(@Nonnull String path, @Nonnull Map<String, String> requestMap,
       boolean isGroupQuery) {
     final String browsePathFieldName = _config.getBrowsePathFieldName();
     final String browseDepthFieldName = _config.getBrowseDepthFieldName();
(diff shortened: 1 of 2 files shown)


dao-impl/elasticsearch-dao/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
```diff
@@ -145,7 +145,8 @@ private QueryBuilder buildQueryString(@Nonnull String path, @Nonnull Map<String,
     }
 
     requestMap.forEach((field, val) -> {
-      if (_config.hasFieldInSchema(field)) {
+      String[] fieldParts = field.split("\\.");
+      if (_config.hasFieldInSchema(fieldParts[0])) {
```
jywadhwani, line 149: any way we can add tests for this?

jywadhwani: also this restricts to one subfield? what if we have multiple?

shreybatra: This does not restrict to any number, we can have a long string, but the root field would only be present in the schema config right? So as long as `parts[0]` is present, it should work.

I will check for tests, there is none already present for it.

shreybatra: Added the testcases @jywadhwani .

jywadhwani: Thanks for adding the test case. 

I am not clear how user friendly the interface is.
Are you suggesting that the client (like DataHub app) should be aware of the subfields that exist in search index and pass them in the request map as input?

jywadhwani (changes requested):

jywadhwani:
Do we still need this PR?

shreybatra:
I don't think so, let me close this.! Thanks

PR #121, opened 2021-08-03
refactor(dao/utils): Move createRelationshipFilter from Neo4jUtil to QueryUtils

EnricoMi:
These methods are not Neo4j-specific so they can be moved out of Neo4jUtil. This removes Neo4jUtil imports from classes (in [datahub](https://github.com/linkedin/datahub)), that have no other reference to neo4j.

Besides moving these two methods into QueryUtils, this renames them to match existing methods in the target utils class.

Methods in Neo4jUtil are marked as deprecated.

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/utils/QueryUtils.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/utils/QueryUtils.java
@@ -7,6 +7,9 @@ import com.linkedin.metadata.query.Condition;
 import com.linkedin.metadata.query.Criterion;
 import com.linkedin.metadata.query.CriterionArray;
 import com.linkedin.metadata.query.Filter;
+import com.linkedin.metadata.query.RelationshipDirection;
+import com.linkedin.metadata.query.RelationshipFilter;
+
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
@@ -54,6 +57,34 @@ public class QueryUtils {
   @Nonnull
   public static Filter newFilter(@Nonnull String field, @Nonnull String value) {
     return newFilter(Collections.singletonMap(field, value));
+  }
+
+
+  /**
+   * Create {@link RelationshipFilter} using filter and relationship direction.
+   *
+   * @param filter {@link Filter} filter
+   * @param relationshipDirection {@link RelationshipDirection} relationship direction
+   * @return RelationshipFilter
+   */
+  @Nonnull
+  public static RelationshipFilter newRelationshipFilter(@Nonnull Filter filter,
+      @Nonnull RelationshipDirection relationshipDirection) {
+    return new RelationshipFilter().setCriteria(filter.getCriteria()).setDirection(relationshipDirection);
+  }
+
+  /**
+   * Create {@link RelationshipFilter} using filter conditions and relationship direction.
+   *
+   * @param field field to create a filter on
+   * @param value field value to be filtered
+   * @param relationshipDirection {@link RelationshipDirection} relationship direction
+   * @return RelationshipFilter
+   */
+  @Nonnull
+  public static RelationshipFilter newRelationshipFilter(@Nonnull String field, @Nonnull String value,
+      @Nonnull RelationshipDirection relationshipDirection) {
+    return newRelationshipFilter(newFilter(field, value), relationshipDirection);
   }
 
   /**
--- a/dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
+++ b/dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
@@ -6,8 +6,6 @@ import com.linkedin.metadata.dao.utils.RecordUtils;
 import com.linkedin.metadata.query.Condition;
 import com.linkedin.metadata.query.CriterionArray;
 import com.linkedin.metadata.query.Filter;
-import com.linkedin.metadata.query.RelationshipDirection;
-import com.linkedin.metadata.query.RelationshipFilter;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -20,8 +18,6 @@ import org.apache.commons.lang3.ClassUtils;
 import org.neo4j.driver.types.Node;
 import org.neo4j.driver.types.Path;
 import org.neo4j.driver.types.Relationship;
-
-import static com.linkedin.metadata.dao.utils.QueryUtils.*;
 
 
 public class Neo4jUtil {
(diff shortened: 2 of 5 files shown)


EnricoMi:
@jplaisted @gabe-lyons what do you think?

EnricoMi:
@jjoyce0510 this is required by linkedin/datahub#3010

jplaisted:
I'm no longer on this project, this seems like a good idea to break the coupling here. Since this a breaking change maybe bump the semi major version here

https://github.com/linkedin/datahub-gma/blob/master/version.properties

EnricoMi:
> maybe bump the semi major version here

@jplaisted Thanks for the pointer. You probably mean the minor version like `0.2.*` -> `0.3.*`.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
```diff
@@ -248,31 +244,4 @@ public static String getTypeOrEmptyString(@Nullable Class<? extends RecordTempla
...
-   * @param field field to create a filter on
-   * @param value field value to be filtered
-   * @param relationshipDirection {@link RelationshipDirection} relationship direction
-   * @return RelationshipFilter
-   */
-  @Nonnull
-  public static RelationshipFilter createRelationshipFilter(@Nonnull String field, @Nonnull String value,
-      @Nonnull RelationshipDirection relationshipDirection) {
-    return createRelationshipFilter(newFilter(field, value), relationshipDirection);
-  }
```
gabe-lyons, line 277: let's leave this method but mark it deprecated so we don't need to issue a breaking change version bump

version.properties
```diff
@@ -1 +1 @@
-version=0.2.*
+version=0.3.*
```
gabe-lyons, line 1: we'll do a release after merging this PR in @EnricoMi , no need to make this change.

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
EnricoMi: good point, will do that

version.properties
EnricoMi: done

jplaisted: @gabe-lyons gma has a continuous release on every PR. It should be bumped here.

gabe-lyons (approved):

dao-impl/neo4j-dao/src/test/java/com/linkedin/metadata/dao/Neo4jUtilTest.java
```diff
@@ -178,19 +176,6 @@ public void testGetTypeOrEmptyString() {
...
-    String field = "field";
-    String value = "value";
-    RelationshipDirection direction = RelationshipDirection.OUTGOING;
-
-    RelationshipFilter relationshipFilter = new RelationshipFilter().setCriteria(new CriterionArray(
-        Collections.singletonList(new Criterion().setField(field).setValue(value).setCondition(Condition.EQUAL))))
-        .setDirection(direction);
-
-    assertEquals(createRelationshipFilter(field, value, direction), relationshipFilter);
-  }
```
gabe-lyons, line 192: is this intentional?

EnricoMi: Yes, this test moved to `QueryUtilTest`, as the tested method moved from `Neo4jUtil` to `QueryUtils`: https://github.com/linkedin/datahub-gma/pull/121/files#diff-16eb18eb7c23412171efca76506c7c0f4c056e86e5f853c0f070bb2dcc47940bR46-R58

dao-api/src/main/java/com/linkedin/metadata/dao/utils/QueryUtils.java
```diff
@@ -56,6 +59,34 @@ public static Filter newFilter(@Nonnull String field, @Nonnull String value) {
...
+
+  /**
+   * Create {@link RelationshipFilter} using filter and relationship direction.
+   *
+   * @param filter {@link Filter} filter
+   * @param relationshipDirection {@link RelationshipDirection} relationship direction
+   * @return RelationshipFilter
+   */
+  @Nonnull
+  public static RelationshipFilter newRelationshipFilter(@Nonnull Filter filter,
```
gabe-lyons, line 71: why was this rename done? Why did you prefer this name over createRelationshipFilter?

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
```diff
@@ -251,26 +251,30 @@ public static String getType(@Nonnull Class<? extends RecordTemplate> recordClas
...
    * @param field field to create a filter on
    * @param value field value to be filtered
    * @param relationshipDirection {@link RelationshipDirection} relationship direction
    * @return RelationshipFilter
    */
   @Nonnull
+  @Deprecated
   public static RelationshipFilter createRelationshipFilter(@Nonnull String field, @Nonnull String value,
       @Nonnull RelationshipDirection relationshipDirection) {
     return createRelationshipFilter(newFilter(field, value), relationshipDirection);
```
gabe-lyons, line 280: could we update the method bodies to reference QueryUtil's implementations?

dao-api/src/main/java/com/linkedin/metadata/dao/utils/QueryUtils.java
EnricoMi: because all other "create" methods in this class are also called "new"

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
EnricoMi: sure, I was thinking about his, then we make sure future changes to the new methods also "updates" the deprecated methods, will do that

dao-api/src/main/java/com/linkedin/metadata/dao/utils/QueryUtils.java
gabe-lyons: Got it, that makes sense. Thanks!

dao-impl/neo4j-dao/src/main/java/com/linkedin/metadata/dao/Neo4jUtil.java
EnricoMi: done

shirshanka (approved):
LGTM. Thanks!

PR #123, opened 2021-08-04
refactor(test): Add test base class for BaseGraphWriterDAO and BaseQueryDAO

EnricoMi:
Refactoring of `Neo4jGraphWriterDAOTest` and `Neo4jQueryDAOTest` similar to linkedin/datahub/pull/2944. This makes these tests reusable for other implementations of `BaseGraphWriterDAO` and `BaseQueryDAO`.

:information_source: This includes #121, so this requires rebase once that is merged or removal otherwise.


EnricoMi:
@jplaisted @keremsahin1 what are your thoughts on this?

gabe-lyons:
@EnricoMi - this change is OK, but these tests are not run against the new graph services in https://github.com/linkedin/datahub/. If you want to make changes to those graph services, I would focus on updating the tests there. Sorry for the confusion!

EnricoMi:
@gabe-lyons as you said in https://github.com/linkedin/datahub/issues/3084#issuecomment-902047121, these interfaces are not used anymore, there is no benefit of creating a base class as there will be no more implementations. I would suggest you delete those interfaces and implementations. I will try to ignore them :smile:.

PR #124, opened 2021-08-06
feat: track total hits in ES7 if configured

jywadhwani:
The default behavior of ES currently is to _not_ track the total number of hits if more than 10,000 (https://www.elastic.co/guide/en/elasticsearch/reference/7.13/search-your-data.html#track-total-hits) This change does the following
1. Changes the default behavior to return the exact hits
2. Exposes methods to set the lower bound of hits  

## Checklist


dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
```diff
@@ -48,12 +48,35 @@
 public class ESBrowseDAO extends BaseBrowseDAO {
   private final RestHighLevelClient _client;
   private final BaseBrowseConfig _config;
+  private static final Integer ES_LOWER_BOUND_HITS = 10000;
+  private static final Integer DEFAULT_LOWER_BOUND_HITS = 2147483647;
```
keremsahin1, line 52: You mean Integer.MAX_VALUE here?

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
```diff
@@ -48,12 +48,35 @@
...
     this._client = esClient;
     this._config = config;
   }
 
+  /**
+   * Set "track_total_hits" query parameter to false if you do not need accurate results. It is a good trade off to
+   * speed up searches if you don’t need the accurate number of hits after a certain threshold. If set to false, it will
+   * use the default lower bound set by Elasticsearch.
+   */
+  public void setTrackTotalHits(boolean trackTotalHits) {
```
keremsahin1, line 65: Maybe you can get rid of this method and keep the other one. Any user wants to limit max hits could set it accordingly.

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
jywadhwani: This is what ES had set it automatically when using `"track_total_hits": true` in search query. Your suggestion can be incorporated.

dao-impl/elasticsearch-dao-7/src/main/java/com/linkedin/metadata/dao/browse/ESBrowseDAO.java
jywadhwani: Sure

keremsahin1 (approved):
LGTM. Thanks for the quick fix!

PR #126, opened 2021-09-09
feat: add pre-update hooks to BaseLocalDAO

jsdonn:
Introduce infrastructure for pre-update hooks, motivated by the need for a pre-update hook to remove duplicate elements from certain aspects.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -83,7 +83,11 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   // Maps an aspect class to the corresponding retention policy
   private final Map<Class<? extends RecordTemplate>, Retention> _aspectRetentionMap = new HashMap<>();
 
-  // Maps as aspect class to a list of post-update hooks
+  // Maps an aspect class to a list of pre-update hooks
+  private final Map<Class<? extends RecordTemplate>, List<BiConsumer<Urn, RecordTemplate>>> _aspectPreUpdateHooksMap =
+      new HashMap<>();
+
+  // Maps an aspect class to a list of post-update hooks
   private final Map<Class<? extends RecordTemplate>, List<BiConsumer<Urn, RecordTemplate>>> _aspectPostUpdateHooksMap =
       new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -155,6 +159,30 @@ public void setClock(@Nonnull Clock clock) {
...
   }
 
+  /**
+   * Registers a pre-update hook for a specific aspect.
+   *
+   * <p>The hook will be invoked with the latest value of an aspect before it's updated. There's no guarantee on the
+   * order of invocation when multiple hooks are added for a single aspect. Adding the same hook again will result in
+   * {@link IllegalArgumentException} thrown. Hooks are invoked in the order they're registered.
+   */
+  public <URN extends Urn, ASPECT extends RecordTemplate> void addPreUpdateHook(@Nonnull Class<ASPECT> aspectClass,
```
jywadhwani, line 169: there is bunch of code that seems repeated between this method and `addPostUpdateHook`. Can we move some of the common code out and reuse for both the hooks?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -276,24 +304,30 @@ public boolean isLocalSecondaryIndexEnabled() {
...
       final ASPECT newValue = updateLambda.apply(Optional.ofNullable(oldValue));
       checkValidAspect(newValue.getClass());
+
       if (_modelValidationOnWrite) {
         validateAgainstSchema(newValue);
       }
 
-      // 2. Skip saving if there's no actual change
+      // 2. Invoke pre-update hooks, if any
+      if (_aspectPreUpdateHooksMap.containsKey(aspectClass)) {
```
jywadhwani, line 313: any reason you chose to add pre-update hook inside the transaction? The longer the transaction, the longer (in terms of duration) will be the locks on DB rows which might not be ideal.
Consider moving this out of the transaction, just like post-update hook?

jywadhwani (changes requested):

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: discussed offline. the pre-update hook needs to be within the transaction as it requires access to newValue, which is computed within the transaction.

jywadhwani (approved):

PR #57, opened 2020-12-02
fix: actually fix commas in URN search queries not working

jplaisted:
We split on commas in a few places for some reason, not sure why. I've added more TODOs to stop the comma parsing.

#51 

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #128, opened 2021-09-29
fix: issue with commas in filters of search queries with ES7

jywadhwani:
## Checklist

Similar fix was done for ES5 was made here https://github.com/linkedin/datahub-gma/pull/57
 

keremsahin1 (approved):
I guess we should be able to abstract this logic from the DAOs and share between 5 and 7.

jywadhwani:
> I guess we should be able to abstract this logic from the DAOs and share between 5 and 7.

I guess eventually we want to get rid of the ES5 utils/files once we have a plan to migrate OSS users as well from 5->7

PR #138, opened 2021-11-12
fix: use DELETED string literal for null metadata

jywadhwani:
In the PR https://github.com/linkedin/datahub-gma/pull/130, we made the `metadata` column nullable to support soft deletion of aspects. Since changing/altering a column in MySQL to make it nullable entails a good amount of downtime (for large sized tables), we will rather use a placeholder string `DELETED` in this case to denote a soft deleted aspect. All APIs/interfaces in DAO will stay the same, only while saving in DB via EbeanLocalDAO we will store `DELETED` string when inserting a `null` metadata. Similarly while retrieving `DELETED` string from DB, we will interpret it as `null` metadata.
 
## Checklist


nagarjunakanamarlapudi (approved):
LGTM

PR #139, opened 2021-11-17
feat(soft-deleted-aspect): support for adding metadata after soft deleting an aspect

jywadhwani:
This change provides support to add metadata for an aspect after soft deleting that aspect. Soft deleted aspect is just another version of metadata and should not be treated the same way as no existing aspect, even though the return type of a `get` request is the same for an aspect that doesn't exist and an aspect that has been soft deleted.
Also adding more tests testing behavior of different APIs in `EbeanLocalDAO` corresponding to soft deleted aspects. 

Also changing the behavior of following methods in `BaseLocalDAO` and correspondingly `EbeanLocalDAO` to support soft deletion of aspects.
1. `listUrns` method that returns urns containing a given aspect, will not include an urn whose latest version has been soft deleted
2. Similarly `list` method will not return metadata corresponding to soft deleted aspect.
3. Also `listVersions` method will returns versions of an aspect, not including versions of soft deleted aspects.

## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -59,6 +59,23 @@
...
+   */
+  @Value
+  static class AspectMetadata<ASPECT extends RecordTemplate> {
+    @Nullable
+    ASPECT aspect;
+
+    @Nullable
+    ExtraInfo extraInfo;
+
+    boolean isSoftDeleted;
```
RealChrisL, line 76: Can this flag be part of the definition in the [ExtraInfo](https://github.com/linkedin/datahub-gma/blob/1a66e495b5915b4bb58d05f7f7da025bc2c56605/dao-api/src/main/pegasus/com/linkedin/metadata/query/ListResultMetadata.pdl)?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -59,6 +59,23 @@
...
+  }
+
   @Value
   static class AspectEntry<ASPECT extends RecordTemplate> {
```
jsdonn, line 80: Is it possible to get rid of this class as it has been made redundant due to AspectMetadata?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jywadhwani: I wouldn't touch `ExtraInfo` model since this is being used in `ListResultMetadata` and correspondingly `ListResult` model - changing these models will change the API contract with DAO as many public methods exposed in local dao return `ListResult`. I can create a separate model similar to `ExtraInfo` (let's call it `MoreInfo`) that has another field `isSoftDeleted`. I don't have strong opinion here (on whether creating a new model is better or keeping it the current way). What are your thoughts @RealChrisL ?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jywadhwani: I think that's a valid suggestion, assuming we can have a default value for `isSoftDeleted`. Do you think we can address it in another PR, to restrict the scope of this PR (and not make reviews more difficult)? Let me know if that is acceptable to you, I will create an issue and work on it in the next PR.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
RealChrisL: That's also what I thought, but ideally, `ExtraInfo` literally should contain all other Infos associated with the AspectMetadata :). Yet, I assume the flag will be specifically used in the localDAO, so there is no need to create a model wrapper (unless you might have more config need to pass). Keep current way lgtm!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Yes, that sounds good to me. Thanks for addressing.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jywadhwani: Agree. Thanks for the input! Resolving this thread.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jywadhwani: Will create an issue if this PR is successfully merged. Thanks for the input.

jsdonn (approved):
Hopefully others get a chance to review as well, but LGTM.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
keremsahin1: Do we really need this field? Doesn't aspect field being null already mean that this aspect is soft deleted?

jywadhwani: aspect field could be null if no version of the aspect existed in metadata_aspect table. That's why we need a new field to distinguish the two scenarios - 1) no metadata exists already 2) corresponds to soft deleted aspect
aspect will be null for both the scenarios.

keremsahin1: I see, thanks for the explanation.

jywadhwani: I was initially concerned if an OSS datahub-gma user has implemented their own DAO - they might refer to `AspectEntry` variable (similar to our internal espresso DAO). However if we set a default value for isSoftDeleted flag, then this change will be ok.
Thanks for the suggestion, added the flag to `AspectEntry`.

keremsahin1 (commented):
LGTM otherwise

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -813,7 +829,8 @@ public boolean isLocalSecondaryIndexEnabled() {
       int start, int pageSize);
 
   /**
-   * Paginates over all versions of an aspect for a specific Urn.
+   * Paginates over all versions of an aspect for a specific Urn. It does not return metadata corresponding to versions
```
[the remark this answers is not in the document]
jywadhwani: Thanks for asking this. The major concern/challenge with returning metadata corresponding to soft deleted aspects is the return type (`ListResult` model https://github.com/linkedin/datahub-gma/blob/master/dao-api/src/main/java/com/linkedin/metadata/dao/ListResult.java contains a list of values `List<T> values` - i.e. each item in the list of metadata returned is expected to be non-null).

Thoughts?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -292,10 +292,10 @@ public static ServerConfig createTestingH2ServerConfig() {
   @Override
   protected <ASPECT extends RecordTemplate> long saveLatest(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass,
       @Nullable ASPECT oldValue, @Nullable AuditStamp oldAuditStamp, @Nullable ASPECT newValue,
-      @Nonnull AuditStamp newAuditStamp) {
+      @Nonnull AuditStamp newAuditStamp, boolean isDeleted) {
```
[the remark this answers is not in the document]
jywadhwani: Sure, I can do that. Thanks for the careful review.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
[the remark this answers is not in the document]
keremsahin1: Isn't it allowed to add `null` to the list?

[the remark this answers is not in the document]
keremsahin1: My only concern with this is the inconsistency between the two APIs. Let's say I get all versions of an aspect as {0, 1, 2} and version 1 being "DELETED"/null aspect using the other API. This API will only return two elements corresponding to versions 0, 2 and I don't have a way to really tell which is which.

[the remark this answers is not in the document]
jywadhwani: Passing `null` to the list is allowed, I think. But, I would rather use `List<Optional<T>>` than `List<T>` to avoid null pointer exceptions in the method(s) using this returned list of results - however changing the return type might not be ideal.
To answer your question on how to understand which version the metadata corresponds to, there is actually `ListResultMetadata` inside `ListResult` that returns array of versions corresponding to the array of metadata/values returned. So if you get metadata corresponding to versions 0 and 2, `extraInfos` field inside `ListResultMetadata` will return the versions as 0 and 2 for the metadata returned.

I understand your concern on inconsistency between `listVersions` method returning all versions, including soft deleted aspect, while other APIs/methods are not. I can change the behavior of `listVersions` to also _not_ return versions corresponding to soft deleted aspect. Will that work for you?

I am open to other suggestions you may have.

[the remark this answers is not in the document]
keremsahin1: Thanks for the detailed response. I agree with your proposal. "Changing the behavior of listVersions to also not return versions corresponding to soft deleted aspect" makes sense to me.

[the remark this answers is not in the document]
jywadhwani: Added. Thanks for the feedback.

PR #142, opened 2021-12-17
[fix] support alias in unions

jywadhwani:
While setting aspects in aspect union, you could use aliases for different aspect classes. However currently GMA throws error when a union is defined using aliases. This change fixes it.

## Checklist


sacsar (commented):
Thank you for fixing it so fast.

RealChrisL (approved):

PR #144, opened 2022-01-18
fix: properly validate that aspects can only be of type record

jywadhwani:
This change will ensure that aspects, relationships and entities can only be of type record. We assume this in our DAO, but currently don't validate the same.

## Checklist

### DIFF
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/dummy/DummyRecord.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/dummy/DummyRecord.pdl
@@ -3,4 +3,7 @@ namespace com.linkedin.metadata.dummy
 /**
  * Not to be used
  */
-typeref DummyAspect = union[DummyRecord]
+record DummyRecord {
+
+  value: string
+}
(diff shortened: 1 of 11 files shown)


sacsar:
Should we add tests making sure we reject enums now? (And maps/arrays, for good measure?)

It probably makes sense to explicitly reject `UnionTemplate` here, too, so we can fail early when someone is like me and defines the record for the aspect and only remembers to add it to the aspect union later.

jywadhwani:
> Should we add tests making sure we reject enums now? (And maps/arrays, for good measure?)

Added, thanks for the suggestion

> It probably makes sense to explicitly reject `UnionTemplate` here, too, so we can fail early when someone is like me and defines the record for the aspect and only remembers to add it to the aspect union later.

I agree that one can define an aspect as a union, and later realize that they cannot add it to aspect union. Currently all our validation is on the aspect union which this PR is trying to address.

I will see where we can validate the aspect alone and have a separate PR for it. Thanks!

RealChrisL (approved):

PR #158, opened 2022-05-05
fix: use json to store deleted aspect metadata

jywadhwani:
## Checklist

Instead of storing a simple text/reserved keyword to represent soft deleted aspect, let us use json to represent the same.

### DIFF
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
@@ -2219,6 +2219,7 @@ public class EbeanLocalDAOTest {
     // latest version of metadata should be null
     EbeanMetadataAspect aspect = getMetadata(urn, aspectName, 0);
     assertEquals(aspect.getMetadata(), EbeanLocalDAO.DELETED_VALUE);
+    assertEquals(EbeanLocalDAO.DELETED_VALUE, "{\"GMA_DELETED\":true}");
     Optional<AspectFoo> fooOptional = dao.get(AspectFoo.class, urn);
     assertFalse(fooOptional.isPresent());
 
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
@@ -2219,7 +2219,7 @@ public class EbeanLocalDAOTest {
     // latest version of metadata should be null
     EbeanMetadataAspect aspect = getMetadata(urn, aspectName, 0);
     assertEquals(aspect.getMetadata(), EbeanLocalDAO.DELETED_VALUE);
-    assertEquals(EbeanLocalDAO.DELETED_VALUE, "{\"GMA_DELETED\":true}");
+    assertEquals(EbeanLocalDAO.DELETED_VALUE, "{\"gma_deleted\":true}");
     Optional<AspectFoo> fooOptional = dao.get(AspectFoo.class, urn);
     assertFalse(fooOptional.isPresent());
 
--- a/dao-api/build.gradle
+++ b/dao-api/build.gradle
@@ -10,6 +10,7 @@ dependencies {
   compile externalDependency.commonsLang
 
   dataModel project(':core-models')
+  dataModel project(':validators')
 
   compileOnly externalDependency.lombok
 
--- a/dao-api/src/test/java/com/linkedin/metadata/dao/utils/ModelUtilsTest.java
+++ b/dao-api/src/test/java/com/linkedin/metadata/dao/utils/ModelUtilsTest.java
@@ -3,6 +3,7 @@ package com.linkedin.metadata.dao.utils;
 import com.google.common.collect.ImmutableSet;
 import com.linkedin.common.CommonTestAspect;
 import com.linkedin.common.urn.Urn;
+import com.linkedin.testing.AspectUnionWithSoftDeletedAspect;
 import com.linkedin.testing.DeltaUnionAlias;
 import com.linkedin.testing.EntityAspectUnionAliasArray;
 import com.linkedin.testing.EntityDeltaAlias;
@@ -77,6 +78,12 @@ public class ModelUtilsTest {
     Set<Class<? extends RecordTemplate>> validTypes = ModelUtils.getValidAspectTypes(EntityAspectUnion.class);
 
     assertEquals(validTypes, ImmutableSet.of(AspectFoo.class, AspectBar.class));
+  }
+
+  @Test
+  public void testSoftDeletedAspect() {
+    assertThrows(InvalidSchemaException.class,
+        () -> ModelUtils.getValidAspectTypes(AspectUnionWithSoftDeletedAspect.class));
   }
 
   @Test
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -7,6 +7,7 @@ import com.linkedin.data.schema.DataSchema;
 import com.linkedin.data.schema.RecordDataSchema;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.data.template.UnionTemplate;
+import com.linkedin.metadata.aspect.SoftDeletedAspect;
 import com.linkedin.metadata.dao.exception.ModelConversionException;
 import com.linkedin.metadata.dao.exception.RetryLimitReached;
 import com.linkedin.metadata.dao.producer.BaseMetadataEventProducer;
@@ -61,7 +62,6 @@ import javax.persistence.OptimisticLockException;
 import javax.persistence.RollbackException;
 import javax.persistence.Table;
 import lombok.Value;
-import org.json.simple.JSONObject;
 
 import static com.linkedin.metadata.dao.EbeanMetadataAspect.*;
 
@@ -72,11 +72,9 @@ import static com.linkedin.metadata.dao.EbeanMetadataAspect.*;
 public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     extends BaseLocalDAO<ASPECT_UNION, URN> {
 
-  // String literal stored in metadata_aspect table for soft deleted aspect
-  private static final Map<String, Boolean> DELETED_METADATA = new HashMap<String, Boolean>() {{
-    put("GMA_DELETED", true);
-  }};
-  public static final String DELETED_VALUE = new JSONObject(DELETED_METADATA).toString();
+  // String stored in metadata_aspect table for soft deleted aspect
+  private static final RecordTemplate DELETED_METADATA = new SoftDeletedAspect().setGma_deleted(true);
+  public static final String DELETED_VALUE = RecordUtils.toJsonString(DELETED_METADATA);
 
   private static final int INDEX_QUERY_TIMEOUT_IN_SEC = 5;
   private static final String EBEAN_MODEL_PACKAGE = EbeanMetadataAspect.class.getPackage().getName();
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -61,6 +61,7 @@ import javax.persistence.OptimisticLockException;
 import javax.persistence.RollbackException;
 import javax.persistence.Table;
 import lombok.Value;
+import org.json.simple.JSONObject;
 
 import static com.linkedin.metadata.dao.EbeanMetadataAspect.*;
 
@@ -72,7 +73,10 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     extends BaseLocalDAO<ASPECT_UNION, URN> {
 
   // String literal stored in metadata_aspect table for soft deleted aspect
-  public static final String DELETED_VALUE = "DELETED";
+  private static final Map<String, Boolean> DELETED_METADATA = new HashMap<String, Boolean>() {{
+    put("GMA_DELETED", true);
+  }};
+  public static final String DELETED_VALUE = new JSONObject(DELETED_METADATA).toString();
 
   private static final int INDEX_QUERY_TIMEOUT_IN_SEC = 5;
   private static final String EBEAN_MODEL_PACKAGE = EbeanMetadataAspect.class.getPackage().getName();
--- a/testing/test-models/src/main/pegasus/com/linkedin/testing/AspectUnionWithSoftDeletedAspect.pdl
+++ b/testing/test-models/src/main/pegasus/com/linkedin/testing/AspectUnionWithSoftDeletedAspect.pdl
@@ -1,0 +1,6 @@
+namespace com.linkedin.testing
+
+/**
+ * For unit tests
+ */
+typeref AspectUnionWithSoftDeletedAspect = union[AspectFoo, AspectBar, SoftDeletedAspectCopy]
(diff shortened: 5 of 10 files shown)


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -2219,6 +2219,7 @@ public void testGetSoftDeletedAspect() {
     // latest version of metadata should be null
     EbeanMetadataAspect aspect = getMetadata(urn, aspectName, 0);
     assertEquals(aspect.getMetadata(), EbeanLocalDAO.DELETED_VALUE);
+    assertEquals(EbeanLocalDAO.DELETED_VALUE, "{\"GMA_DELETED\":true}");
```
nagarjunakanamarlapudi, line 2222: We just need to make sure that an aspect is not created with this field like below. Probably document it somewhere, better enforced.

record aspect1 {
GMA_DELETED: string
}

jywadhwani: Thanks for the review. I have used pegasus to model the JSON on soft deleted aspect. I have also enforced it in the validators so that we don't have an aspect with the same form.

keremsahin1:
I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from soft deletion. Am I right on that?

Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted versions?

jywadhwani:
> I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> 
> Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?

Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.

In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?

keremsahin1:
> > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> 
> Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> 
> In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?

I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.

Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?

jywadhwani:
> > > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> > 
> > 
> > Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> > In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?
> 
> I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.
> 
> Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?

So here is how we are implementing it currently. When an aspect is soft deleted, the latest version (i.e. version=0) will store reserved keyword/JSON in `metadata` column to indicate that the latest aspect is soft deleted. You can un-delete by adding metadata in which case the latest version will reflect undo the soft deletion.

If I understand your suggestion correctly, when we soft delete an aspect, we do the following
1. Retrieve v0 of the aspect
2. Retrieve the current max version (Vmax)
3. Write the old value back as (Vmax + 1)
We do NOT perform the following
4. Write the new value back as v0

One challenge with this approach is that we have lost track of which version has been soft deleted from the audit trail?

keremsahin1:
> > > > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > > > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> > > 
> > > 
> > > Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> > > In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?
> > 
> > 
> > I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.
> > Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?
> 
> So here is how we are implementing it currently. When an aspect is soft deleted, the latest version (i.e. version=0) will store reserved keyword/JSON in `metadata` column to indicate that the latest aspect is soft deleted. You can un-delete by adding metadata in which case the latest version will reflect undo the soft deletion.
> 
> If I understand your suggestion correctly, when we soft delete an aspect, we do the following
> 
> 1. Retrieve v0 of the aspect
> 2. Retrieve the current max version (Vmax)
> 3. Write the old value back as (Vmax + 1)
>    We do NOT perform the following
> 4. Write the new value back as v0
> 
> One challenge with this approach is that we have lost track of which version has been soft deleted from the audit trail?

Yeah I thought about that but forgot to mention. To keep the history consistent, we can still log a SOFT_DELETE marker row during soft delete just that we don't have to store this as version 0.

jywadhwani:
> > > > > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > > > > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> > > > 
> > > > 
> > > > Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> > > > In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?
> > > 
> > > 
> > > I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.
> > > Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?
> > 
> > 
> > So here is how we are implementing it currently. When an aspect is soft deleted, the latest version (i.e. version=0) will store reserved keyword/JSON in `metadata` column to indicate that the latest aspect is soft deleted. You can un-delete by adding metadata in which case the latest version will reflect undo the soft deletion.
> > If I understand your suggestion correctly, when we soft delete an aspect, we do the following
> > 
> > 1. Retrieve v0 of the aspect
> > 2. Retrieve the current max version (Vmax)
> > 3. Write the old value back as (Vmax + 1)
> >    We do NOT perform the following
> > 4. Write the new value back as v0
> > 
> > One challenge with this approach is that we have lost track of which version has been soft deleted from the audit trail?
> 
> Yeah I thought about that but forgot to mention. To keep the history consistent, we can still log a SOFT_DELETE marker row during soft delete just that we don't have to store this as version 0. That means we'll need to update 2 rows during soft delete; Vmax+1 and Vmax+2 (soft delete update).

So we still have a reserved keyword/marker in `metadata` column for Vmax+2 version to indicate that this is a soft deleted aspect. How is it different from having a reserved keyword in `metadata` column for latest version to indicate the same? If someone were to ask for metadata of an aspect corresponding to version Vmax+2 - we will still need to understand that this reserved keyword/json corresponds to soft deleted aspect and enforce it in aspect unions as well so that users don't use the same metadata in a real aspect?

keremsahin1:
> > > > > > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > > > > > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> > > > > 
> > > > > 
> > > > > Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> > > > > In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?
> > > > 
> > > > 
> > > > I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.
> > > > Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?
> > > 
> > > 
> > > So here is how we are implementing it currently. When an aspect is soft deleted, the latest version (i.e. version=0) will store reserved keyword/JSON in `metadata` column to indicate that the latest aspect is soft deleted. You can un-delete by adding metadata in which case the latest version will reflect undo the soft deletion.
> > > If I understand your suggestion correctly, when we soft delete an aspect, we do the following
> > > 
> > > 1. Retrieve v0 of the aspect
> > > 2. Retrieve the current max version (Vmax)
> > > 3. Write the old value back as (Vmax + 1)
> > >    We do NOT perform the following
> > > 4. Write the new value back as v0
> > > 
> > > One challenge with this approach is that we have lost track of which version has been soft deleted from the audit trail?
> > 
> > 
> > Yeah I thought about that but forgot to mention. To keep the history consistent, we can still log a SOFT_DELETE marker row during soft delete just that we don't have to store this as version 0. That means we'll need to update 2 rows during soft delete; Vmax+1 and Vmax+2 (soft delete update).
> 
> So we still have a reserved keyword/marker in `metadata` column for Vmax+2 version to indicate that this is a soft deleted aspect. How is it different from having a reserved keyword in `metadata` column for latest version to indicate the same? If someone were to ask for metadata of an aspect corresponding to version Vmax+2 - we will still need to understand that this reserved keyword/json corresponds to soft deleted aspect and enforce it in aspect unions as well so that users don't use the same metadata in a real aspect?

How do we return the metadata for soft deleted aspects today? Isn't there an issue already when reading soft deleted aspect versions today because metadata schema will not match?

```
ComplianceInfo ci = _dao.read(urn, ComplianceInfo.aspect, version=280)
```

Considering 280 is a soft deleted aspect version, this will fail anyways, right?

jywadhwani:
> > > > > > > I find putting a magic word into json is super super hacky. I vaguely remember we discussed on why a simple empty json "{}" doesn't work in this case. If I remember correctly, that is because there might be an aspect with all optional fields and "{}" might really be the actual value of the aspect and it's hard to distinguish it from a soft deleted aspect. Am I right on that?
> > > > > > > Then, as an another approach, instead of marking metadata with some magic words, can we allocate a special version like "-1" for soft deleted aspects?
> > > > > > 
> > > > > > 
> > > > > > Correct on why we cannot use empty json to represent soft deleted aspect. An aspect with all optional fields and none being populated will be stored as "{}" from my understanding/recollection.
> > > > > > In some way we have enforced (while validating an aspect union) that an aspect cannot look like soft deleted aspect metadata. Do you see any risks with using a reserved JSON to represent soft deleted aspect and enforcing it so that no (real) aspect can look like it?
> > > > > 
> > > > > 
> > > > > I think what you're doing is something workable for sure. But you need to document that and users need to be aware of that limitation while designing their models etc. That's a point of friction. Ideally, one less thing to worry about for users is always good if we can do that, right? And the good thing with versions is that users are not involved with version management but platform handles that automatically for them.
> > > > > Actually thought about this more, we don't even need to introduce a new version. If we get rid of version 0, non-existence of version 0 can mean an aspect is soft deleted?
> > > > 
> > > > 
> > > > So here is how we are implementing it currently. When an aspect is soft deleted, the latest version (i.e. version=0) will store reserved keyword/JSON in `metadata` column to indicate that the latest aspect is soft deleted. You can un-delete by adding metadata in which case the latest version will reflect undo the soft deletion.
> > > > If I understand your suggestion correctly, when we soft delete an aspect, we do the following
> > > > 
> > > > 1. Retrieve v0 of the aspect
> > > > 2. Retrieve the current max version (Vmax)
> > > > 3. Write the old value back as (Vmax + 1)
> > > >    We do NOT perform the following
> > > > 4. Write the new value back as v0
> > > > 
> > > > One challenge with this approach is that we have lost track of which version has been soft deleted from the audit trail?
> > > 
> > > 
> > > Yeah I thought about that but forgot to mention. To keep the history consistent, we can still log a SOFT_DELETE marker row during soft delete just that we don't have to store this as version 0. That means we'll need to update 2 rows during soft delete; Vmax+1 and Vmax+2 (soft delete update).
> > 
> > 
> > So we still have a reserved keyword/marker in `metadata` column for Vmax+2 version to indicate that this is a soft deleted aspect. How is it different from having a reserved keyword in `metadata` column for latest version to indicate the same? If someone were to ask for metadata of an aspect corresponding to version Vmax+2 - we will still need to understand that this reserved keyword/json corresponds to soft deleted aspect and enforce it in aspect unions as well so that users don't use the same metadata in a real aspect?
> 
> How do we return the metadata for soft deleted aspects today? Isn't there an issue already when reading soft deleted aspect versions today because metadata schema will not match?
> 
> ```
> ComplianceInfo ci = _dao.read(urn, ComplianceInfo.aspect, version=280)
> ```
> 
> Considering 280 is a soft deleted aspect version, this will fail anyways, right? Are we doing something smart to make this work?
> 
> PS: I also think that there is probably no practical example of this (reading a non-zero version) but I get it, API supports this use case.

To summarize our discussion:
1. We need to record when an aspect was soft deleted for audit purposes.
2. Even if we store soft deleted aspect as another version (Vmax+2) and not latest version (0), we will still need to exclude those rows from the methods `listUrns` and `listVersions` - whatever JSON we choose to store in `metadata` column of soft deleted aspect shouldn't be a part of any aspect.

Please let me know if you have other thoughts or suggestions. Thank you for the detailed review.

keremsahin1 (approved):
Thanks for bearing with me @jywadhwani

PR #196, opened 2022-08-12
fix(new schema): fix soft-deletion of aspects and remaining new schema unit tests

jsdonn:
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -14,22 +14,17 @@ import io.ebean.EbeanServer;
 import io.ebean.SqlQuery;
 import io.ebean.SqlRow;
 import io.ebean.SqlUpdate;
-import java.math.BigDecimal;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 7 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -129,15 +140,14 @@ public List<URN> listUrns(@Nonnull IndexFilter indexFilter, @Nullable IndexSortC
...
-    TODO: filter out soft-deleted aspects
-    TODO: start/OFFSET in sql query causes _total_count to be inaccurate.
-     (e.g. if there are 3 results but the start/OFFSET is 4, _total_count will be 0 even though it is expected to be 3.
-    */
-    final List<URN> values =
-        sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString("urn"), _urnClass)).collect(Collectors.toList());
+    if (sqlRows.size() == 0) {
+      int actualTotalCount = createFilterSqlQuery(indexFilter, indexSortCriterion, null, 0, DEFAULT_PAGE_SIZE).findList().size();
+      final SqlRow dummyRow = new DummySqlRow(actualTotalCount);
+      return toListResult(Collections.emptyList(), Collections.singletonList(dummyRow), start, pageSize);
```
zhixuanjia, line 148: Just think aloud, is it possible to code another `toListResult(List<T> values, int totalCount, int start, int pageSize)`. It seems to me that only info that's useful in DummySqlRow is the total count. I feel it's not necessary to introduce a new concept/class just for passing in the total count.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -153,7 +153,9 @@ public void setupTest() {
       _server.execute(Ebean.createSqlUpdate(readSQLfromFile(GMA_DROP_ALL_SQL)));
       _server.execute(Ebean.createSqlUpdate(readSQLfromFile(GMA_CREATE_ALL_SQL)));
     } else {
-      _server.execute(Ebean.createSqlUpdate(readSQLfromFile(NEW_SCHEMA_CREATE_ALL_SQL)));
+      synchronized (this) {
+        _server.execute(Ebean.createSqlUpdate(readSQLfromFile(NEW_SCHEMA_CREATE_ALL_SQL)));
+      }
```
zhixuanjia, line 158: Is this synchronization introduced to prevent me from dropping the tables you're creating? :-) I don't think this will helpful since the code is running on different machines.

jsdonn: Oh yeah oops I was thinking maybe it will help (before I realized why we were seeing those weird errors). Forgot to remove, will do so now.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: Probably a better idea than introducing a new class, let me change that.

zhixuanjia (approved):
LGTM, thanks for addressing comments.

PR #199, opened 2022-08-15
fix(new schema): fix compare results methods

jsdonn:
* add .equals() methods to all objects that will be compared in these methods
* add sufficient units tests to ensure dual-read schema will properly catch mismatched results (between new and old schema)

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/ListResult.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/ListResult.java
@@ -40,24 +40,21 @@ public class ListResult<T> {
 
   @Override
   public boolean equals(Object o) {
-    if (o == this) {
-      return true;
-    }
-
-    if (!(o instanceof ListResult)) {
+    if (o == null) {
       return false;
     }
-
+    if (o.getClass() != this.getClass()) {
+      return false;
+    }
     ListResult<T> other = (ListResult<T>) o;
 
     if (this.values.size() != other.values.size()) {
       return false;
     }
 
-    // TODO: this comparison has worst case O(n^2) runtime. find a more efficient way.
-    // TODO: need to add .equals method for all T values possible.
     return this.values.containsAll(other.values)
-        && other.values.containsAll(this.values)
+        // either both metadata fields are null or both are equal (need to check this.metadata != null to avoid NPE)
+        && ((this.metadata == null && other.metadata == null) || (this.metadata != null && this.metadata.equals(other.metadata)))
         && this.nextStart == other.nextStart
         && this.havingMore == other.havingMore
         && this.totalCount == other.totalCount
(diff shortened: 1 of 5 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/ListResult.java
```diff
@@ -40,24 +40,21 @@
...
+    if (o == null) {
+      return false;
     }
-
-    if (!(o instanceof ListResult)) {
+    if (o.getClass() != this.getClass()) {
       return false;
     }
-
     ListResult<T> other = (ListResult<T>) o;
```
zhixuanjia, line 49: nit: Change name `Object o` to `Object other` so this line is not needed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanMetadataAspect.java
```diff
@@ -63,6 +61,23 @@ public static class PrimaryKey {
...
+
+    @Override
+    public boolean equals(Object o) {
+      if (o == null) {
+        return false;
+      }
+      if (o.getClass() != this.getClass()) {
+        return false;
+      }
+      PrimaryKey other = (PrimaryKey) o;
```
zhixuanjia, line 73: Same as above, just use `Object other` so this line is not needed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanMetadataAspect.java
```diff
@@ -84,4 +99,25 @@ public static class PrimaryKey {
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null) {
+      return false;
+    }
+    if (o.getClass() != this.getClass()) {
+      return false;
+    }
+    EbeanMetadataAspect other = (EbeanMetadataAspect) o;
```
zhixuanjia, line 111: Same as above, just use Object other so this line is not needed.

dao-api/src/main/java/com/linkedin/metadata/dao/ListResult.java
jsdonn: Actually, this line is needed when overriding `equals()` methods, in my experience. Otherwise we cannot access methods/fields specific to ListResult since the passed in object is of type Object, not ListResult. When I remove this line and rename the variable, I get "cannot resolve symbol <>" for `other.variable` and `other.method()`.

I see in many examples in stack overflow of this casting being necessary. e.g.
https://stackoverflow.com/questions/28880263/whats-the-correct-way-to-override-equal-method-in-java

zhixuanjia (approved):

PR #206, opened 2022-08-23
fix: replace WITH with MySQL backward compatible syntax

zhixuanjia:
## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -350,16 +350,16 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
       @Nonnull AuditStamp auditStamp, @Nonnull EqualityTester<ASPECT> equalityTester) {
 
     final ASPECT oldValue = latest.getAspect() == null ? null : latest.getAspect();
+    final AuditStamp oldAuditStamp = latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit();
+
     // Skip saving if there's no actual change
-    if ((oldValue == null && newValue == null) || oldValue != null && newValue != null && equalityTester.equals(
-        oldValue, newValue)) {
+    if ((oldValue == null && newValue == null) || oldValue != null && newValue != null
+        && equalityTester.equals(oldValue, newValue)) {
       return new AddResult<>(oldValue, oldValue, aspectClass);
     }
 
     // Save the newValue as the latest version
-    long largestVersion =
-        saveLatest(urn, aspectClass, oldValue, latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit(),
-            newValue, auditStamp, latest.isSoftDeleted);
+    long largestVersion = saveLatest(urn, aspectClass, oldValue, oldAuditStamp, newValue, auditStamp, latest.isSoftDeleted);
 
     // Apply retention policy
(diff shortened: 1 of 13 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -140,15 +133,12 @@ public static <ASPECT extends RecordTemplate> String createAspectUpsertSql(@Nonn
    */
   public static String createFilterSql(String tableName, @Nonnull IndexFilter indexFilter,
       @Nullable IndexSortCriterion indexSortCriterion) {
+    String whereClause = parseIndexFilter(indexFilter);
+    String totalCountSql = String.format("SELECT COUNT(urn) FROM %s %s", tableName, whereClause);
     StringBuilder sb = new StringBuilder();
-    sb.append(String.format(SQL_FILTER_TEMPLATE_START, tableName));
+    sb.append(String.format(SQL_FILTER_TEMPLATE, totalCountSql, tableName));
     sb.append("\n");
-    sb.append(parseIndexFilter(indexFilter));
```
jsdonn, line 146: We're not using indexSortCriterion anymore? Why not?

jsdonn: Nvm I see we append it later .

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -53,21 +53,14 @@ public class SQLStatementUtils {
...
+   *  Filter query has pagination params in the existing APIs. To accommodate this, we use subquery to include total result counts in the query response.
    *  For example, we will build the following filter query statement:
    *
-   *  <p>WITH _temp_results AS (SELECT * FROM metadata_entity_foo
-   *      WHERE
-   *        i_testing_aspectfoo$value >= 25 AND
-   *        i_testing_aspectfoo$value < 50
-   *      ORDER BY i_testing_aspectfoo$value ASC)
-   *    SELECT *, (SELECT count(urn) FROM _temp_results) AS _total_count FROM _temp_results
+   *  <p>SELECT *, (SELECT COUNT(urn) FROM metadata_entity_foo WHERE i_aspectfoo$value >= 25\n"
```
jsdonn, line 59: Why do we need the subquery in the first place? I feel like I am missing something obvious, but doesn't this query also work?

`
SELECT *, COUNT(urn) as _total_count FROM metadata_entity_foo WHERE i_aspectfoo$value >= 25\nAND i_aspectfoo$value < 50 AND a_aspectfoo != '{\"gma_deleted\":true}')
`

jsdonn: Ok I realize why my query doesn't work. But I still feel like there may be a better way instead of having to perform the same where clause filtering twice.

jsdonn: https://stackoverflow.com/a/55708509

After some discussion, it seems that using 2 queries may actually be the preferred way to go. `SQL_CALC_FOUND_ROWS` will be deprecated in MySQL 8.0.17+. There may still be a better way to do this query so let's keep it in mind if we see bad performance during performance testing. But for now, we can move on.

jsdonn (approved):
LGTM if all tests are still passing.

PR #210, opened 2022-08-31
feat(new schema): add a_urn column to help with urn path extraction

jsdonn:
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -7,6 +7,8 @@ import com.linkedin.data.template.SetMode;
 import com.linkedin.metadata.aspect.AuditedAspect;
 import com.linkedin.metadata.dao.builder.BaseLocalRelationshipBuilder;
 import com.linkedin.metadata.dao.builder.LocalRelationshipBuilderRegistry;
+import com.linkedin.metadata.dao.scsi.EmptyPathExtractor;
+import com.linkedin.metadata.dao.scsi.UrnPathExtractor;
 import com.linkedin.metadata.dao.utils.ModelUtils;
 import com.linkedin.metadata.dao.utils.RecordUtils;
 import com.linkedin.metadata.dao.utils.SQLSchemaUtils;
@@ -46,6 +48,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   private final EbeanServer _server;
   private final Class<URN> _urnClass;
   private final String _entityType;
+  private UrnPathExtractor<URN> _urnPathExtractor;
   private final EbeanLocalRelationshipWriterDAO _localRelationshipWriterDAO;
   private LocalRelationshipBuilderRegistry _localRelationshipBuilderRegistry;
 
@@ -56,12 +59,17 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   private static final String ASPECT_JSON_PLACEHOLDER = "__PLACEHOLDER__";
   private static final String DEFAULT_ACTOR = "urn:li:principal:UNKNOWN";
 
-  public EbeanLocalAccess(EbeanServer server, ServerConfig serverConfig, @Nonnull Class<URN> urnClass) {
+  public EbeanLocalAccess(EbeanServer server, ServerConfig serverConfig, @Nonnull Class<URN> urnClass, UrnPathExtractor<URN> urnPathExtractor) {
     _server = server;
     _urnClass = urnClass;
+    _urnPathExtractor = urnPathExtractor;
     _entityType = ModelUtils.getEntityTypeFromUrnClass(_urnClass);
     _localRelationshipWriterDAO = new EbeanLocalRelationshipWriterDAO(_server);
     createSchemaEvolutionManager(serverConfig).ensureSchemaUpToDate();
+  }
+
+  public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor) {
+    _urnPathExtractor = urnPathExtractor;
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -398,6 +406,13 @@ public static String toJsonString(@Nonnull final AuditedAspect auditedAspect) {
     return RecordUtils.toJsonString(auditedAspect).replace("\"" + ASPECT_JSON_PLACEHOLDER + "\"",  aspect);
   }
 
+  @Nonnull
+  public String toJsonString(@Nonnull URN urn) {
```
zhixuanjia, line 410: Can this just be a private method?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -59,14 +61,19 @@
...
+  public EbeanLocalAccess(EbeanServer server, ServerConfig serverConfig, @Nonnull Class<URN> urnClass, UrnPathExtractor<URN> urnPathExtractor) {
     _server = server;
     _urnClass = urnClass;
+    _urnPathExtractor = urnPathExtractor;
     _entityType = ModelUtils.getEntityTypeFromUrnClass(_urnClass);
     _localRelationshipWriterDAO = new EbeanLocalRelationshipWriterDAO(_server);
     createSchemaEvolutionManager(serverConfig).ensureSchemaUpToDate();
   }
 
+  public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor) {
```
zhixuanjia, line 73: is this set method necessary?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -79,6 +86,7 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
 
     final SqlUpdate sqlUpdate = _server.createSqlUpdate(SQLStatementUtils.createAspectUpsertSql(urn, aspectClass))
         .setParameter("urn", urn.toString())
+        .setParameter("a_urn", toJsonString(urn))
```
zhixuanjia, line 89: I believe this makes a_urn a must-have column for all entity tables and I think most of entity tables will have null or some dummy value for this column, since there is no user-defined path extractor. Can we treat it same as aspect column? User needs to `alter table ... add a_urn` and only set this value if urn path extractor is provided.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -316,4 +322,43 @@ public void testAtomicityWithLocalRelationshipBuilder() throws URISyntaxExceptio
...
+
+  @Test
+  public void testUrnExtraction() {
+    FooUrn urn1 = makeFooUrn(1);
+    AspectFoo foo1 = new AspectFoo().setValue("foo");
+    _ebeanLocalAccessFoo.add(urn1, foo1, AspectFoo.class, makeAuditStamp("actor", 1234L));
+
+    // generate a virtual column i_urn$fooId from the a_urn column.
+    // (in regular use cases, this should be part of the schema evolution scripts and not in code)
+    _server.execute(Ebean.createSqlUpdate("ALTER TABLE metadata_entity_foo ADD COLUMN i_urn$fooId VARCHAR(255)\n"
```
zhixuanjia, line 353: Why not just put it in ebean-local-access-create-all.sql?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: That's a fair suggestion. We would need to either enforce that if `ALTER TABLE ...` is in their schema initialization script, they MUST also provide an UrnPathExtractor or somehow catch this kind of error if both don't exist.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
jsdonn: Yes I could also do that lol. Thanks for the suggestion.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: Initially set to public since I can't test it if private, but I do agree it should be private. Since I added another test that covers the urn path extraction case, let me remove this method's test case and make it private.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: Added a comment in EbeanLocalAccess and uploaded user onboarding doc that you must specify either BOTH of NONE of the following:
1) non-default UrnPathExtractor
2) ALTER TABLE table ADD COLUMN a_urn JSON

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -72,13 +80,20 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
     final long timestamp = auditStamp.hasTime() ? auditStamp.getTime() : System.currentTimeMillis();
     final String actor = auditStamp.hasActor() ? auditStamp.getActor().toString() : DEFAULT_ACTOR;
     final String impersonator = auditStamp.hasImpersonator() ? auditStamp.getImpersonator().toString() : null;
+    final boolean urnExtraction = !(_urnPathExtractor instanceof EmptyPathExtractor);
```
zhixuanjia, line 83: do we need to add a null check? `urnExtraction = _urnPathExtractor != null && !(_urnPathExtractor instanceof EmptyPathExtractor);`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -72,13 +80,20 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
...
         .setParameter("urn", urn.toString())
         .setParameter("lastmodifiedon", new Timestamp(timestamp).toString())
         .setParameter("lastmodifiedby", actor);
 
-    // newValue is null if soft-delete aspect.
+    // If a non-default UrnPathExtractor is provided, the user MUST specify in their schema generation scripts
+    // 'ALTER TABLE table ADD COLUMN a_urn JSON'.
+    if (urnExtraction) {
+      sqlUpdate.setParameter("a_urn", toJsonString(urn));
+    }
```
zhixuanjia, line 94: I think we only needs to execute this block if newValue is not null. So can be moved down after `if (newValue == nuill)`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: I don't think we _need_ it but it doesn't hurt to add it. Thanks for the suggestion.

zhixuanjia (approved):

PR #213, opened 2022-09-13
fix(new schema): fix backfill api + minor bug fixes

jsdonn:
bug before:
backfill method calls get() to get aspects to backfill. however, for the new schema, this get() will try to get from the new schema tables, which won't have the existing data in it (hence the whole point of needing to backfill). 

solution:
create separate backfill method that circumvents existing backfill logic.

## Checklist

### DIFF
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/backfill/BackfillMode.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/backfill/BackfillMode.pdl
@@ -16,12 +16,7 @@ enum BackfillMode {
   MAE_ONLY
 
   /**
-   * Backfill only new entity tables
-   */
-  ENTITY_TABLES_ONLY
-
-  /**
-   * Backfill all secondary stores, excluding ENTITY_TABLES_ONLY
+   * Backfill all secondary stores
    */
   BACKFILL_ALL
 }
--- a/dao-impl/ebean-dao/src/test/resources/schema-evolution-create-all.sql
+++ b/dao-impl/ebean-dao/src/test/resources/schema-evolution-create-all.sql
@@ -11,7 +11,7 @@ CREATE TABLE metadata_id (
 );
 
 CREATE TABLE metadata_aspect (
-    urn VARCHAR(1000) NOT NULL,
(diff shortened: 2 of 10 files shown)


dao-impl/ebean-dao/src/test/resources/schema-evolution-create-all.sql
```diff
@@ -18,7 +18,7 @@ CREATE TABLE metadata_aspect (
     createdon TIMESTAMP NOT NULL,
     createdby VARCHAR(255) NOT NULL,
     createdfor VARCHAR(255),
-    CONSTRAINT pk_metadata_aspect_ PRIMARY KEY (urn,aspect,version)
+    CONSTRAINT pk_metadata_aspect_ PRIMARY KEY (urn(50),aspect,version)
```
zhixuanjia, line 21: nit: I know it's just test SQL, but instead of truncating the urn, I would suggest reduce 1000 to 500 if key is too large `urn VARCHAR(500) NOT NULL,`

dao-api/src/main/pegasus/com/linkedin/metadata/backfill/BackfillMode.pdl
```diff
@@ -15,11 +15,6 @@ enum BackfillMode {
    */
   MAE_ONLY
 
-  /**
-   * Backfill only new entity tables
-   */
-  ENTITY_TABLES_ONLY
-
   /**
    * Backfill all secondary stores, excluding ENTITY_TABLES_ONLY
```
zhixuanjia, line 19: Do you need to change the comment here as well?

jsdonn: Yes good catch. Resolved.

dao-impl/ebean-dao/src/test/resources/schema-evolution-create-all.sql
jsdonn: Thanks for the suggestion, resolved.

zhixuanjia (approved):
LGTM.

PR #216, opened 2022-10-19
feat(new schema): add Flyway support for multiple db instances

jsdonn:
Testing:
- change evolution.conf to be gma_dev.conf, which is the name of the DB_SCHEMA as defined in MySQLDevInstance.java
- verify all unit tests with -Ptest-ebean-dao flag are working
- added unit test to ensure database url parsing method works

## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/FlywaySchemaEvolutionManager.java
```diff
@@ -48,4 +49,27 @@ public void ensureSchemaUpToDate() {
...
+    }
+    // First, get the path, which includes the database name. In our example above, it would be "my_first_db?autoReconnect=true"
+    String databaseNameWithOptions = connectionUrl.substring(connectionUrl.lastIndexOf('/') + 1);
+    if (databaseNameWithOptions.contains(":")) {
+      throw new IllegalArgumentException("Please double check that your database connection URL is in the following format:"
+          + "jdbc:mysql://<host>:<port>/<db_name>?<options>");
+    }
+    // Next, strip off any characters after the '?' so that we are left with just the database name. Ex. "my_first_db"
+    return databaseNameWithOptions.contains("?") ? databaseNameWithOptions.split("\\?")[0] : databaseNameWithOptions;
+  }
```
zhixuanjia, line 74: this looks simpler
``` 
 public String test() {
    String url = "jdbc:mysql://example.linkedin.com:1234/my_first_db?options";
    String cleanURI = url.substring(5);
    URI uri = URI.create(cleanURI);
    return uri.getPath().substring(1);
  }
  ```

jsdonn: Tried some of the test cases against this. URI.create("stringwithnospecialcharacters") will actually just return the whole string as the database name even though an error is expected to be thrown. (I know this case should be nonexistent though). 

Also with the substring calls, I will have to check for StringIndexOutOfBoundsExceptions anyways so the code will end up looking similar.

If you don't feel too strongly about changing it, I would suggest to leave it as is. Let me know, thanks.

zhixuanjia: sounds good. As long as the code works

PR #228, opened 2022-12-22
fix(new schema): fix EbeanMetadataAspect metadata field equality check

jsdonn:
Seeing some false negative logs before when comparing old and new schema results because the EbeanMetadataAspect metadata check was a simple String.equals check. However, metadata is JSON in string form so the order of JSON mappings may be different between the different objects. Fixed by comparing JSONObjects, the underlying implementation of JSONObject.equals is AbstractMap.equals which compares all entry sets, ignoring order.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanMetadataAspect.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanMetadataAspect.java
@@ -15,6 +15,9 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import lombok.Setter;
+import lombok.SneakyThrows;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
 
 
 /**
@@ -85,6 +88,7 @@ public class EbeanMetadataAspect extends Model {
   @Column(name = CREATED_FOR_COLUMN, nullable = true)
   private String createdFor;
 
+  @SneakyThrows
   @Override
   public boolean equals(Object o) {
     if (o == null) {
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanMetadataAspect.java
```diff
@@ -94,13 +98,22 @@ public boolean equals(Object o) {
...
+        && ((this.metadata == null && other.metadata == null) || (this.metadata != null && other.metadata != null))
         && this.createdOn.equals(other.getCreatedOn())
         && this.createdBy.equals(other.getCreatedBy())
         // either both createdFor fields are null or both are equal (need to check this.createdFor != null to avoid NPE)
         && ((this.createdFor == null && other.getCreatedFor() == null) || (this.createdFor != null && this.createdFor.equals(other.getCreatedFor())));
+    if (!primitiveEqualityCheck) {
+      return false;
+    }
+
+    JSONParser parser = new JSONParser();
```
yangyangv2, line 113: do we have to instantiate a new JSONParter each time?

jsdonn: Yes, unfortunately. We found earlier that JSONParser is not thread-safe and saw some weird errors when we had defined a single JSONParser as a class variable.

yangyangv2: have you tried ThreadLocal to create per-thread singleton?

private static final ThreadLocal<JSONParser> jsonParser = 
	ThreadLocal.<JSONParser>withInitial
	(() -> new JSONParser());

... ...

jsdonn: Does this change much though? We will still have 1 JSONParser instance for each EbeanMetadataAspect object right? Or you want me to define in another class (EbeanLocalDAO for example)?

yangyangv2: The above logic guarantees the same json parser instance to be used in one thread. If you previously run into thread-safe issue it should avoid it. But If you safe thread-safe issue before, it means EbeanMetadataAspect is used in concurrent environment as well, so if JSONParser was defined as member field, it won't be thread-safe.

I would't recommend going too complicated with this. I just checked the JSONParser doc, instantiation doesn't take much computation. Just FYI the LocalThread approach if you need it in the future.

jsdonn: Thanks for the input, Yang. I will just keep it as is for now then for simplicity. Do you mind resolving this conversation if you're satisfied with the current implementation and giving an approval on the PR?

yangyangv2 (approved):

PR #229, opened 2023-01-12
feat(new schema): add method to backfill local relationships from entity tables

jsdonn:
This method should only be used in cases where data has already been ingested into the entity tables but not the local relationship tables. For example, this can happen when the entity tables were set up before the local relationship tables.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -628,6 +628,17 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   public abstract <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass);
 
   /**
+   * Backfill local relationships from the new schema entity tables. This method is new/dual schema only. It should NOT
+   * be used to add local relationships under normal circumstances and should ONLY be used in the case where the aspects
+   * exist in the entity tables but the relationships don't exist in the local relationship tables (e.g. if the local
+   * relationship tables got dropped or if they were set up after data was already ingested into entity tables).
+   *
+   * @param urn the URN for the entity the aspect (which the local relationship is derived from) is attached to
+   * @param aspectClass class of the aspect to backfill
+   */
+  public abstract <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass);
+
+  /**
    * Returns list of urns from local secondary index that satisfy the given filter conditions.
    *
    * <p>Results are ordered by the order criterion but defaults to sorting lexicographically by the string
--- a/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
+++ b/dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
@@ -72,6 +72,11 @@ public class BaseLocalDAOTest {
 
     }
 
+    @Override
+    public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntityTables(@Nonnull FooUrn urn, @Nonnull Class<ASPECT> aspectClass) {
+
+    }
+
     @Nonnull
     @Override
     protected <T> T runInTransactionWithRetry(Supplier<T> block, int maxTransactionRetry) {
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -94,7 +94,6 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
 
     // newValue is null if aspect is to be soft-deleted.
     if (newValue == null) {
-
       /*
       TODO:
       Local relationship is derived from an aspect. If an aspect metadata is deleted, then the local relationships derived from it
@@ -112,12 +111,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
     }
 
     // Add local relationships if builder is provided.
-    if (_localRelationshipBuilderRegistry != null && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
-      List<BaseLocalRelationshipBuilder<ASPECT>.LocalRelationshipUpdates> localRelationshipUpdates =
-          _localRelationshipBuilderRegistry.getLocalRelationshipBuilder(newValue).buildRelationships(urn, newValue);
-
-      _localRelationshipWriterDAO.processLocalRelationshipUpdates(localRelationshipUpdates);
-    }
+    addRelationships(urn, newValue, aspectClass);
 
     AuditedAspect auditedAspect = new AuditedAspect()
         .setAspect(RecordUtils.toJsonString(newValue))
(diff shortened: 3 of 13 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -112,12 +113,59 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
...
-          _localRelationshipBuilderRegistry.getLocalRelationshipBuilder(newValue).buildRelationships(urn, newValue);
+          _localRelationshipBuilderRegistry.getLocalRelationshipBuilder(relationship).buildRelationships(urn, relationship);
 
       _localRelationshipWriterDAO.processLocalRelationshipUpdates(localRelationshipUpdates);
     }
+  }
+
+  @VisibleForTesting
+  @Transactional
+  public <ASPECT extends RecordTemplate> int addWithoutRelationships(@Nonnull URN urn, @Nullable ASPECT newValue, @Nonnull Class<ASPECT> aspectClass,
```
zhixuanjia, line 147: Why we want to unit test addWithoutRelationships? Is it being used anywhere?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -344,11 +344,28 @@ public <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn,
...
       _localAccess.add(urn, toRecordTemplate(aspectClass, result).orElse(null), aspectClass, auditStamp);
-      return null; // return value not used
+      return null; // unused
+    }, 1);
+  }
+
+  public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
+    if (_schemaConfig == SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("The DAO is set to use the OLD_SCHEMA_ONLY. Please check that the schemaConfig"
+          + "parameter is properly set to NEW_SCHEMA_ONLY or DUAL_SCHEMA if you wish to use local relationship tables.");
```
zhixuanjia, line 358: nit: This sentence makes sense to us. People outside gma not know what schemaConfig param refers to. I would just say `Backfilling local relationship table is not supported under OLD_SCHEMA_ONLY mode.`

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -72,6 +72,11 @@ public <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull FooUrn u
 
     }
 
+    @Override
+    public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntityTables(@Nonnull FooUrn urn, @Nonnull Class<ASPECT> aspectClass) {
+
+    }
+
```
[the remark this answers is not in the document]
jsdonn: Dummy method for DummyLocalDAO class that is within BaseLocalDAOTest.java. Since I added it as an abstract method to BaseLocalDAO so it needs to be overwritten by this dummy class.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -627,6 +627,18 @@ public abstract <ASPECT extends RecordTemplate> void updateLocalIndex(@Nonnull U
    */
   public abstract <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass);
 
+  /**
+   * Backfill local relationships from the new schema entity tables. This method is new/dual schema only. It should not
+   * be used to add local relationships under normal circumstances, since add() will add both entities and local relationships
```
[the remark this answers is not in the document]
jsdonn: Done

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: Changed to package-protected method.

@VisibleForTesting means that the method is only for testing purposes and should have weaker access privilege otherwise, so I should have had it as public in the first place.

We want to use this method to set up the (test) situation where there is data in the entity tables, but not in the relationship tables. The alternative to writing a method to do this is just to use raw SQL queries, but I am too lazy to do that :P

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: Thanks for the suggestion, updated to make more sense to users outside of our team

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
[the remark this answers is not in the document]
zhixuanjia: Makes sense.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -627,6 +627,19 @@ public abstract <ASPECT extends RecordTemplate> void updateLocalIndex(@Nonnull U
    */
   public abstract <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass);
 
+  /**
+   * Backfill local relationships from the new schema entity tables. This method is new/dual schema only. It should not
+   * be used to add local relationships under normal circumstances, since {@link BaseLocalDAO#add(Urn, Class, Function, AuditStamp, int) add()}
```
zhixuanjia, line 632: I think the add method you want to refer to is the one in EbeanLocalAccess class. So `{@link IEbeanLocalAccess#add(Urn, RecordTemplate, Class, AuditStamp)}` My last comment just mean to be an example.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -47,9 +47,9 @@
 
 public class EbeanLocalAccessTest {
   private static EbeanServer _server;
-  private static IEbeanLocalAccess<FooUrn> _ebeanLocalAccessFoo;
-  private static IEbeanLocalAccess<BarUrn> _ebeanLocalAccessBar;
-  private static IEbeanLocalAccess<BurgerUrn> _ebeanLocalAccessBurger;
+  private static EbeanLocalAccess<FooUrn> _ebeanLocalAccessFoo;
+  private static EbeanLocalAccess<BarUrn> _ebeanLocalAccessBar;
+  private static EbeanLocalAccess<BurgerUrn> _ebeanLocalAccessBurger;
```
zhixuanjia, line 52: Why these need to be changed to concrete implementation?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
zhixuanjia: oh, if that's the case. We can just use `setLocalRelationshipBuilderRegistry` to change builder registery to null and then add entities. This will skip local relationship table. Whenever you want to add to relationship tables, you can just put builder registry back.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Ah, my bad. I got rid of the `addWithoutLocalRelationships` method anyways.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jsdonn: Thanks, I implemented your suggestion. I made the parameter of `setLocalRelationshipBuilderRegistry` to be nullable (it was nonnull before). I think this also makes sense if a user wants to stop ingesting relationship data for whatever reason, so they will have a way to unset the relationship registry instead of having to modify the relationship builder logic. Let me know if you have concerns about that.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -112,12 +113,59 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
...
+        .setCanonicalName(aspectClass.getCanonicalName())
+        .setLastmodifiedby(actor)
+        .setLastmodifiedon(new Timestamp(timestamp).toString())
+        .setCreatedfor(impersonator, SetMode.IGNORE_NULL);
+
+    return sqlUpdate.setParameter("metadata", toJsonString(auditedAspect)).execute();
+  }
+
+  @Override
+  public <ASPECT extends RecordTemplate> void addRelationships(@Nonnull URN urn, @Nonnull ASPECT relationship, @Nonnull Class<ASPECT> aspectClass) {
```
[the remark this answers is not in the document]
jsdonn: done

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
jsdonn: We can drop this, I changed it back to interface.

Issue #136, opened 2021-11-03
Sync get, getAll, batchGet, and search getInternal behavior


jsdonn:
#### Is your feature request related to a problem? Please describe.

[PR#132](https://github.com/linkedin/datahub-gma/pull/132) changes the intended behavior of `get()` when parameter `aspectNames` is empty. Instead of throwing a 404, it will return VALUE with no aspects. To accomplish this behavior, `getInternalNonEmpty` call was changed to `getInternal`. We need to evaluate this behavior with `getAll()`, `batchGet()`, and `search()` methods to make sure they are aligned with `get()`'s new behavior.

#### Describe the solution you'd like

Double check and update calls to getInternalNonEmpty if needed to match with new `get()` behavior.

PR #233, opened 2023-01-27
fix(batchGet): synchronize batchGet() behavior with get()

jsdonn:
To help address issue #136 .

Before:
batchGet would still return results for urns which do not exist in the db.

Now:
batchGet will throw a 404 if all urns passed in do not exist in the db.
If some or all of the urns exist, it will only return results for those that do exist in the db.

## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
@@ -21,8 +21,11 @@ import com.linkedin.metadata.query.ListResultMetadata;
 import com.linkedin.metadata.query.MapMetadata;
 import com.linkedin.parseq.Task;
 import com.linkedin.restli.common.EmptyRecord;
+import com.linkedin.restli.common.HttpStatus;
+import com.linkedin.restli.server.BatchResult;
 import com.linkedin.restli.server.CollectionResult;
 import com.linkedin.restli.server.PagingContext;
+import com.linkedin.restli.server.RestLiServiceException;
 import com.linkedin.restli.server.annotations.Action;
 import com.linkedin.restli.server.annotations.ActionParam;
 import com.linkedin.restli.server.annotations.Finder;
@@ -36,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -164,10 +164,13 @@ public Task<VALUE> get(@Nonnull KEY id,
   public Task<Map<KEY, VALUE>> batchGet(
       @Nonnull Set<KEY> ids,
       @QueryParam(PARAM_ASPECTS) @Optional @Nullable String[] aspectNames) {
-    // TODO: discuss and sync with get()'s intended behavior (check if urn exists). https://github.com/linkedin/datahub-gma/issues/136
     return RestliUtils.toTask(() -> {
       final Map<URN, KEY> urnMap =
-          ids.stream().collect(Collectors.toMap(id -> toUrn(id), Function.identity()));
+          ids.stream().collect(Collectors.toMap(this::toUrn, Function.identity()));
+      urnMap.keySet().removeIf(urn -> !getLocalDAO().exists(urn));
```
zhixuanjia, line 170: This line will be expensive when batchGet 100 urns. Assuming each existence check takes just 10ms, then it's still 1 second in total. Batch get 100 urns is very realistic in my opinion. 

I think there could be two options.
1.  Create a new exists method that can take in a set of keys such as `exists(Set<KEY> keys)`, the in SQL we can do `select urn from table where urn=a or urn=b or urn=c or ...` in which a,b,c ...  are the keys. This will give us all the valid urns.

2. Just remvoe this existence check and directly pass it to DAO. It could generate a `select statement UNION select statement ...` with a bunch invalid urns and nothing returned. But I imagine this case is rare.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -164,10 +164,13 @@ public Task<VALUE> get(@Nonnull KEY id,
   public Task<Map<KEY, VALUE>> batchGet(
```
zhixuanjia, line 164: Can we do something similar the following?
```
  public Task<BatchResult<KEY, VALUE>> batchGetV2(
      @Nonnull Set<KEY> ids,
      @QueryParam(PARAM_ASPECTS) @Optional @Nullable String[] aspectNames) {
    Map<KEY, VALUE> batch = new HashMap<KEY, VALUE>();
    Map<KEY, RestLiServiceException> errors = new HashMap<KEY, RestLiServiceException>();
    for (KEY id : ids)
    {
      VALUE g = getLocalDAO().get(id);
      if (g != null)
      {
        batch.put(id, g);
      }
      else
      {
        errors.put(id, new RestLiServiceException(HttpStatus.S_404_NOT_FOUND));
      }
    }
    return RestliUtils.toTask(() -> new BatchResult<KEY, VALUE>(batch, errors));
  }
```

zhixuanjia: This gives a consistent and more granular view on the batch get result. In my opinion, it's better than throwing a 404 when all not found but no error at when just one record not found.

jsdonn: I agree that this is better. However, this would not be a simple change since it will affect batchGet's downstream users as we will be changing the return type from Task<Map<KEY, VALUE>> to Task<BatchGetResult<Key, VALUE>>. I would argue that we can avoid this hassle by using a cleaner, but maybe not fully "correct" approach. The core thing is that we will not omit 404 under any circumstances so that we don't give potentially misleading informatino. I see two options like:

1) For keys that do not exist in the db, we will just not add them to the map (I think that's what you suggested in the other comment for #2). So if all keys don't exist, we will just return empty map.
2) For keys that do not exist in the db, we can add `(key, null)` to the map. So if all keys don't exist, we will return a non-empty map but all values will be null.

Either way, we will need to form some SLA contract with users on this behavior so they know what to expect with this API. I think 1) is probably better since it's easier to check how many results were returned by the call. WDYT about removing 404s altogether?

jsdonn: The upside is this approach is easier to implement and should not affect any existing downstream users. The downside is it will not be consistent with `get` behavior since we will not throw any 404s. And users need to be aware of the behavior we decide on when they use this API.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -157,21 +161,57 @@ public Task<VALUE> get(@Nonnull KEY id,
...
+      final Map<URN, VALUE> batchResult = getInternal(urnMap.keySet(), parseAspectsParam(aspectNames));
+      batchResult.entrySet().removeIf(entry -> {
+        if (!entry.getValue().data().isEmpty()) {
+          // don't remove if there is a non-empty value associated with the key
+          return false;
+        }
+        // if there is this key's value is empty, then this key doens't exist in the db.
+        // mark this key with 404 and remove the entry from the map
+        errors.put(urnMap.get(entry.getKey()), new RestLiServiceException(HttpStatus.S_404_NOT_FOUND));
+        statuses.put(urnMap.get(entry.getKey()), HttpStatus.S_404_NOT_FOUND);
```
zhixuanjia, line 209: why we need both status and errors? Both seems tell user same thing: 404 not found.

jsdonn: Thanks for pointing that out. Ideally I think users can just use statuses, but BatchResult only allows constructors with either just data + errors or data + stasuses + errors. I added logic to add 200s for successful results to the stasuses map.

zhixuanjia (approved):

PR #235, opened 2023-01-31
feat: disable auto-apply for db evolution scripts

zhixuanjia:
This PR does two things:

1. Disable auto-apply for db evolution scripts
2. Fix the unit tests for DAOs. Previously, we have one db instance and one schema (gma_dev). The issue with this approach is that gradle will run tests in each class in parallel and those ebean-*.sql scripts are stepping on each other's toes. With new approach in this PR, we still have one db instance but multiple schemas (named by each class name such as EbeanLocalDaoTest). In this way, each script is executed in its own schemas without impacting others.

## Checklist


dao-impl/ebean-dao/build.gradle
```diff
@@ -30,15 +30,15 @@ dependencies {
...
-    exclude '**/EbeanLocalAccessTest.class'
-    exclude '**/EbeanLocalDAOTest.class'
-    exclude '**/EbeanLocalRelationshipWriterDAOTest.class'
-    exclude '**/EbeanLocalRelationshipQueryDAOTest.class'
-    exclude '**/FlywaySchemaEvolutionManagerTest.class'
-  }
-}
+//test {
+//  if (!project.hasProperty('test-ebean-dao')) {
+//    exclude '**/EbeanLocalAccessTest.class'
```
jsdonn, line 35: why not just remove these altogether instead of commenting?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -270,6 +270,13 @@ public SchemaConfig getSchemaConfig() {
     return _schemaConfig;
   }
 
+  /**
+   * Ensure table schemas is up-to-date with db evolution scripts.
+   */
+  public void ensureSchemaUpToDate() {
+    _localAccess.ensureSchemaUpToDate();
```
jsdonn, line 277: Do we need to add a check here to make sure that we are running in new schema/dual schema mode?

dao-impl/ebean-dao/src/test/resources/EbeanLocalDAOTest.conf
```diff
@@ -0,0 +1,3 @@
+version_table=my_version_table
```
jsdonn, line 1: So there will be 1 db server per test class instance? But one EmbeddedMariaInstance overall?

compared to 1 EmbeddedMariaInstance per test class instance before? Just want to make sure I'm understanding the changes.

jsdonn (changes requested):

dao-impl/ebean-dao/build.gradle
zhixuanjia: agree, removed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
zhixuanjia: makes sense, added a check.

dao-impl/ebean-dao/src/test/resources/EbeanLocalDAOTest.conf
zhixuanjia: So previously, we have one db instance and one schema (gma_dev). The issue with this approach is that gradle will run tests in each class in parallel and those ebean-*.sql scripts are stepping on each other's toes. 

With new approach in this PR, we still have one db instance but multiple schemas (named by each class name such as EbeanLocalDaoTest). In this way, each all those scripts are executed in its own schemas without impacting others.

Since we inferring the schema name through conf file name, we have to name each conf file also by its corresponding test classname.

jsdonn: Gotcha, thanks for clarifying. Can we also add that to the description of the PR to make it easier to understand the changes? thanks

jsdonn (approved):

PR #236, opened 2023-02-01
fix(browse): fix browse query to include must clause for removed field

rajatxgupta:
## Context
Currently, ESBrowseDAO uses a `must_not` clause in its query to filter active datasets by the `removed` _(boolean)_ field. 
This is the clause:
```
"must_not": { "term": { "removed": "true" } }
```

This causes an issue wherein those entities for which the **removed** field itself is missing from their corresponding search documents would also get surfaced in the results. 
This is also in contrast to the **search** query results, wherein the queries primarily have a '**must**' clause for '**removed = false**'.
Changes in this PR are to fix this issue in the browse behaviour. This change will make sure that only entities successfully marked _(indexed)_ as `removed = false` would surface in the results.

### List of changes:
- Update the 'buildQueryString' method in ESBrowseDAO
- Added corresponding unit test

## Checklist


zhixuanjia (approved):

yangyangv2 (commented):
LGTM

PR #237, opened 2023-02-01
feat: Soft-delete local relationships

zhixuanjia:
Context: To support multi-master setup, we need to remove usage of hard deletes.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,10 @@ jobs:
         with:
           # Needed to get all tags
           fetch-depth: 0
+
+      - name: Install lib
+        run: sudo apt-get install libncurses5
+
       - name: Release
         uses: ./.github/actions/release
         env:
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -10,11 +10,9 @@ import com.linkedin.metadata.dao.utils.RecordUtils;
 import com.linkedin.metadata.dao.utils.SQLSchemaUtils;
 import com.linkedin.metadata.dao.utils.SQLStatementUtils;
 import com.linkedin.metadata.query.Condition;
-import com.linkedin.metadata.query.Filter;
 import com.linkedin.metadata.query.LocalRelationshipCriterion;
 import com.linkedin.metadata.query.LocalRelationshipFilter;
 import com.linkedin.metadata.query.RelationshipDirection;
-import com.linkedin.metadata.query.RelationshipFilter;
 import io.ebean.EbeanServer;
 import io.ebean.SqlRow;
 import java.util.ArrayList;
@@ -27,7 +25,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
 import org.javatuples.Pair;
-import org.javatuples.Triplet;
 
 /**
  * An Ebean implementation of {@link BaseQueryDAO} backed by local relationship tables.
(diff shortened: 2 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -15,6 +15,10 @@ jobs:
         with:
           # Needed to get all tags
           fetch-depth: 0
+
+      - name: Install lib
```
jsdonn, line 19: Didn't you add this in your previous pr?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -108,14 +105,6 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
     return results;
   }
 
-  @Nonnull
-  public <SRC_ENTITY extends RecordTemplate, RELATIONSHIP extends RecordTemplate, INTER_ENTITY extends RecordTemplate> List<RecordTemplate> findEntities(
```
jsdonn, line 112: Why are we getting rid of this method?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -35,6 +35,7 @@ private static class CommonColumnName {
     private static final String METADATA = "metadata";
     private static final String LAST_MODIFIED_ON = "lastmodifiedon";
     private static final String LAST_MODIFIED_BY = "lastmodifiedby";
+    private static final String DELETED_TS = "deleted_ts";
```
jsdonn, line 38: Do we have to use their recommended version of soft deletion?

It might be a little confusing for users/future devs to have GMA version of soft deletion for entity tables but a different version for relationship tables.

Can we just do something similar with relationships where we set metadata = "{gma_deleted" = true}"?

.github/workflows/release.yml
zhixuanjia: That's for another workflow file. Last time it was for `build-and-test.yml`  and this is for `release.yml`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
zhixuanjia: This method was added when we previously extends the BaseQueryDao so we have to throw unimplemented exception. But now we're not extending that base class any more (change with local relationship filter), I'm thinking just remove it to clean up the code.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
zhixuanjia: first, using `deleted_ts` is a requirement for golden gate but even if it's not required, using `deleted_ts` is still a better solution because we can add index to `deleted_ts` column which speeds up the query a lot.

jsdonn: Yeah, that makes sense to me. However, my biggest concern is with mixing 2 different approaches of soft deletion. In the entity tables, we are setting each aspect column to something non-null when soft deleted, but also something that is obviously not that aspect's metadata. For relationships, if we go with this `ts_deleted` approach, we will not be touching the "metadata" column of the relationships. As Jyoti pointed out before in her soft-deletion doc:

`What does the metadata column in metadata_aspect table store if the aspect is soft-deleted i.e. isDeleted field is set? If it holds a non-null value, then this could lead to confusion as to why the metadata field/column is non-empty when the metadata has been deleted. If it holds a null value then the purpose of adding a new column is defeated (we later talk about storing a null value in the metadata column for soft-deleted metadata in the proposed solution).`

What are your thoughts on how to reconcile these 2 different soft deletion methods? I feel like that is an ideal scalable solution.

jsdonn: Maybe we can add changes to the metadata column of relationships similar to the entities? We don't have to use the metadata column for checking soft deletion status (we can use `ts_deleted` instead). But at least so when someone queries the table and sees that a relationship has been soft deleted, they won't be confused as to why the metadata column still has metadata information.

jsdonn: Offline discussion with Jesse -
1) we need to consolidate soft deletion methods eventually to resolve tech debt
2) current GMA soft deletion is not conceptually correct (we should treat soft-delete as if we hard-deleted a row, so if we want to "restore" the soft-deleted aspect, we need to create a new record)
3) for now, let's use `deleted_ts` on local relationship table, since it is a requirement for golden gate + there's not existing use cases for the local relationship tables yet anyways.

jsdonn (approved):

PR #240, opened 2023-02-24
always use current time when setting row-level lastmodifiedon

zhixuanjia:
We have two `lastmodifiedon` timestamps. One is in the aspect json column (a_aspect) and the other one is on row level. Current entity table backfill method will populate the row level `lastmodifiedon` using same timestamp as in  a_aspect. In this example below, during the backfill the row level `lastmodifiedon` could be set as 2021-01-01.
```
a_aspectfoo 2021-01-01
a_aspectbar 2023-01-01
```
The row level `lastmodifiedon` on entity table should always use current time whenever the row is modified.

## Checklist


jsdonn (approved):
Thanks for catching this bug and for the fix.

PR #245, opened 2023-03-31
add deleted_ts is null check

zhixuanjia:
## Summary

We're seeing over more then 2T binlog stacked in mysql db in one day which exhausted disk space. After examining the content of binlog, majority of the data is from soft deletion of local relationship table. Existing SQL will update all the rows (thousands) even it's already soft deleted. We should only update the ones that's NOT softed deleted yet.

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
@@ -63,7 +63,8 @@ public class SQLStatementUtils {
 
   private static final String DELETE_BY_DESTINATION = "UPDATE %s SET deleted_ts=NOW() WHERE destination = :destination AND deleted_ts IS NULL";
 
-  private static final String DELETE_BY_SOURCE_AND_DESTINATION = "UPDATE %s SET deleted_ts=NOW() WHERE destination = :destination AND source = :source AND deleted_ts IS NULL";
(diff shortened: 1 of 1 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -59,11 +59,11 @@ public class SQLStatementUtils {
       + "destination_type, lastmodifiedon, lastmodifiedby) VALUE (:metadata, :source, :destination, :source_type,"
       + " :destination_type, :lastmodifiedon, :lastmodifiedby)";
 
-  private static final String DELETE_BY_SOURCE = "UPDATE %s SET deleted_ts=NOW() WHERE source = :source";
```
jsdonn, line 62: Agreed we definitely need this change. Although Oleksii also mentioned we should add this to our read queries as well. Want to do that in the same PR?

zhixuanjia: I think we're doing this already for read queries, check MultiHopsTraversalSqlGenerator.java class. But in case you saw other places missing it, let's fix it separate PR.

jsdonn: Oh I see nvm. A little confused since I didn't see the SQL in SQLStatementUtils :p. No worries. Thanks.

jsdonn (approved):

PR #156, opened 2022-04-11
feat: Create base tracking manager and integrate states in ESSearchDAO.

RealChrisL:
- Define the generic base tracking manager and process states to be extended.
- Bump up minor version due to backward compatible DAO API changes.

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [x] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #251, opened 2023-05-08
feat: support of Tracking Context in MXE.

RealChrisL:
## Checklist
- This adds support for tracking context in MXE as the interface to propagate the tracking information for identifying the request lifecycle.
- Adding a new field is backward compatible if a default value is provided. The reader will substitute that default for missing data in records written with an older schema. At LinkedIn, we require that this new field is of a type that contains a union with null and that null be the default to allow readers to identify clearly records where the data is missing.
- Place the model in the dao-api module as it is event-specific.

### DIFF
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
@@ -3,9 +3,9 @@ namespace com.linkedin.metadata.events
 import com.linkedin.avro2pegasus.events.UUID
 
 /**
- * Descriptor for the tracking context. To represent the lifecycle of the trackable item.
+ *  Base model to capture tracking context and represent the lifecycle of the trackable item.
  */
-record TrackingContext {
+record BaseTrackingContext {
 
   /**
    * A UUID of a trackable item. This UUID should be propagated in all events referencing the trackable item.
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
@@ -1,0 +1,17 @@
+namespace com.linkedin.metadata.events
+
+/**
+ * Descriptor for the ingestion tracking context. To represent the lifecycle of the trackable ingestion item.
+ */
+record IngestionTrackingContext includes BaseTrackingContext{
+
+  /**
+   * The name of the service from which the ingestion event is being emitted.
+   */
+  emitter: optional string
+
+  /**
+   * The time at which the ingestion event was emitted into kafka.
+   */
+  emitTime: optional long
+}
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
@@ -3,7 +3,7 @@ namespace com.linkedin.metadata.events
 /**
  * Descriptor for the ingestion tracking context. To represent the lifecycle of the trackable ingestion item.
  */
-record IngestionTrackingContext includes BaseTrackingContext{
+record IngestionTrackingContext includes BaseTrackingContext {
 
   /**
    * The name of the service from which the ingestion event is being emitted.
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
@@ -1,0 +1,14 @@
+namespace com.linkedin.metadata.events
+
+import com.linkedin.avro2pegasus.events.UUID
+
+/**
+ * Descriptor for the tracking context. To represent the lifecycle of the trackable item.
+ */
+record TrackingContext {
+
+  /**
+   * A UUID of a trackable item. This UUID should be propagated in all events referencing the trackable item.
+   */
+  trackingId: optional UUID
+}
--- a/gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
+++ b/gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
@@ -5,7 +5,7 @@ namespace @(eventSpec.getNamespace())
 
 import com.linkedin.avro2pegasus.events.KafkaAuditHeader
 import com.linkedin.metadata.events.ChangeType
-import com.linkedin.metadata.events.TrackingContext
+import com.linkedin.metadata.events.IngestionTrackingContext
 import @eventSpec.getUrnType()
 import @eventSpec.getFullValueType()
 
@@ -41,7 +41,7 @@ record MetadataAuditEvent {
   changeType: optional union[null, ChangeType] = null
 
   /**
-   * Tracking context to identify the lifecycle of the trackable item.
+   * Tracking context to identify the lifecycle of the trackable ingestion item.
    */
-  trackingContext: optional union[null, TrackingContext] = null
+  ingestionTrackingContext: optional union[null, IngestionTrackingContext] = null
 }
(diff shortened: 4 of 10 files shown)


dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Descriptor for the tracking context. To represent the lifecycle of the trackable item.
+ */
+record TrackingContext {
+
+  /**
+   * A UUID of a trackable item. This UUID should be propagated in all events referencing the trackable item.
+   */
+  trackingId: optional UUID
```
yangyangv2, line 13: If we get the trackingID from Kafka audit header, would the type UUID still fit?

dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
```diff
@@ -0,0 +1,14 @@
+namespace com.linkedin.metadata.events
+
+import com.linkedin.avro2pegasus.events.UUID
+
+/**
+ * Descriptor for the tracking context. To represent the lifecycle of the trackable item.
+ */
+record TrackingContext {
```
yangyangv2, line 8: should this be IngestionTracingContext? I assume you will include emitter, ingestion time ..etc later.

and also, would it be beneficial to make it an abstract class? like TrackableEvent, TrackableIngestionEvent ..etc

dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
RealChrisL: Yes, the UUID fixed 16 types is how we defined the `messageID` in kafkaAuditHeader.

dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
yangyangv2: I'm wondering if we need to narrow down the scope of this context. The messageId is related to ingestion, it doesn't apply for general tracking such as queries.

And the for attributes like emitter, and ingestion time, they will be in kafka audit, but it's generated by kafka framework. We will still need a context object that contains the emitter to keep the context consistent through the entire ingestion cycle. 

I don't see a immediate need for "include" right now, but if you're going to place "messageId" into this context, then changing it to IngestionTrackingContext looks a better naming to me.

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
```diff
@@ -38,4 +39,9 @@ record MetadataAuditEvent {
    * Change type.
    */
   changeType: optional union[null, ChangeType] = null
+
```
jsdonn, line 42: Can you give me a quick summary for what this `.rythm` class is for?

jsdonn (commented):
Overall, LGTM. I will let you and Yang to resolve your conversation first before approving. I added 1 non-blocking comment.

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
RealChrisL: It is a template to be used for rendering the schema.

dao-api/src/main/pegasus/com/linkedin/metadata/events/TrackingContext.pdl
RealChrisL: Per discussion, create the `BaseTrackingContext` to capture the generic tracking properties and extend to `IngestionTrackingContext` to include ingestion pipeline-specific tracking properties.

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
```diff
@@ -0,0 +1,17 @@
+namespace com.linkedin.metadata.events
+
+/**
+ * Descriptor for the ingestion tracking context. To represent the lifecycle of the trackable ingestion item.
+ */
+record IngestionTrackingContext includes BaseTrackingContext{
```
yangyangv2, line 6: nit: space in "BaseTrackingContext {"

dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ *  Base model to capture tracking context and represent the lifecycle of the trackable item.
+ */
+record BaseTrackingContext {
+
+  /**
+   * A UUID of a trackable item. This UUID should be propagated in all events referencing the trackable item.
+   */
+  trackingId: optional UUID
```
yangyangv2, line 13: would this ID only make sense on Ingestion? Do you have use case for non-ingestion scenarios?

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
```diff
@@ -0,0 +1,17 @@
...
+
+  /**
+   * The name of the service from which the ingestion event is being emitted.
+   */
+  emitter: optional string
```
yangyangv2, line 11: Is it safe to say all the emitters will have an urn? or there're cases we only have the string format of emitter?

I'm good with string, which is safe to me, but want to push it a bit to see if we can make it a stronger type.

dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
RealChrisL: There may be a case the client wants to measure the query performance with network latency from the very beginning when they initiate the query. Hence they can assign the trackingId from their side and pass it through by the java template of `QureyTrackingContext` that includes the `BaseTrackingContext`.

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
RealChrisL: With strong urn type is definitely a better choice, yet, considering our upstream `kafkaAuditHeader` defines it as a string, to align the setting to propagate the information without adding any opinionated process.
Besides, there should be manageable emitters to serve the system, which is less likely to have confliction.

dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
yangyangv2: do you mean to have user to assign this trackingId explicitly? Can you elaborate the full use case cycle? e.g. if they assign UUID: 1234 in the tracking request, which returns in 200ms. And then who this UUID: 1234 in the TrackingContext helps the user?

RealChrisL: End to end will be:
1. Client initiates the query from their side and assigns tracking id.
2. Pass the tracking context to include the measurement of latency from the network.
3. GMA relays the trackingContext via DAO and storage.

Currently. we have implemented the step in the [searchDAO](https://github.com/linkedin/datahub-gma/pull/156/files) under the GMS.

With steps 1 & 2, we are able to extend the measurement to the client side, which requires passing tracking context.

jsdonn (approved):
Lgtm!

zhixuanjia (approved):
LGTM

dao-api/src/main/pegasus/com/linkedin/metadata/events/BaseTrackingContext.pdl
yangyangv2: sound good to me, thanks for explanation.

yangyangv2 (approved):

yangyangv2 (approved):

PR #252, opened 2023-05-09
Add toggle-able logic for job-gms record retrieval

jphui:
## Checklist

"""
`<feat>[getLatest()]: add support to use two alternative methods for checking record existence during record insertion`

Based off of logs introduced in a previous changeset, there is good reason to believe that `Ebean's EbeanServer.find(Class, Object)` method does not work accurately at scale, returning no matching existing records when it is clear that such a record exists. This appears to be the cause of mass duplicity introduced into the database. Luckily, Ebean has several ways to accomplish a task such as checking record existence, so this changeset attempts to leverage two of these in attempt to stop further duplicates from being introduced.

Might introduce breaking change in future if either (or both) of the alternative pathways is removed after adoption.
"""

- [x] Related issues linked
    - Added unit testing to cover all existing tests (which make use of EbeanLocalDAO's .find() ) but with each of retrieval mechanisms

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -402,8 +402,8 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   }
 
   /**
-   * Toggle using direct SQL query through Ebean when retrieving + checking existing records during insertion
-   * See GCN-38382
+   * Toggle using direct SQL query through Ebean when retrieving + checking existing records during insertion.
+   * See GCN-38382.
    */
   public void useDirectSqlQuery() {
     _directSqlRetrieval = true;
@@ -411,8 +411,8 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   }
 
   /**
-   * Toggle using Ebean's find() query builder when retrieving + checking existing records during insertion
-   * See GCN-38382
+   * Toggle using Ebean's find() query builder when retrieving + checking existing records during insertion.
+   * See GCN-38382.
    */
   public void useEbeanFindBuild() {
     _ebeanFindBuilder = true;
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -37,6 +37,8 @@ import io.ebean.EbeanServer;
 import io.ebean.ExpressionList;
 import io.ebean.PagedList;
 import io.ebean.Query;
+import io.ebean.SqlQuery;
+import io.ebean.SqlRow;
 import io.ebean.SqlUpdate;
 import io.ebean.Transaction;
 import io.ebean.config.ServerConfig;
@@ -85,6 +87,9 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   private IEbeanLocalAccess<URN> _localAccess;
   private UrnPathExtractor<URN> _urnPathExtractor;
   private SchemaConfig _schemaConfig = SchemaConfig.OLD_SCHEMA_ONLY;
+
+  // Flag for direct SQL execution for latest record retrieval
+  private boolean _directSqlRetrieval = false;
 
   public enum SchemaConfig {
     OLD_SCHEMA_ONLY, // Default: read from and write to the old schema table
@@ -155,6 +160,21 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   public EbeanLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer,
       @Nonnull ServerConfig serverConfig, @Nonnull Class<URN> urnClass, @Nonnull SchemaConfig schemaConfig) {
     this(aspectUnionClass, producer, createServer(serverConfig), serverConfig, urnClass, schemaConfig);
+  }
+
+  /**
+   * Constructor for EbeanLocalDAO with a parameter to pass in the GMS name, used temporarily to isolate job-gms-specific changes for duplicity.
+   *
+   * @param aspectUnionClass containing union of all supported aspects. Must be a valid aspect union defined in com.linkedin.metadata.aspect
+   * @param producer {@link BaseMetadataEventProducer} for the metadata event producer
+   * @param serverConfig {@link ServerConfig} that defines the configuration of EbeanServer instances
+   * @param urnClass Class of the entity URN
+   * @param gmsName Name of the gms in dash (-) delimiter format (ex. job-gms) -- not case-sensitive
+   */
+  public EbeanLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer,
+      @Nonnull ServerConfig serverConfig, @Nonnull Class<URN> urnClass, @Nonnull String gmsName) {
+    this(aspectUnionClass, producer, createServer(serverConfig), urnClass);
+    _directSqlRetrieval = gmsName.toLowerCase().equals("job-gms");
   }
 
   @VisibleForTesting
@@ -395,12 +415,46 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     }, 1);
   }
 
+  private  <ASPECT extends RecordTemplate> EbeanMetadataAspect queryLatest(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
+    final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
+
+    if (_directSqlRetrieval) {
+      return _server.find(EbeanMetadataAspect.class, key);
+    } else {
+      final String selectQuery = "SELECT * FROM metadata_aspect "
+          + "WHERE urn = :urn and aspect = :aspect and version = 0 "
+          + "ORDER BY createdOn DESC";
+
+      final SqlQuery query = _server.createSqlQuery(selectQuery);
+      query.setParameter("urn", urn.toString());
+      query.setParameter("aspect", aspectClass.getCanonicalName());
+
+      List<SqlRow> results = query.findList();
+
+      if (!results.isEmpty()) {
+        if (results.size() > 1) {
+          log.warn("Two version=0 records found for {}, {}", urn, aspectClass.getSimpleName());
+        }
+        SqlRow latestResult = results.get(0);
+
+        final EbeanMetadataAspect aspect = new EbeanMetadataAspect();
+        aspect.setKey(key);
+        aspect.setMetadata(latestResult.getString("metadata"));
+        aspect.setCreatedOn(latestResult.getTimestamp("createdOn"));
+        aspect.setCreatedBy(latestResult.getString("createdBy"));
+        aspect.setCreatedFor(latestResult.getString("createdFor"));
+
+        return aspect;
+      }
+      return null;
+    }
+  }
+
   @Override
   @Nonnull
   protected <ASPECT extends RecordTemplate> AspectEntry<ASPECT> getLatest(@Nonnull URN urn,
       @Nonnull Class<ASPECT> aspectClass) {
-    final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
-    EbeanMetadataAspect latest = _server.find(EbeanMetadataAspect.class, key);
+    EbeanMetadataAspect latest = queryLatest(urn, aspectClass);
     if (latest == null) {
       return new AspectEntry<>(null, null);
     }
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -402,7 +402,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   }
 
   /**
-   * Toggle using direct SQL query through Ebean when retrieving + checking existing records during insertion.
+   * Toggle using direct SQL query through Ebean's .findNative() when retrieving + checking existing records during insertion.
    * See GCN-38382.
    */
   public void useDirectSqlQuery() {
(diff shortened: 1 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +415,46 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+    if (!_directSqlRetrieval) {
+      return _server.find(EbeanMetadataAspect.class, key);
+    } else {
+      final String selectQuery = "SELECT * FROM metadata_aspect "
+          + "WHERE urn = :urn and aspect = :aspect and version = 0 "
+          + "ORDER BY createdOn DESC";
+
+      final SqlQuery query = _server.createSqlQuery(selectQuery);
+      query.setParameter("urn", urn.toString());
+      query.setParameter("aspect", aspectClass.getCanonicalName());
```
ibuenros, line 430: should you be using `ModelUtils.getAspectName` as above?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -157,6 +162,21 @@ public EbeanLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull Bas
...
+   * @param aspectUnionClass containing union of all supported aspects. Must be a valid aspect union defined in com.linkedin.metadata.aspect
+   * @param producer {@link BaseMetadataEventProducer} for the metadata event producer
+   * @param serverConfig {@link ServerConfig} that defines the configuration of EbeanServer instances
+   * @param urnClass Class of the entity URN
+   * @param gmsName Name of the gms in dash (-) delimiter format (ex. job-gms) -- not case-sensitive
+   */
+  public EbeanLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer,
+      @Nonnull ServerConfig serverConfig, @Nonnull Class<URN> urnClass, @Nonnull String gmsName) {
+    this(aspectUnionClass, producer, createServer(serverConfig), urnClass);
+    _directSqlRetrieval = gmsName.toLowerCase().equals("job-gms");
```
ibuenros, line 177: This seems hacky. Can't we just configure on a per-job basis? Or Lix it (understand this requires more code, so not necessary)?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: It does the same thing, but yes I agree for consistency it's worth making the change:
```
  public static <T extends DataTemplate> String getAspectName(@Nonnull Class<T> aspectClass) {
    return aspectClass.getCanonicalName();
  }
```

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: > configure on a per-job basis
Right, so the idea is that `job-gms` will [call this constructor](https://jarvis.corp.linkedin.com/codesearch/result/?name=BaseLocalDaoFactory.java&path=job-gms%2Fjob-factories%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fdatajob%2Ffactories%2Fdao&reponame=multiproduct%2Fjob-gms#126) and activate the flag.

Is there a better way to recognize "who" is the caller... just rattling some alternate ideas maybe adding a public method to EbeanLocalDAO that can "trigger" the flag manually? Might be a safer and less hacky way :)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +401,65 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+    if (_ebeanFindBuilder) {
+      results = _server.find(EbeanMetadataAspect.class)
+        .where()
+        .eq(URN_COLUMN, urn.toString())
+        .eq(ASPECT_COLUMN, ModelUtils.getAspectName(aspectClass))
+        .eq(VERSION_COLUMN, 0L)
+        .orderBy()
+        .desc(CREATED_ON_COLUMN)
+        .findList();
+    }
```
zhixuanjia, line 443: Under what condition, we choose this instead of direct sql retrieval?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +401,65 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
     }, 1);
   }
 
+  public void useDirectSqlQuery() {
```
zhixuanjia, line 404: I would suggest we add some documentation why have feature, maybe mention GCN ticket, google doc. This confuse new comers (future team members) why we have this method.

jphui: Ok sounds good! I'll add descriptions here as well!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +401,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+          + "WHERE urn = :urn and aspect = :aspect and version = 0 "
+          + "ORDER BY createdOn DESC";
+
+      results = _server.findNative(EbeanMetadataAspect.class, selectQuery)
+        .setParameter("urn", urn.toString())
+        .setParameter("aspect", ModelUtils.getAspectName(aspectClass))
+        .findList();
+    }
+    
+    if (_ebeanFindBuilder) {
```
jsdonn, line 442: What's the difference between this _server.find() method and the original one (line 426)?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -110,9 +117,20 @@ public class EbeanLocalDAOTest {
   private static final String GMA_CREATE_ALL_SQL = "gma-create-all.sql";
   private static final String GMA_DROP_ALL_SQL = "gma-drop-all.sql";
 
+  // run the tests 1 time for each FindMethology (3 total)
+  // See GCN-38382
+  private final FindMethodology _findMethodology;
+
+  private static enum FindMethodology {
+    ORIGINAL,
```
jsdonn, line 125: nit: original doesn't mean anything to people that don't have the previous context. can we rename this to something more meaningful?

jsdonn (changes requested):

jsdonn:
Can you also add some description to the PR? :) It will help with context for reviewers that aren't looped in on the original issue.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
ibuenros: Why do you need to determine who is the caller? I see there's a cfg as input to the constructor, can't the job just configure which DAO it wants to use?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
zhixuanjia: Can we at least add a TODO for the second find method to clean it up? Future code reader will be confused on why we implemented 3 find in three different way...

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: `ServerConfig` is part of the Ebean package and is meant to be used to configure the Server; it is not meant to be used for DAO configuration. See the updated way to set the flags; I got rid of this hacky way to do it and make it more explicit :)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +401,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+   */
+  public void useEbeanFindBuild() {
+    _ebeanFindBuilder = true;
+    _directSqlRetrieval = false;
+  }
+
+  private  <ASPECT extends RecordTemplate> EbeanMetadataAspect queryLatest(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
+    final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
+
+    if (!_directSqlRetrieval && !_ebeanFindBuilder) {
```
yangyangv2, line 425: Is this for experimental to have both? and you will only keep one SQL solution right?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -395,12 +401,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+   * Toggle using Ebean's find() query builder when retrieving + checking existing records during insertion.
+   * See GCN-38382.
+   */
+  public void useEbeanFindBuild() {
+    _ebeanFindBuilder = true;
+    _directSqlRetrieval = false;
+  }
+
+  private  <ASPECT extends RecordTemplate> EbeanMetadataAspect queryLatest(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
+    final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
```
yangyangv2, line 423: Is "key" used outside line 425? if not, please move it into the if-block.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -168,6 +192,19 @@ private <URN extends Urn> EbeanLocalDAO<EntityAspectUnion, URN> createDao(@Nonnu
     if (urnClass == BarUrn.class) {
       dao.setUrnPathExtractor((UrnPathExtractor<URN>) new BarUrnPathExtractor());
     }
+
+    // Toggle find logic based on flags set
+    // See GCN-38382
+    switch (_findMethodology) {
+      case DIRECT_SQL:
+        dao.useDirectSqlQuery();
```
yangyangv2, line 200: can we set FindMethodology directly to dao instead of operating on boolean flags?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: The idea is that we keep whichever one works with a priority on the direct SQL execution in order to wean off of Ebean's hidden implementation details.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jphui: > directly to dao

What does this mean? Is there a better way to accomplish this config?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -415,39 +401,58 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
     }, 1);
   }
 
+  public void useDirectSqlQuery() {
+    _directSqlRetrieval = true;
+    _ebeanFindBuilder = false;
+  }
+
+  public void useEbeanFindBuild() {
```
ibuenros, line 417: Is there a default? What happens if the user calls neither useDirectSqlQuery nor useEvenaFindBuild?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: Moved in most recent commit, thanks for the suggestion

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: Both start as `false` and the first line in `queryLatest()` is the default: `if (!_directSqlRetrieval && !_ebeanFindBuilder)` --> use original retrieval mechanism...

Thoughts on maybe a better design to have the default at the "bottom" rather than having an if-clause at the beginning employ it?

ibuenros: Oh, so in total there are three options:legacy, "ebeanFindBuilder", and "directSqlRetrieval".  Just add clear documentation about this behavior. Each public method should be documented, and you can mention default behavior there.

jphui: > documented

Is something brief like

```
Raise flag to use direct SQL execution through Ebean's findNative() method during record insertion
By default, Ebean's find(Class, Object) is used, see GCN-38382
```

sufficient?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jsdonn: I think Yang means similar to what we did for the new schema flags. Create an enum in EbeanLocalDAO class and operate on that instead of boolean flags. Although I feel that this is temporary work for now right? We will experiment with which method will resolve the issue first and then clean up the code after. So it shouldn't matter too much? Correct me if I'm wrong @jphui .

jphui: Ah I see... Based off of @ibuenros 's feedback, is it more prudent to:

1. add a new constructor that sets this Enum
2. have a public method serve as a toggle (as it is currently implemented)

jphui: Did a quick look at the implementation for dual schema and am uploading a new PR with that approach.

jphui:
(Had to force push because was running into a `[remote rejected] ... (failure)` issue (no error message)

ibuenros (approved):

jsdonn (approved):

PR #260, opened 2023-05-27
Refactor logging, introduce JDBC

jphui:
## Checklist

`<refactor / feat> :` refactor logging to be more concise and consistent, introduce JDBC query for cross checking with Ebean results

Suggestions from @dnbaker surrounding logging conventions in https://github.com/linkedin/datahub-gma/pull/255 have been incorporated. In addition, a JDBC query has been introduced to be used in cross-checking against Ebean's results to determine if the duplicity issue arises from the Ebean layer.

NOTE that JDBC is _not_ added as a 4th Find Methodology; it's intentionally made to be run in parallel in order to allow cross-checking.

#### The Big Idea

The end result of this PR is that _**preceding the existing logging**_ we will be able to see the results that JDBC "would have gotten" ... because if BOTH JDBC and Ebean return `null` then we can be assured that Ebean is not the cuprit!

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -452,6 +452,31 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
 
     // TODO(@jphui) added for job-gms duplicity debug, throwaway afterwards
 
+
+    // JDBC sanity check: should MATCH Ebean's results
+    if (log.isDebugEnabled()) { 
+      final String sqlQuery = "SELECT * FROM metadata_aspect "
+          + "WHERE urn = ? and aspect = ? and version = 0";
+        
+      try (Transaction transaction = _server.beginTransaction()) {
+      
+        // use PreparedStatement 
+        try (PreparedStatement stmt = transaction.getConnection().prepareStatement(sqlQuery)) {
+          stmt.setString(1, urn.toString());
+          stmt.setString(2, aspectName);
+          
+          try (ResultSet rset = stmt.executeQuery()) {
+            rset.last();  // go to the last returned record
+            log.debug("JDBC found {} existing records", rset.getRow());
+          }
+        }
+      
+        transaction.commit();
+      } catch (SQLException e) {
+        log.debug("JDBC ran into a SQLException: {}", e.getMessage());
+      }
+    }
+
     List<EbeanMetadataAspect> results = Collections.emptyList();
     Query<EbeanMetadataAspect> query = Ebean.find(EbeanMetadataAspect.class); // non-null placeholder to be overridden
 
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -453,7 +453,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     // TODO(@jphui) added for job-gms duplicity debug, throwaway afterwards
 
     // JDBC sanity check: should MATCH Ebean's results
-    if (log.isDebugEnabled() && "com.linkedin.dataJob.azkaban.AzkabanFlowInfo".equals(aspectName)) { 
+    if (log.isDebugEnabled() && "AzkabanFlowInfo".equals(aspectClass.getSimpleName())) { 
       final String sqlQuery = "SELECT * FROM metadata_aspect "
           + "WHERE urn = ? and aspect = ? and version = 0";
         
@@ -502,7 +502,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     results = query.findList();
 
     // Encouraged to run AFTER query execution based on getGeneratedSql() documentation
-    if (log.isDebugEnabled() && "com.linkedin.dataJob.azkaban.AzkabanFlowInfo".equals(aspectName)) {
+    if (log.isDebugEnabled() && "AzkabanFlowInfo".equals(aspectClass.getSimpleName())) {
       log.debug("Using {} retrieval; " + "Generated SQL: {}; urn: {}, aspect: {}, version: {}",
           _findMethodology.toString(),
           query.getGeneratedSql(),
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -595,13 +595,6 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     update.setParameter("createdBy", aspect.getCreatedBy());
     update.setParameter("oldTimestamp", oldTimestamp);
 
-    // TODO(yanyang) added for job-gms duplicity debug, throwaway afterwards
-    if (log.isDebugEnabled()) {
-      if ("AzkabanFlowInfo".equals(aspectClass.getSimpleName())) {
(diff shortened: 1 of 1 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -439,54 +443,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
     if (_findMethodology == FindMethodology.UNIQUE_ID) {
-      final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
+      final PrimaryKey key = new PrimaryKey(urn.toString(), aspectName, 0L);
       return _server.find(EbeanMetadataAspect.class, key);
     }
 
+    // TODO(@jphui) added for job-gms duplicity debug, throwaway afterwards
+
+    // JDBC sanity check: should MATCH Ebean's results
+    if (log.isDebugEnabled() && "com.linkedin.dataJob.azkaban.AzkabanFlowInfo".equals(aspectName)) {
```
yangyangv2, line 456: I added some comment in the other PR: https://github.com/linkedin/datahub-gma/pull/265. What is the difference between them?

jphui: I made this change in response to @dnbaker 's suggested changes to use a single SOT for the Aspect name.

The difference:
- `ModelUtils.getAspectName()` = `.getCanonicalName()` = "get the fully qualified class name"
- `.getSimpleName()` = "get only the suffix" like 'AzkabanFlowInfo'

yangyangv2: sorry what I mean is the what are the difference between these two PRs https://github.com/linkedin/datahub-gma/pull/265 and this one? should I move my comments in #265 to here?

jphui: I can move them here to expedite just saw this sorry

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -439,54 +443,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
     }
 
+    // TODO(@jphui) added for job-gms duplicity debug, throwaway afterwards
+
+    // JDBC sanity check: should MATCH Ebean's results
+    if (log.isDebugEnabled() && "com.linkedin.dataJob.azkaban.AzkabanFlowInfo".equals(aspectName)) { 
+      final String sqlQuery = "SELECT * FROM metadata_aspect "
+          + "WHERE urn = ? and aspect = ? and version = 0";
+        
+      try (Transaction transaction = _server.beginTransaction()) {
```
jphui, line 460: I am just copying the flow of logic from: https://ebean.io/docs/intro/queries/jdbc-query , does it hurt to include?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -439,54 +443,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+
     if (_findMethodology == FindMethodology.UNIQUE_ID) {
-      final PrimaryKey key = new PrimaryKey(urn.toString(), ModelUtils.getAspectName(aspectClass), 0L);
+      final PrimaryKey key = new PrimaryKey(urn.toString(), aspectName, 0L);
       return _server.find(EbeanMetadataAspect.class, key);
     }
 
+    // TODO(@jphui) added for job-gms duplicity debug, throwaway afterwards
+
+    // JDBC sanity check: should MATCH Ebean's results
```
jphui, line 455: how about wrapping this into a method and put it inside the FindMethodology block. Then this can be enabled from the conflg file

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -439,54 +443,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
+        
+      try (Transaction transaction = _server.beginTransaction()) {
+      
+        // use PreparedStatement 
+        try (PreparedStatement stmt = transaction.getConnection().prepareStatement(sqlQuery)) {
+          stmt.setString(1, urn.toString());
+          stmt.setString(2, aspectName);
+          
+          try (ResultSet rset = stmt.executeQuery()) {
+            rset.last();  // go to the last returned record
```
jphui, line 468: why not finishing this by reading from the resultSet and set values to EbeanMetadataAspect?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -439,54 +443,73 @@ public <ASPECT extends RecordTemplate> void backfillLocalRelationshipsFromEntity
...
-      // TODO(yanyang) added for job-gms duplicity debug, throwaway afterwards
-      if (log.isDebugEnabled()) {
-        if ("AzkabanFlowInfo".equals(aspectClass.getSimpleName())) {
-          log.debug("Using QUERY_BUILDER retrieval.");
-          log.debug("queryLatest SQL: " + query.getGeneratedSql());
-          log.debug("queryLatest Result: " + results);
-        }
-      }
+    // Encouraged to run AFTER query execution based on getGeneratedSql() documentation
+    if (log.isDebugEnabled() && "com.linkedin.dataJob.azkaban.AzkabanFlowInfo".equals(aspectName)) {
```
jphui, line 505: I would suggest using aspectClass.getSimpleName() to compare with "AzkabanFlowInfo" for performance reason.

how about moving this log into the the method block, so we can control from the config file?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: @yangyangv2 The JDBC functionality is intentionally run in **parallel** alongside the Ebean find() methods. As it is right now, we want them to run simultaneously to cross check results.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: @yangyangv2 The JDBC functionality is intentionally run in parallel alongside the Ebean find() methods. As it is right now, we want them to run simultaneously to cross check results.

I can add JDBC as a 4th `FIND_METHODOLOGY` if that approach makes more sense?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: I moved this out of the block because both `DIRECT_SQL` and `QUERY_BUILDER` will run the same logging lines.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
yangyangv2: Got it. Let's leave the JDBC with the current then.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
yangyangv2: I see, then how about moving this logic to line:506, so you don't have to run the same if-check again?

yangyangv2 (approved):
LGTM

PR #268, opened 2023-05-31
Ensure most recent metadata is served despite duplicate version 0's

jphui:
## Checklist


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -3094,6 +3094,29 @@ public void testNewSchemaExactMatchEmptyArray() {
...
+      AspectFoo v1 = new AspectFoo().setValue("foo");
+      AspectFoo v0 = new AspectFoo().setValue("bar");
+
+      // introduce two v0's
+      addMetadataWithAuditStamp(urn, aspectName, 0, v1, 1234L, "fakeCreator", "fakeImpersonator");
+      addMetadataWithAuditStamp(urn, aspectName, 0, v0, 0L, "fakeCreator", "fakeImpersonator");
+
+      // when retrieving a v0-duplicate record, the returned value should be the LATEST (timestamp=1234L)
+      AspectFoo actual = dao.get(AspectFoo.class, urn).get();
+      System.out.println(actual);
```
yangyangv2, line 3115: nit: remove this printout?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -892,6 +892,8 @@ private List<EbeanMetadataAspect> batchGetUnion(@Nonnull List<AspectKey<URN, ? e
       }
     }
 
+    sb.append(" ORDER BY createdOn DESC");
```
yangyangv2, line 895: I think this needs to be inside the For loop. Basically you want

    //   SELECT * FROM metadata_aspect WHERE urn = 'urn0' AND aspect = 'aspect0' AND version = 0 ORDER BY createdOn DESC limit 1
    //   UNION ALL
    //   SELECT * FROM metadata_aspect WHERE urn = 'urn0' AND aspect = 'aspect1' AND version = 0 ORDER BY createdOn DESC limit 1

does it make sense?

yangyangv2 (changes requested):
minor comment, overall looks good

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jphui: I thought through this as well and believed it should be outside because in the case where you're running multiple SELECT's, we want the _**final result**_ to be ordered, not just each subsection.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jphui: Good catch

PR #286, opened 2023-06-23
feat: Add isBackfill field in MetadataAuditEvent

JiaoMaWHU:
## Summary
- This pr proposes to add a `isBackfill` field into the mae pyaload
- `isBackfill = true` mae messages (can come from ^ API, the self service tool, and the backfill API) will be treated as a normal mae message even if the new value could be same as the old value. In this way, we will start to allow adhoc (batch in the future) backfill to secondary store without requiring config change and backup index.

## Checklist


dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionTrackingContext.pdl
```diff
@@ -14,4 +14,9 @@ record IngestionTrackingContext includes BaseTrackingContext {
    * The time at which the ingestion event was emitted into kafka.
    */
   emitTime: optional long
+
+  /**
+   * Indicates whether the ingestion event is a backfill event. empty or false means it is not a backfill event.
+   */
+  isBackfill: optional boolean
```
jsdonn, line 21: default false?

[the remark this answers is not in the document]
JiaoMaWHU: - if we don't add a new field and use emitTime instead, 
  - for mce, that means we will replace the current ingest API with the new idempotent ingest API, completely. Currently, the ingest API overwrites whenever there is an event received, meaning we take event arrival time as the ts. If we replace it with the new time-aware ingest API, we will change to use event send time as ts. I consider it as a good thing to do since we can provide Eventually Consistency based on this change. However, more effort and consideration will be required. Especially making sure every client creats the correct ts.
  - for mae, based on my knowledge it's impossible since we don't store last modified ts in es. Though it's possible to add a new field, more effort is needed.
 - Good question. I want tracking context to piggyback this field because we will probably expend this model to all ingestion events in the future so I don't want to duplicate the effort.

JiaoMaWHU: I'm referring the pdl schema [wiki](https://linkedin.github.io/rest.li/pdl_schema):
```
... Though the definition language allows default values for both required and optional fields, it is recommended to use default values only for required fields.
```
My understanding is you prefer to have a default value for required field so that you don't need to explicit set it in creation, but when it's optional you don't need to create it anyway, so no reason to have a default value. But not sure is this correct.

jsdonn: It's true, but I am thinking from conceptual standpoint, a boolean should always be set to true or false to indicate something. It's only optional to enforce backwards compatibility.

JiaoMaWHU: agreed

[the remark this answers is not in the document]
JiaoMaWHU: will add `isBackfill` field in mae payload in the next pr

gradle-plugins/metadata-annotations-test-models/src/main/pegasus/com/linkedin/testing/mxe/bar/annotatedAspectBar/MetadataAuditEvent.pdl
```diff
@@ -41,4 +41,17 @@ record MetadataAuditEvent {
    * Tracking context to identify the lifecycle of the trackable ingestion item.
    */
   ingestionTrackingContext: optional union[null, IngestionTrackingContext] = null
+
+  /**
+   * Flag to indicate if the event is a backfill request. True if the event is a backfill request. False and null mean it's not.
+   *
+   * Historically, the term "backfill" in gma may refer to
+   * 1. hitting the rest.li backfill endpoint of a resource
+   * 2. bootstrap a new aspect/entity in the graph store from scratch. Currently, if the oldValue and the newValue are the same,
```
jsdonn, line 50: nit: currently will not make sense once this feature is implemented completely :)

jsdonn (commented):
Code changes look fine but I am wondering, will there still be a use case of setting oldValue = newValue if we introduce this isBackfill flag?

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
```diff
@@ -44,4 +44,17 @@ record MetadataAuditEvent {
...
+   *
+   * Historically, the term "backfill" in gma may refer to
+   * 1. hitting the rest.li backfill endpoint of a resource
+   * 2. bootstrap a new aspect/entity in the graph store from scratch. Currently, if the oldValue and the newValue are the same,
+   * we treat it as a bootstrap mae and only consume it if a specific config is enabled.
+   *
+   * The field here indicates it's a backfill request resulted from 1, and should overwrite secondary store in live
+   * regardless of the config mentioned in 2.
+   */
+  isBackfill: optional boolean
```
keremsahin1, line 59: What benefit does this extra param provide say over you emit an MAE with oldValue = null and newValue non-null? If you emit such MAE, that will help you ingest into secondary stores?

JiaoMaWHU:
> Code changes look fine but I am wondering, will there still be a use case of setting oldValue = newValue if we introduce this isBackfill flag?

Good question. My plan is to not touch the existing backfill/bootstrap guide from you and Naga. I do agree they will cause less impact to prod index.

I will first use this field to allow oncall backfill search/graph events easily without switching index. Then will probably apply in the the regular backfill in platform consistency OKR, but both of them should only have trivial amount of documents per run, ideally less than 1000.

in the future, if we want to make the bootstrap more easier, we can switch it to include this tag.

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
JiaoMaWHU: Currently, we need to switch index in order to backfill the search/graph events even for just a few datasets to close an oncall ticket. I plan to create a new backfill endpoint with this tag included to reduce oncall ticket resolve time.

But I get your point, are you suggesting creating the new backfill endpoint by setting the oldValue as null? that's a good suggestion.

yangyangv2: In general, models are created for clarity, using old/new value combination to represent backfill dilute this purpose. The "old value" attribute has its purpose. Hacking this attribute for backfill breaks the integrity, which is an anti-pattern IMO.

yangyangv2 (commented):
LGTM overall unless there's strong objections/concerns about isBackfill

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
keremsahin1: @yangyangv2 Your comment is fair, I agree we shouldn't hack `oldValue` field to determine the behavior of MAE consumers. However, `isBackfill` sounds like yet another hack to me or I miss the point. I think whatever is contained in the MAE shouldn't dictate the behavior of processors.

From that perspective, we know that DAO wouldn't emit an MAE if there is no change on the aspect. And if there is an MAE emitted from the DAO where oldValue == newValue, this clearly shows that this is a backfill MAE. Then, why do we have to explicitly tag that this is a backfill MAE?

JiaoMaWHU: @keremsahin1 

It's because one condition can't help us do two things.

1. Currently, we use `oldValue == newValue` to determine it's a backfill and directly apply it on the backup index, which only makes sense for the bootstrap use case to avoid spike traffic to prod index.

2. However, we also want to backfill documents directly to prod index when it comes to oncall tickets or the regular sync in platform consistency.

So, either we move the use of `oldValue == newValue` from 1 to 2, or have something new to support 2. The former option will break the current bootstrap guide, and even if we fix it, every bootstrap will start to be applied on prod index, which to me, is dangerous considering our sync update nature in es graph job.

I'm open to any solution that can achieve both bootstrap and backfill use cases without hacking mae payload.

JiaoMaWHU: Please allow me to think this from the beginning. I assume the `oldValue` and `newValue` in the payload have two purposes:
1. serve the word `Audit` so that we know what's before the change and what's the data now.
2. dedup. If `oldValue` is the same as `newValue`, we should not process it since secondary store most likely has already updated this data, should skip it to reduce traffic.

The assumption that gms will not emit MAE with `oldValue` the same as `newValue` is a gms spec, which should not affect what we think on the secondary store side.

Then, with the fact that processing could fail, we want to add something on top of the purpose 2 so that it can be processed. Currently, we're using configs and then apply it to backup index, which should be the same as having a new field in the payload to overwrite it into the live index.

To me, the only problem left, is to migrate the config backfill part completely to use `isBackfill` tag, which we can prioritize in the future.

keremsahin1: > It's because one condition can't help us do two things.

That's MAE processor's concern not the MAE message. I think MAE message shouldn't dictate how the processors process MAE message. `isBackfill` tag from MAE message perspective seems redundant to me because oldValue == newValue tells us that already, right? You shouldn't move a downstream concern to the upstream, that will create a tight coupling. What happens if you wanna backfill only search index but not graph index? But I think you want to trigger a global backfill event in this case and that's fine.

Imo ideally what you want to have is a dynamic configuration mechanism on the Samza job so that whenever you want to write backfill events into prod index, you can flip that flag. We don't have that support unfortunately.

In short: I think `isBackfill` sounds like a hack to achieve easier dynamic configuration of the MAE processor jobs :) If that's the case and if adding this config into MAE schema is only feasible option, we can at least be more clear on the intent and use a better naming on the flag - a short form of `process and ingest this MAE into the production index`.

wdyt?

JiaoMaWHU: @keremsahin1 @yangyangv2 I want to go with setting `oldValue` as null approach.

We all agree either 1) adding `isBackfill` or 2) setting `oldValue` as null is fair to be considered as hack solution.

The struggle we have is the poor performance of updating es index. We don't want to affect prod index so we decided to produce to backup index first.

Considering we're likely moving to HS, which supports KPJ backfill against prod index, I don't think we will have this struggle any more. Then, I believe it's safe to say 2) impacts us less as it doesn't change mae payload. By the time HS migration finishes, we can stop the hacky use of 1).

JiaoMaWHU:
Close the pr based on discussion in https://github.com/linkedin/datahub-gma/pull/286#discussion_r1277752227

PR #290, opened 2023-07-28
feat: Add a new backfill method that sets the old value as null in MAE

JiaoMaWHU:
## Summary
- We want to support low volume backfill against the live es index. Currently, such request needs manual operation to switch es indexes, which is dangerous and requires quite effort. In reality, we can't even close an oncall ticket.
- As the continuation in https://github.com/linkedin/datahub-gma/pull/286#discussion_r1277752227, we considered 1) adding tag in MAE payload; vs 2) setting old value as null to bypass consumer check. Eventually decided to go with 2) as it doesn't impact the MAE payload and can be easily deprecated as we migrate to other secondary store.
- Implement 2) in `BaseAspectRoutingResource` and `BaseEntityResource`.

## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/RestliConstants.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/RestliConstants.java
@@ -11,6 +11,7 @@ public final class RestliConstants {
   public static final String ACTION_BACKFILL = "backfill";
   public static final String ACTION_BACKFILL_ENTITY_TABLES = "backfillEntityTables";
   public static final String ACTION_BACKFILL_WITH_URNS = "backfillWithUrns";
+  public static final String ACTION_BACKFILL_ES_LIVE_INDEX = "backfillESLiveIndex";
   public static final String ACTION_BACKFILL_LEGACY = "backfillLegacy";
   public static final String ACTION_BROWSE = "browse";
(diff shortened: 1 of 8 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/RestliConstants.java
```diff
@@ -11,6 +11,7 @@ private RestliConstants() { }
   public static final String ACTION_BACKFILL = "backfill";
   public static final String ACTION_BACKFILL_ENTITY_TABLES = "backfillEntityTables";
   public static final String ACTION_BACKFILL_WITH_URNS = "backfillWithUrns";
+  public static final String ACTION_BACKFILL_ES_LIVE_INDEX = "backfillESLiveIndex";
```
zhixuanjia, line 14: I think backfill to live index is applicable to both HS and ES? Can we name the endpoint infra agnostic? like `backfillLiveSearchIndex`

JiaoMaWHU: discussed offline. Tho this ES is added for future deprecation purpose, it's better to avoid specific storage engine name in a generic library.

zhixuanjia (approved):
LGTM

PR #291, opened 2023-08-16
disable MAEv4 emission by default

zhixuanjia:
disable MAEv4 emission by default. Can be turned on case-by-case if necessary.


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -170,7 +170,7 @@ public static class AspectUpdateLambda<ASPECT extends RecordTemplate> {
   // Enable updating multiple aspects within a single transaction
   private boolean _enableAtomicMultipleUpdate = false;
 
-  private boolean _emitAuditEvent = true;
```
RealChrisL, line 173: thanks for setting the default value, let's also make a major version bump considering it will change the underlying behavior (OSS).

zhixuanjia: makes sense. Bumped the major version.

RealChrisL (approved):

PR #293, opened 2023-08-22
feat(auditing): add AuditStamp to MAE schema

jsdonn:
- add (already existing) AuditStamp to MAE schema so that metadata version history can be tracked via MAEs
- modify BaseMetadataEventProducer/BaseTrackingMetadataEventProducer to take in AuditStamp as a parameter
- add logic to call produceAspectSpecificMetadataAuditEvent with AuditStamp

Next steps:
- update KafkaMetadataEventProducer in metadata-models to implement new method with AuditStamp param
- test MAE emission and verify AuditStamp attached

For non-LI users:
- update your EventProducer (the one that extends BaseMetadataEventProducer) to use the AuditStamp param in its produce... methods.

## Checklist


gradle-plugins/metadata-annotations-test-models/src/main/pegasus/com/linkedin/testing/mxe/bar/annotatedAspectBar/MetadataAuditEvent.pdl
```diff
@@ -41,4 +42,9 @@ record MetadataAuditEvent {
    * Tracking context to identify the lifecycle of the trackable ingestion item.
    */
   ingestionTrackingContext: optional union[null, IngestionTrackingContext] = null
+
+  /**
+   * Audit info (i.e. createdon, createdby, createdfor) to track the version history of metadata changes.
+   */
+  auditStamp: optional AuditStamp
```
yangyangv2, line 49: should this be mandatory?

gradle-plugins/metadata-events-generator-lib/src/main/resources/MetadataAuditEvent.rythm
```diff
@@ -44,4 +45,9 @@ record MetadataAuditEvent {
    * Tracking context to identify the lifecycle of the trackable ingestion item.
    */
   ingestionTrackingContext: optional union[null, IngestionTrackingContext] = null
+
+  /**
+   * Audit info (i.e. createdon, createdby, createdfor) to track the version history of metadata changes.
+   */
+  auditStamp: optional AuditStamp
```
yangyangv2, line 52: should this be required?

gradle-plugins/metadata-annotations-test-models/src/main/pegasus/com/linkedin/testing/mxe/bar/annotatedAspectBar/MetadataAuditEvent.pdl
jsdonn: Let's discuss tmrw during sync up. I am thinking about if we can track only certain aspects' version history. Like compliance or ownership information. But if we can ignore things like ML entities/aspects for example.

yangyangv2: it is better to make this mandatory to keep the consistency and simplicity. Otherwise every MCE you will need to consider whether to include the auditstamp for both consumers and producers. Let's sync todya on this.

jsdonn: Discussed offline, let's keep it as mandatory.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -548,12 +548,12 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> aspectUpdateHelper(URN
   }
 
   private <ASPECT extends RecordTemplate> ASPECT_UNION unwrapAddResultToUnion(URN urn, AddResult<ASPECT> result,
-      @Nullable IngestionTrackingContext trackingContext) {
-    ASPECT rawResult = unwrapAddResult(urn, result, trackingContext);
+      AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext) {
```
yangyangv2, line 551: add @Nullable to make it clear?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -548,12 +548,12 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> aspectUpdateHelper(URN
...
+    ASPECT rawResult = unwrapAddResult(urn, result, auditStamp, trackingContext);
     return ModelUtils.newEntityUnion(_aspectUnionClass, rawResult);
   }
 
-  private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, AddResult<ASPECT> result,
+  private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, AddResult<ASPECT> result, AuditStamp auditStamp,
```
yangyangv2, line 556: @Nullable

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -576,9 +576,9 @@ private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, AddResul
     if (_emitAspectSpecificAuditEvent) {
       if (_alwaysEmitAspectSpecificAuditEvent || oldValue != newValue) {
         if (_trackingProducer != null) {
-          _trackingProducer.produceAspectSpecificMetadataAuditEvent(urn, oldValue, newValue, trackingContext);
+          _trackingProducer.produceAspectSpecificMetadataAuditEvent(urn, oldValue, newValue, auditStamp, trackingContext);
```
yangyangv2, line 579: If we pass a null auditStamp to the _trackingProducer, will it be created before sending MAE?

jsdonn: I made the field on the MAE a `union[null, AuditStamp]`. I am thinking the audit stamp can be null when we emit backfill MAEs.

jsdonn: So if a null AuditStamp if passed to the producer, it will just set it to null and it won't populate it.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Actually, I think it should be Nonnull. I added annotations. Only time it should be null is when we got through the backfill() API. Otherwise if dao.add is called by the GMS, it should always pass in an audit stamp.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
yangyangv2: Can we make MAE always has the AuditStamp? You may have createFor to be optional, but it should have createdOn and createdBy, which is the most important information in auditing.

jsdonn: What do we do for backfill MAEs then? Create dummy `createdBy` value?

If we do need to do auditing, we also have to do the additional filter on backfill MAEs. I think it is simpler to just set backfill MAE auditStamp to null for this purpose.

jsdonn: Other option is I can add a validation step before producing the MAE to check `(auditStamp == null && oldValue == newValue) || auditStamp != null`

yangyangv2: the question is to whether audit the backfill. backfill is initiated by someone or service so should have created by. But it doesn't result in change history so I'm on the fence on whether to keep the MAE for pure change history auditing or also taking backfill (which will result in the secondary schema update) into account.

yangyangv2: But since we don't audit the secondary storage update today, let's just take the simple step using auditStamp == null for now. If later we want to audit the secondary storage update, we can include the auditStamp and another attribute to indicate this is a backfill. (or using oldValue == newValue, but not efficient in query.)

yangyangv2 (approved):
LGTM

PR #308, opened 2023-10-27
[Part 1][Persist message id] Adjust interface to persist message id

zhixuanjia:
IngestionTrackingContext contains the message id associated with input MCE. This PR adjust the method interface to persist this message id into entity table. There will be a follow-up PR after this one to actually persist the message id.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -1,5 +1,6 @@
 package com.linkedin.metadata.dao;
 
+import com.linkedin.avro2pegasus.events.UUID;
 import com.linkedin.common.AuditStamp;
 import com.linkedin.common.urn.Urn;
 import com.linkedin.data.template.RecordTemplate;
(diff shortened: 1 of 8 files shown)


JiaoMaWHU:
we will use messageId to do consistency match?

zhixuanjia:
> we will use messageId to do consistency match?

That's correct

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -91,7 +92,7 @@ public void ensureSchemaUpToDate() {
   @Override
   @Transactional
   public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPECT newValue, @Nonnull Class<ASPECT> aspectClass,
-      @Nonnull AuditStamp auditStamp) {
+      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext) {
```
JiaoMaWHU, line 95: Using messageid to do consistency match seems a great idea to me : )

I have one comment, why not just pass in the messageId as the parameter? ingestionTrackingContext is a variable tied to ingest, which is at rest.li API level. Here we're passing it down to the dao level, I think dao don't need to understand what's a ingestionTrackingContext.

zhixuanjia: hmm, good point, let me update.

yangyangv2 (approved):
LGTM, please link the design doc to the PR and clarify the purpose. thanks!

PR #310, opened 2023-11-01
fix(new schema): read metadata_aspect during getLatest regardless of schema mode

jsdonn:
There is a bug with a recent PR #298 which is causing OptimisticLockException to be thrown. Essentially, getLatest is reading from the new schema (in new schema only mode) but updateWithOptimisticLocking is trying to apply optimistic locking on the old schema table using the timestamp returned from the getLatest call on the entity table. The timestamps in the old and new schema tables may be off by a few ms so the update will fail during timestamp comparation.

This PR aims to revert part of that PR by reading the old schema table during getLatest always UNLESS change log is disabled.
## Checklist


yangyangv2 (approved):
LGTM

PR #311, opened 2023-11-01
APA-93246 fix dual-write issue on update wiht optimitic locking

yangyangv2:
## Context

This change is to fix a bug in the optimistic locking update, whereas in NEW/DUAL Schema mode, we use last_modified time getting from getLatest() on entity table to run the UPDATE WHEN last_modified = :oldTimestamp on aspect table. Even in most cases the last modified time between entity/aspect table are the same. However, this is not guaranteed. The fix is to write a new optimistic locking logic for entity tables. The logic is written updateWithOptimisticLocking and  we expect this method to behave as below 

1.  In OLD_SCHEMA mode, 
    - getLatest runs over the aspect table
    - updateWithOptimisticLocking() with compare-and-set latest version on aspect table.
    - if updateWithOptimisticLocking fails, all the updates will be rollback

2. In NEW/DUAL schema model, when changeLog is enabled (default mode)
    - getLatest runs over entity tables
    - updateWithOptimisticLocking() with compare-and-set latest version on entity tables.
    - aspect table will be dual-written.
    - if updateWithOptimisticLocking fails, all the updates will be rollback
   
3. When changeLog is disabled (cold-archive mode, NEW schema only)
    - updateWithOptimisticLocking() shouldn't be called, an UnsupportedOperationExecption will be thrown.
    - in cold-archive mode, we don't need optimistic locking anymore since the updates on entity table is atomistic.

## Known Limitations
1. updateWithOptimisticLocking() is using JSON_EXTRACT to extract last modified time form AuditedAspect, which could have performance impact. But considering the primary key (urn) will restrict the search scope to be O(1), the impact should be trivial but worth of observation.
2. for entities which as been soft_deleted, there's no timestamp we can leverage from AuditedAspect for optimistic locking check. It is a feature gap which we should include in the future. The current solution is consider soft_deleted record will always satisfy optimistic locking. It may cause inconsistency between entity and aspect table. But considering this is an edge case, and the only consumer of aspect table is MAD which will migrate away after 11/10. We will consider this as an acceptable risk.  

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -736,10 +736,10 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   //      by disregarding any user change.
   // Ideally, another column for the sake of optimistic locking would be preferred but that means a change to
   // metadata_aspect schema and we don't take this route here to keep this change backward compatible.
-  static final String OPTIMISTIC_LOCKING_UPDATE_SQL = "UPDATE metadata_aspect "
+  private static final String OPTIMISTIC_LOCKING_UPDATE_SQL = "UPDATE metadata_aspect "
       + "SET urn = :urn, aspect = :aspect, version = :version, metadata = :metadata, createdOn = :createdOn, createdBy = :createdBy "
       + "WHERE urn = :urn and aspect = :aspect and version = :version";
-  
+
   /**
    * Assembly SQL UPDATE script for old Schema.
    * @param aspect {@link EbeanMetadataAspect}
@@ -747,7 +747,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
    *                     with oldTimestamp during the update.
    * @return {@link SqlUpdate} for SQL update execution
    */
-  private SqlUpdate assemblyOldSchemaSqlUpdate(@Nonnull EbeanMetadataAspect aspect, @Nullable Timestamp oldTimestamp) {
+  private SqlUpdate assembleOldSchemaSqlUpdate(@Nonnull EbeanMetadataAspect aspect, @Nullable Timestamp oldTimestamp) {
 
     final SqlUpdate oldSchemaSqlUpdate;
     if (oldTimestamp == null) {
@@ -786,7 +786,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
       // In NEW_SCHEMA or DUAL_SCHEMA, since entity table is the SOT and the getLatest (oldTimestamp) is from the entity
       // table, therefore, we will apply compare-and-set with oldTimestamp on entity table (addWithOptimisticLocking)
       // aspect table will apply regular update over (urn, aspect, version) primary key combination.
-      oldSchemaSqlUpdate = assemblyOldSchemaSqlUpdate(aspect, null);
+      oldSchemaSqlUpdate = assembleOldSchemaSqlUpdate(aspect, null);
       numOfUpdatedRows = runInTransactionWithRetry(() -> {
         UUID messageId = trackingContext != null ? trackingContext.getTrackingId() : null;
         // DUAL WRITE: 1) update aspect table, 2) update entity table.
@@ -797,7 +797,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     } else {
       // In OLD_SCHEMA mode since aspect table is the SOT and the getLatest (oldTimestamp) is from the aspect table
       // therefore, we will apply compare-and-set with oldTimestamp on aspect table (assemblyOldSchemaSqlUpdate)
-      oldSchemaSqlUpdate = assemblyOldSchemaSqlUpdate(aspect, oldTimestamp);
+      oldSchemaSqlUpdate = assembleOldSchemaSqlUpdate(aspect, oldTimestamp);
       numOfUpdatedRows = _server.execute(oldSchemaSqlUpdate);
     }
     // If there is no single updated row, emit OptimisticLockException
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -663,7 +663,7 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
       @Nonnull Class<ASPECT> aspectClass) {
 
     EbeanMetadataAspect result;
-    if (_changeLogEnabled) {
+    if (_schemaConfig == SchemaConfig.OLD_SCHEMA_ONLY) {
       final String aspectName = ModelUtils.getAspectName(aspectClass);
       final PrimaryKey key = new PrimaryKey(urn.toString(), aspectName, LATEST_VERSION);
       if (_findMethodology == FindMethodology.DIRECT_SQL) {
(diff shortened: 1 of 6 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -728,6 +728,43 @@ private <ASPECT extends RecordTemplate> EbeanMetadataAspect buildMetadataAspectB
...
+      + "WHERE urn = :urn and aspect = :aspect and version = :version";
+  
+  /**
+   * Assembly SQL UPDATE script for old Schema.
+   * @param aspect {@link EbeanMetadataAspect}
+   * @param oldTimestamp old timestamp. If provided, the generated SQL will use optimistic locking and do compare-and-set
+   *                     with oldTimestamp during the update.
+   * @return {@link SqlUpdate} for SQL update execution
+   */
+  private SqlUpdate assemblyOldSchemaSqlUpdate(@Nonnull EbeanMetadataAspect aspect, @Nullable Timestamp oldTimestamp) {
```
jsdonn, line 750: nit: assembleOldSchemaSqlUpdate

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -728,6 +728,43 @@ private <ASPECT extends RecordTemplate> EbeanMetadataAspect buildMetadataAspectB
...
 
+  // Build manual SQL update query to enable optimistic locking on a given column
+  // Optimistic locking is supported on ebean using @version, see https://ebean.io/docs/mapping/jpa/version
+  // But we can't use @version annotation for optimistic locking for two reasons:
+  //   1. That prevents flag guarding optimistic locking feature
+  //   2. When using @version annotation, Ebean starts to override all updates to that column
+  //      by disregarding any user change.
+  // Ideally, another column for the sake of optimistic locking would be preferred but that means a change to
+  // metadata_aspect schema and we don't take this route here to keep this change backward compatible.
+  static final String OPTIMISTIC_LOCKING_UPDATE_SQL = "UPDATE metadata_aspect "
```
jsdonn, line 739: nit: private?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -736,43 +773,37 @@ protected <ASPECT extends RecordTemplate> void updateWithOptimisticLocking(@Nonn
...
+      // aspect table will apply regular update over (urn, aspect, version) primary key combination.
+      oldSchemaSqlUpdate = assemblyOldSchemaSqlUpdate(aspect, null);
       numOfUpdatedRows = runInTransactionWithRetry(() -> {
         UUID messageId = trackingContext != null ? trackingContext.getTrackingId() : null;
-        _localAccess.add(urn, (ASPECT) value, aspectClass, newAuditStamp, messageId);
-        return _server.execute(update);
+        // DUAL WRITE: 1) update aspect table, 2) update entity table.
+        // Note: when cold-archive is enabled, this method: updateWithOptimisticLocking will not be called.
+        _server.execute(oldSchemaSqlUpdate);
+        return _localAccess.addWithOptimisticLocking(urn, (ASPECT) value, aspectClass, newAuditStamp, oldTimestamp, messageId);
```
jsdonn, line 795: don't we still want to throw an optimistic locking error here if one of these updates didn't complete?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -48,6 +48,14 @@ public class SQLStatementUtils {
       "INSERT INTO %s (urn, a_urn, %s, lastmodifiedon, lastmodifiedby) VALUE (:urn, :a_urn, :metadata, :lastmodifiedon, :lastmodifiedby) "
           + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon;";
 
+  private static final String SQL_UPDATE_ASPECT_TEMPLATE =
+      "UPDATE %s SET %s = :metadata, lastmodifiedon = :lastmodifiedon, lastmodifiedby = :lastmodifiedby "
+          + "WHERE urn = :urn and (JSON_EXTRACT(%s, '$.lastmodifiedon') = :oldTimestamp OR JSON_EXTRACT(%s, '$.gma_deleted') IS NOT NULL);";
```
jsdonn, line 53: nit - leave comment here explaining why we are checking soft deleted aspect + assumption that optimistic locking will hold for soft deleted aspects

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -3085,17 +3089,62 @@ public void testOptimisticLockException() {
...
+
+      dao.insert(fooUrn, fooAspect, AspectFoo.class, makeAuditStamp("fooActor", _now), 0, null);
+      // make inconsistent timestamp on aspect table
+      aspect.setCreatedOn(new Timestamp(_now + 1));
+      _server.update(aspect);
+
+      // When: update with old timestamp matches the lastmodified time in entity table
+      try {
+        fooAspect.setValue("bar");
+        dao.updateWithOptimisticLocking(fooUrn, fooAspect, AspectFoo.class, makeAuditStamp("fooActor", _now + 200), 0,
```
jsdonn, line 3122: I'm confused how does `_now+200` match `_now+1`?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
yangyangv2: it's thrown at line 804

jsdonn: My bad, misread code.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
yangyangv2: the now+1 mimics the scenario that aspect and entity table may have different creation time. It's not the now+200 we match here, but more a `_now` as oldTimestamp coming from entity table's getLatest, to compare with _now in stored in entity table when running the optimistic locking update.

now+200 is the timestamp of updating a new ingestion (newer timestamp)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
yangyangv2: good suggestion, added in the comments

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jsdonn: Understood, thanks.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -206,6 +217,22 @@ public static <ASPECT extends RecordTemplate> String createAspectUpsertSql(@Nonn
...
+   * @param <ASPECT> aspect type
+   * @param aspectClass aspect class
+   * @return aspect upsert sql
+   */
+  public static <ASPECT extends RecordTemplate> String createAspectUpdateWithOptimisticLockSql(@Nonnull Urn urn,
+      @Nonnull Class<ASPECT> aspectClass, boolean urnExtraction) {
+    final String tableName = getTableName(urn);
+    final String columnName = getAspectColumnName(aspectClass);
+    return String.format(urnExtraction ? SQL_UPDATE_ASPECT_WITH_URN_TEMPLATE : SQL_UPDATE_ASPECT_TEMPLATE, tableName,
+        columnName, columnName, columnName, columnName);
```
zhixuanjia, line 233: Why column name is repeated four times? Based on SQL template, seems just one of them is enough?

zhixuanjia: sorry, my bad, didn't see there're more %s in the template.

zhixuanjia (approved):

PR #312, opened 2023-11-02
Use emitTime in ingestionTrackingContext as last modified time

derekpham:
We might want to backfill by re-emitting MCE's that have failed to process. However, we also want to avoid overwriting new data with old data. To resolve that, we will use `emitTime` in `IngestionTrackingContext` to save as `lastModifiedTime` for the aspect in the database. There will be a follow up PR that will discard a change if `lastModifiedTime > emitTime`.

There are a lot of places where we extract `AuditStamp`'s time and save it to the database. I think overwriting the stamp's time with the `emitTime` should be the easiest way to save it as `lastModifiedTime`.
I've also checked that all of `emitTime` setters use clock current in milliseconds, so this change should not cause any issue.

## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
@@ -236,7 +236,6 @@ public abstract class BaseEntityResource<
   public Task<Void> ingest(@ActionParam(PARAM_SNAPSHOT) @Nonnull SNAPSHOT snapshot) {
     return ingestInternal(snapshot, Collections.emptySet(), null);
   }
-
   /**
    * Same as {@link #ingest(RecordTemplate)} but with tracking context attached.
    * @param snapshot Snapshot of the metadata change to be ingested
@@ -257,6 +256,9 @@ public abstract class BaseEntityResource<
     return RestliUtils.toTask(() -> {
       final URN urn = (URN) ModelUtils.getUrnFromSnapshot(snapshot);
       final AuditStamp auditStamp = getAuditor().requestAuditStamp(getContext().getRawRequestContext());
+
+      overwriteAuditStampTimeWithEmitTime(auditStamp, trackingContext);
+
       ModelUtils.getAspectsFromSnapshot(snapshot).stream().forEach(aspect -> {
         if (!aspectsToIgnore.contains(aspect.getClass())) {
           getLocalDAO().add(urn, aspect, auditStamp, trackingContext);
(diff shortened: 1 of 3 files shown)


jsdonn:
Doesn't this mean that the lastmodifiedon value in the db no longer represents when that row/aspect was truly last modified? Instead the value will be when the change event for that aspect was emitted. Don't think this is too big of a deal but there may be some risks with this that we haven't identified yet with regards to other DAO logic. I prob need to think harder about this. Wondering if you can think of any possible issues as well.

Also what happens if tracking context or emit time is null? Then the DAO will continue with its existing logic and insert with timestamp `System.currentTimeMillis`. Will that cause any problems when using the backfill tool?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -266,6 +268,23 @@ protected Task<Void> ingestInternal(@Nonnull SNAPSHOT snapshot,
...
+   * more complicated logic down the call chain.
+   * The intention is to use the emitter's emit time as last modified time. This is helpful for backfilling purposes
+   * where events from the past are emitted. If the emit time is less than the last modified time, the ingest
+   * event will be a no-op.
+   */
+  private void overwriteAuditStampTimeWithEmitTime(
+      @Nonnull AuditStamp auditStamp,
+      @Nullable IngestionTrackingContext ingestionTrackingContext) {
+    if (ingestionTrackingContext != null && ingestionTrackingContext.hasEmitTime()) {
+      auditStamp.setTime(ingestionTrackingContext.getEmitTime());
```
JiaoMaWHU, line 284: What is auditstamp used in metadata_aspect vs entity table? is auditstamp time the same as db modified time in entity table? if so, I agree with Justin. Do we need to overwrite the time anyway? can't just use `lastmodifiedon` to compare with `emitTime`?

derekpham:
From my understanding, here's how `auditStamp` is used:
- Set as `createdOn` in the old schema. See [here](https://github.com/linkedin/datahub-gma/blob/e4d9e92d370047aa1da14492c04207040195e7fd/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java#L719) and [here](https://github.com/linkedin/datahub-gma/blob/e4d9e92d370047aa1da14492c04207040195e7fd/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java#L736)
- Set as `lastmodifiedon` in the new schema, which I believe is equivalent to `createdOn` in the old schema. See [here](https://github.com/linkedin/datahub-gma/blob/b38cf9db8f9b5716382bc4f824c49e0d34018dd4/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java#L103)

The advantage of using `emitTime` as `lastmodifiedon` is that we don't need to make any model changes and worry about things like updates not happening because of a small window between `emitTime` and current timestamp when GMS receives the message.
I would be happy to discuss this further because I'm definitely not as familiar with the system as you guys are. This option was originally suggested by Edwina and is the recommended option by me in the MCEv5 backfill design doc.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
derekpham: we're not doing the logic to discard a write if `lastmodifiedon > emitTime` in this PR. That will follow in the upcoming PR. If we compare `lastmodifiedon` to `emitTime`, there will be a scenario where `emitTime < lastmodifiedon` but it should still be processed. For example, there are 2 events with `emitTime` at `t1` and `t2`. Event 1 gets processed at time `t3` `> t1, t2`. In this case, we will discard the second event when we should not.

yangyangv2: lastmodifiedon is the GMS receive time, which is different from emit time which is supposed to be captured at the producer. Let's not repurpose this field.

derekpham:
Closing this and going with a different approach

PR #314, opened 2023-11-08
Add naive approach to backfilling

derekpham:
Based on the last discussion, we'll be first implementing a naive approach where we compare `emitTime` against `lastmodifiedtime`. If `emitTime < lastmodifiedtime`, we'll discard the backfill event.

One scenario pointed out by Edwina where this approach might not be ideal:
```
1. MCE-1 for entity A goes into DLQ (emit time T1)
2. MCE-2 for entity A goes into DLQ (emit time T2)
3. backfill batches MCE-1 and MCE-2, and consolidates into MCE-2.
4. MCE-3 for entity A goes into DLQ (emit time T3)
5. backfill runs, and applies MCE-2 (modified time T4)
6. backfill runs for MCE-3 and is dropped, since T3 < T4.
```

We'll implement this approach and will keep the logic for the old schema.
Future tasks:
- Add `emitTime` column in the new schema and use that as comparison to whether we should do backfill. This should not be too much on operational side since this info will be stored in the json column
- Have the backfill tool to allow override `emitTime` so that the old schema could work

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -425,8 +425,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
       } else if (oldAuditStamp == null || !oldAuditStamp.hasTime()) {
         log.info("Event is a backfill event, but current record has no timestamp. Proceeding as normal.");
       } else if (trackingContext.getEmitTime() < oldAuditStamp.getTime()) {
-        log.info("Event's emit time: {}, which is smaller than last modified time {}. Skipping event.",
-            trackingContext.getEmitTime(), oldAuditStamp.getTime());
+        log.info("Event's emit time: {}, which is smaller than last modified time {}. Skipping event for urn {}.",
+            trackingContext.getEmitTime(), oldAuditStamp.getTime(), urn);
         return new AddResult<>(oldValue, oldValue, aspectClass);
       } else {
         log.info("Event's emit time: {}, which is larger than last modified time {}. Proceeding as normal",
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -417,20 +417,22 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     boolean isBackfillEvent = trackingContext != null
         && trackingContext.hasBackfill() && trackingContext.isBackfill();
     if (isBackfillEvent) {
-      log.info("Encounter backfill event. Tracking context: {}. Urn: {}. Aspect class: {}",
-          trackingContext, urn, aspectClass);
+      boolean shouldBackfill =
+          // the time in old audit stamp represents last modified time of the aspect
+          // if the record doesn't exist, it will be null, which means we should process the record as normal
+          oldAuditStamp != null && oldAuditStamp.hasTime()
+              // ingestionTrackingContext if not null should always have emitTime. If emitTime doesn't exist within
+              // a non-null IngestionTrackingContext, it should be investigated. We'll also skip backfilling in this case
+              && trackingContext.hasEmitTime()
+              // we should only process this backfilling event if the emit time is greater than last modified time
+              && trackingContext.getEmitTime() > oldAuditStamp.getTime();
 
-      if (!trackingContext.hasEmitTime()) {
-        log.info("Event is a backfill event, but no emitTime was given. Proceeding as normal.");
-      } else if (oldAuditStamp == null || !oldAuditStamp.hasTime()) {
-        log.info("Event is a backfill event, but current record has no timestamp. Proceeding as normal.");
-      } else if (trackingContext.getEmitTime() < oldAuditStamp.getTime()) {
-        log.info("Event's emit time: {}, which is smaller than last modified time {}. Skipping event for urn {}.",
-            trackingContext.getEmitTime(), oldAuditStamp.getTime(), urn);
+      log.info("Encounter backfill event. Tracking context: {}. Urn: {}. Aspect class: {}. Old audit stamp: {}. "
+              + "Based on this information, shouldBackfill = {}.",
+          trackingContext, urn, aspectClass, oldAuditStamp, shouldBackfill);
+
+      if (!shouldBackfill) {
         return new AddResult<>(oldValue, oldValue, aspectClass);
-      } else {
-        log.info("Event's emit time: {}, which is larger than last modified time {}. Proceeding as normal",
-            trackingContext.getEmitTime(), oldAuditStamp.getTime());
       }
     }
 
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -414,6 +414,25 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     final ASPECT oldValue = latest.getAspect() == null ? null : latest.getAspect();
     final AuditStamp oldAuditStamp = latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit();
 
+    boolean isBackfillEvent = trackingContext != null
+        && trackingContext.hasBackfill() && trackingContext.isBackfill();
+    if (isBackfillEvent) {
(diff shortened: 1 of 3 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -414,6 +414,25 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
     final ASPECT oldValue = latest.getAspect() == null ? null : latest.getAspect();
     final AuditStamp oldAuditStamp = latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit();
 
+    boolean isBackfillEvent = trackingContext != null
+        && trackingContext.hasBackfill() && trackingContext.isBackfill();
+    if (isBackfillEvent) {
+      log.info("Encounter backfill event. Tracking context: {}", trackingContext);
+
+      if (!trackingContext.hasEmitTime()) {
+        log.info("Event is a backfill event, but no emitTime was given. Proceeding as normal.");
```
jsdonn, line 423: I feel like these generic log messages are kinda useless if they don't include any information about the particular event trying to be backfilled.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -414,6 +414,25 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
+      } else if (oldAuditStamp == null || !oldAuditStamp.hasTime()) {
+        log.info("Event is a backfill event, but current record has no timestamp. Proceeding as normal.");
+      } else if (trackingContext.getEmitTime() < oldAuditStamp.getTime()) {
+        log.info("Event's emit time: {}, which is smaller than last modified time {}. Skipping event.",
```
jsdonn, line 427: Should be at least a WARN level log imo. Also would be good to include the original event or something here so we can debug.

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -436,4 +436,35 @@ public void testAtomicUpdateDisabledUsesMultipleTransactions() throws URISyntaxE
...
+
+  @Test(description = "When event is backfill event, but emit time is smaller than old audit time, it should be a no-op.")
+  public void testAddBackfillEmitTimeLargerThanOldAuditTime() throws URISyntaxException {
+    FooUrn urn = new FooUrn(1);
+    AspectFoo foo = new AspectFoo().setValue("foo");
+    IngestionTrackingContext trackingContext = new IngestionTrackingContext();
+    trackingContext.setEmitTime(5L);
+    trackingContext.setBackfill(true);
+
+    AuditStamp oldAuditStamp = makeAuditStamp("susActor", 6L);
```
jsdonn, line 448: lol

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
derekpham: I was thinking that tracking context is enough, so that when we troubleshoot, we can just query the DLQ with the tracking UUID. What kind of information do you think we should also log?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
derekpham: Should we log the whole aspect (`newValue`)? I was initially hesitant since these payloads could be very big. But given this is rare, this might be ok.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Fair enough, we can use the tracking context.

On another note, I feel like these log messages are also very verbose. We get like 2+ logs per event? Isn't that a lot if there's a big disaster and many MCEs need to be backfilled?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Fair enough, as you said we can use the trackingContext. But it may still help to add like `urn` or `aspect` at least since those are easy to get fields and would speed up debugging process a lot imo

jsdonn (approved):
LGTM, thanks for addressing comments.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
edwinalu: If we're logging skipping the backfill event, can the urn be listed?

derekpham: @edwinalu  the urn is logged in my latest commit

derekpham: oh you mean to also log it when skipping event too?

edwinalu: Yes, can you please log the urn when skipping it. If we run the backfill, but don't see the change for the entity, it would be good to be able to check the logs to know if it was skipped.

edwinalu (approved):
Thanks for adding backfill support!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -414,6 +414,26 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
     final ASPECT oldValue = latest.getAspect() == null ? null : latest.getAspect();
     final AuditStamp oldAuditStamp = latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit();
 
+    boolean isBackfillEvent = trackingContext != null
+        && trackingContext.hasBackfill() && trackingContext.isBackfill();
+    if (isBackfillEvent) {
+      log.info("Encounter backfill event. Tracking context: {}. Urn: {}. Aspect class: {}",
+          trackingContext, urn, aspectClass);
+
+      if (!trackingContext.hasEmitTime()) {
```
yangyangv2, line 423: Each if-else branch represents a scenario that leads to the decision (proceed or skip). Can you add comments on each branch to help future developers to understand the decision making context? For example:

- in what scenario, trackingContext does not have EmitTime? and the reason we should proceed the backfill.
- in what scenario, the current record has no timestamp? and the reason we should proceed the backfill
... ..

these context comment will be very useful for future work. e.g. Sooner we will see similar logics in handling bootstrap events.

yangyangv2: nit: and the branch logs are based on the value of "trackingContext" and "oldAuditStamp". would it be sufficient to just have one generic log with both variables with the decision of "proceed" or "skip"

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -414,6 +414,28 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
     final AuditStamp oldAuditStamp = latest.getExtraInfo() == null ? null : latest.getExtraInfo().getAudit();
 
+    boolean isBackfillEvent = trackingContext != null
+        && trackingContext.hasBackfill() && trackingContext.isBackfill();
+    if (isBackfillEvent) {
+      boolean shouldBackfill =
+          // the time in old audit stamp represents last modified time of the aspect
+          // if the record doesn't exist, it will be null, which means we should process the record as normal
+          oldAuditStamp != null && oldAuditStamp.hasTime()
+              // ingestionTrackingContext if not null should always have emitTime. If emitTime doesn't exist within
```
yangyangv2, line 424: @zhixuanjia , @jsdonn , @JiaoMaWHU should we throw exception for ingestionTrackingContext with null emitTime?

yangyangv2 (approved):
Thanks, much cleaner! 

Left an open question but LGTM overall.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
derekpham: we do create our own tracking context in wherehows-samza if you can't find it. It's just not used to modify the original MCE. I think throwing exception can be a bit dangerous. How about simply logging for now?

jsdonn:
Merge on behalf of Derek who is having permission issues merging.

PR #321, opened 2023-11-21
make GmaEntitiesAnnotationAllowList nullable

ybz1013:
## Summary
This change is a requirement for removing allow list in metadata-models-plugin MP, which currently uses AlwaysAllowList as a workaround to enable reuse of aspects. In case like that, the allow list check becomes trivial, so we wish to remove the check as a whole for Plugins that don't require it.

This change will let the allow list to be nullable, so when the Project doesn't have a GmaEntitiesAnnotationAllowList passed to datahub-gma, we simply skip the check.

## Test Done
./gradlew build

## Checklist


JiaoMaWHU (commented):
Code looks code, can we add proper test in `GmaAnnotationParserTest` to verify the null change?

JiaoMaWHU (approved):

PR #325, opened 2023-11-28
initial commit for adding type agnostic base restli resource

JiaoMaWHU:
## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -61,6 +61,7 @@ import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 
 import static com.linkedin.metadata.dao.utils.IngestionUtils.*;
+import static com.linkedin.metadata.dao.utils.ModelUtils.*;
 
 
 /**
@@ -76,6 +77,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     extends BaseReadDAO<ASPECT_UNION, URN> {
 
   private final Class<ASPECT_UNION> _aspectUnionClass;
+
+  private final Class<URN> _urnClass;
 
   /**
    * Immutable class that corresponds to the metadata aspect along with {@link ExtraInfo} for the same metadata. It also
@@ -184,13 +187,15 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    *     com.linkedin.metadata.aspect
    * @param producer {@link BaseMetadataEventProducer} for the metadata event producer
    */
-  public BaseLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer) {
+  public BaseLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer,
+      @Nonnull Class<URN> urnClass) {
     super(aspectUnionClass);
     _producer = producer;
     _storageConfig = LocalDAOStorageConfig.builder().build();
     _aspectUnionClass = aspectUnionClass;
     _trackingManager = null;
     _trackingProducer = null;
+    _urnClass = urnClass;
   }
 
   /**
@@ -202,13 +207,14 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    * @param trackingManager {@link BaseTrackingManager} for managing tracking requests
    */
   public BaseLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseTrackingMetadataEventProducer trackingProducer,
-      @Nonnull BaseTrackingManager trackingManager) {
+      @Nonnull BaseTrackingManager trackingManager, @Nonnull Class<URN> urnClass) {
     super(aspectUnionClass);
     _producer = null;
     _storageConfig = LocalDAOStorageConfig.builder().build();
     _aspectUnionClass = aspectUnionClass;
     _trackingManager = trackingManager;
     _trackingProducer = trackingProducer;
+    _urnClass = urnClass;
   }
 
   /**
@@ -224,6 +230,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     _aspectUnionClass = producer.getAspectUnionClass();
     _trackingManager = null;
     _trackingProducer = null;
+    _urnClass = null;
   }
 
   /**
@@ -242,6 +249,17 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     _aspectUnionClass = trackingProducer.getAspectUnionClass();
     _trackingManager = trackingManager;
     _trackingProducer = trackingProducer;
+    _urnClass = null;
+  }
+
+  /**
+   * Get the urn class.
+   *
+   * @return the urn class
+   */
+  @Nullable
+  public Class<URN> getUrnClass() {
+    return this._urnClass;
   }
 
   /**
@@ -1090,14 +1108,50 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    */
   @Nonnull
   public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> backfill(
-      @Nonnull BackfillMode mode, @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses, @Nonnull Set<URN> urns) {
-    checkValidAspects(aspectClasses);
+      @Nonnull BackfillMode mode, @Nullable Set<Class<? extends RecordTemplate>> aspectClasses, @Nonnull Set<URN> urns) {
+    Set<Class<? extends RecordTemplate>> aspectToBackfill = aspectClasses;
+    if (aspectClasses == null) {
+      aspectToBackfill = getValidAspectTypes(_aspectUnionClass);
+    }
+    checkValidAspects(aspectToBackfill);
     final Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> urnToAspects =
-        get(aspectClasses, urns);
+        get(aspectToBackfill, urns);
     urnToAspects.forEach((urn, aspects) -> {
       aspects.forEach((aspectClass, aspect) -> aspect.ifPresent(value -> backfill(mode, value, urn)));
     });
     return urnToAspects;
+  }
+
+  @Nonnull
+  public Map<String, Set<String>> backfillMAE(@Nonnull BackfillMode mode, @Nullable Set<String> aspects,
+      @Nonnull Set<String> urns) {
+    // convert string to entity urn
+    if (_urnClass == null) {
+      throw new IllegalStateException("urn class is null, unable to convert string to urn");
+    }
+    final Set<URN> urnSet = urns.stream().map(x -> getUrnFromString(x, _urnClass)).collect(Collectors.toSet());
+
+    // convert string to aspect class
+    Set<Class<? extends RecordTemplate>> aspectSet = null;
+    if (aspects != null) {
+      aspectSet = aspects.stream().map(ModelUtils::getAspectClass).collect(Collectors.toSet());
+    }
+
+    final Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> results = backfill(mode, aspectSet, urnSet);
+    final Map<String, Set<String>> mapToReturn = new HashMap<>();
+    for (URN urn: results.keySet()) {
+      Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>> value = results.get(urn);
+      Set<String> backfilledAspects = new HashSet<>();
+      for (Class<? extends RecordTemplate> clazz: value.keySet()) {
+        if (value.get(clazz).isPresent()) {
+          backfilledAspects.add(clazz.getCanonicalName());
+        }
+      }
+      if (!backfilledAspects.isEmpty()) {
+        mapToReturn.put(String.valueOf(urn), backfilledAspects);
+      }
+    }
+    return mapToReturn;
   }
 
   @Nonnull
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
@@ -1,0 +1,53 @@
+package com.linkedin.metadata.restli;
+
+import com.linkedin.common.urn.Urn;
+import com.linkedin.data.template.UnionTemplate;
+import com.linkedin.metadata.backfill.BackfillMode;
+import com.linkedin.metadata.dao.BaseLocalDAO;
+import com.linkedin.metadata.events.IngestionMode;
+import com.linkedin.restli.client.RestLiResponseException;
+import com.linkedin.restli.common.ErrorResponse;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import static com.linkedin.metadata.dao.utils.IngestionUtils.*;
+
+
+/**
+ * An base class that has entity agnostic operations such as backfill.
+ */
+public abstract class BaseEntityAgnosticResource {
+
+  @Nonnull
+  protected abstract LocalDaoRegistry getLocalDaoRegistry();
+
+  public Map<String, Set<String>> backfillMAE(
+      @Nonnull String entity,
+      @Nonnull Set<String> urns,
+      @Nullable Set<String> aspects,
+      @Nonnull IngestionMode ingestionMode
+  ) throws RestLiResponseException {
+    LocalDaoRegistry localDaoRegistry = getLocalDaoRegistry();
+    BaseLocalDAO<? extends UnionTemplate, ? extends Urn> dao = localDaoRegistry.getLocalDaoByEntity(entity);
+    if (dao == null) {
+      throw new RestLiResponseException(
+        new ErrorResponse()
+          .setStatus(404)
+          .setMessage(String.format("Dao not found for the requested entity: %s", entity))
+      );
+    }
+
+    BackfillMode backfillMode = ALLOWED_INGESTION_BACKFILL_BIMAP.get(ingestionMode);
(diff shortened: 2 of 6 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -184,13 +187,15 @@ public static class AspectUpdateLambda<ASPECT extends RecordTemplate> {
...
-  public BaseLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer) {
+  public BaseLocalDAO(@Nonnull Class<ASPECT_UNION> aspectUnionClass, @Nonnull BaseMetadataEventProducer producer,
+      @Nonnull Class<URN> urnClass) {
     super(aspectUnionClass);
     _producer = producer;
     _storageConfig = LocalDAOStorageConfig.builder().build();
     _aspectUnionClass = aspectUnionClass;
     _trackingManager = null;
     _trackingProducer = null;
+    _urnClass = urnClass;
```
yangyangv2, line 198: you don't need to pass the urnClass as a parameter, BaseLocalDAO already has a generic type URN, you can derive the _urnClass with ParameterizedType

http://www.java2s.com/Tutorials/Java/java.lang.reflect/ParameterizedType/0040__ParameterizedType.getActualTypeArguments_.htm

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1087,19 +1105,53 @@ public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTe
...
    * @param mode backfill mode to scope the backfill process
+   * @param aspectClasses set of aspects to backfill, if null, all valid aspects inside the entity snapshot will be backfilled
+   * @param urns  set of urns to backfill
    */
   @Nonnull
   public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> backfill(
-      @Nonnull BackfillMode mode, @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses, @Nonnull Set<URN> urns) {
-    checkValidAspects(aspectClasses);
+      @Nonnull BackfillMode mode, @Nullable Set<Class<? extends RecordTemplate>> aspectClasses, @Nonnull Set<URN> urns) {
+    Set<Class<? extends RecordTemplate>> aspectToBackfill = aspectClasses;
```
yangyangv2, line 1114: nit: 

`final Set<Class<? extends RecordTemplate>> aspectToBackfill = (aspectClasses == null)? getValidAspectTypes(_aspectUnionClass): aspectClasses;
    }`

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1314,4 +1366,27 @@ protected void validateAgainstSchema(@Nonnull RecordTemplate model) {
       throw new ModelValidationException(result.getMessages().toString());
     }
   }
+
+  /**
+   * Maps backfill results of type map{urn, map{aspect, optional metadata}} to map{urn, aspect fqcn}.
+   */
+  protected Map<String, Set<String>> transformBackfillResultsToStringMap(
```
yangyangv2, line 1373: what is the reasoning for returning as string type rather than a more generic type such as Urn or Class? 

or Why backfillMAE has to return <String, Set<String>> instead of <Urn, Set<Class>>?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,85 @@
...
+   *                                           should represent no-op, empty map will be returned. Backfill is to redo
+   *                                           any metadata update that is missed or skipped in the past.
+   *                                           Bootstrap indicates building the metadata from scratch.
+   * @return a map of urn to the set of aspects that are backfilled
+   * @throws RestLiResponseException for any error
+   */
+  @Action(name = ACTION_BACKFILL_MAE)
+  @Nonnull
+  public Map<String, Set<String>> backfillMAE(
+      @Nonnull String entity,
```
yangyangv2, line 52: entity is not necessary. It could be self conflict (unmatched entity and urns) and becomes a restriction if we want to support multi-type urn inputs

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,85 @@
...
+      @Nonnull Set<String> urns,
+      @Nullable Set<String> aspects,
+      @Nonnull IngestionMode ingestionMode
+  ) throws RestLiResponseException {
+    final BaseLocalDAO<? extends UnionTemplate, ? extends Urn> dao = getLocalDaoByEntity(entity);
```
yangyangv2, line 57: move this lookup after backfillMode check

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,85 @@
...
+    final BackfillMode backfillMode = ALLOWED_INGESTION_BACKFILL_BIMAP.get(ingestionMode);
+    if (backfillMode == null) {
+      return Collections.emptyMap();
+    }
+    try {
```
yangyangv2, line 62: how about we make this logic work like this?

1. remove the entity parameter
2. group urns into <entity, Set<urn>> map
3. for each <entity, Set<urn>>, find the DAO and run the backfill
4. merge the result with <urn, aspect> list.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
JiaoMaWHU: Yes, we could get the Type of _urnClass, and get the class like `Class<?> urnClazz = (Class<?>) Type;`.

But I see passing `_urnClass` being simpler than ^, it was already registered in `EbeanLocalDao`, just we didn't pass in. And also, `_aspectUnionClass` was passed in using the same way, so it doesn't bring any trouble to do the same thing for `_urnClass`. WDYT.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
JiaoMaWHU: good suggestion, will update

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
JiaoMaWHU: sure

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
JiaoMaWHU: is the understanding right that we want to update input parameters to: `Map<urnStr, Set<aspectStr>>, IngestionMode`? If yes, it looks like multi entity is what we want this API to achieve.

What would be the benefit of enabling multi entity in this case? My idea to make this API as simple and intuitive as possible while drafting the PR, but I'm open to implement it in a different way.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Helper method to get the {@link BaseLocalDAO} from class {@link LocalDaoRegistry} for the given entity type.
+   */
+  protected BaseLocalDAO<? extends UnionTemplate, ? extends Urn> getLocalDaoByEntity(String entity)
+      throws RestLiResponseException {
+    LocalDaoRegistry localDaoRegistry = getLocalDaoRegistry();
+    BaseLocalDAO<? extends UnionTemplate, ? extends Urn> dao = localDaoRegistry.getLocalDaoByEntity(entity);
+    if (dao == null) {
+      throw new RestLiResponseException(new ErrorResponse().setStatus(404)
```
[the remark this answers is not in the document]
JiaoMaWHU: I'm all good to make this API finish up as many backfills as possible if we decide to go with multi entity route.

restli-resources/src/main/java/com/linkedin/metadata/restli/LocalDaoRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+import com.linkedin.metadata.dao.utils.ModelUtils;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+
+/**
+ * An immutable class to store and access the mapping from the entity type string to the entity's LocalDao.
+ */
+public class LocalDaoRegistry {
```
zhixuanjia, line 15: Shall we use an interface for DAO registry? `public interface localdaoregistry` and leave the concrete implementation to each GMS service?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
zhixuanjia: Use the logic suggested by Yang. Or have checks to enforce the `Set<String> urns` only contains urn for same entity type.

restli-resources/src/main/java/com/linkedin/metadata/restli/LocalDaoRegistry.java
JiaoMaWHU: Sounds good. I can split the code into two parts:
- `public interface localdaoregistry` with `getLocalDaoByEntity` defined
- a `BaseLocaldaoRegistry` class implements the interface so that ppl can just use this class if they have no custom logic.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
JiaoMaWHU: `have checks to enforce the Set<String> urns only contains urn for same entity type` yes, this PR supports that.

I will maybe sync with Yang and decide on this.

yangyangv2: the benefit is in a batch backfill process, it will reduce the requests sending to the backfill service. And the backfill automation logic could be type agnostic and batch firing the backfill requests from the DLQ.

good callout the different urn types may have different aspect sets. I think `(urnStr, @Optional aspectStr)[]` should be the right parameter type. 

Human users can send (urnStr, null)[] if they don't want (or don't know how) to specific the aspects

JiaoMaWHU: makes sense, I will refactor the code to make it support multi entity

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
JiaoMaWHU: will remove entity parameter

restli-resources/src/main/java/com/linkedin/metadata/restli/LocalDaoRegistry.java
zhixuanjia: Sounds good. I would suggest name the implementation to `DefaultLocaldaoRegistryImpl`

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
JiaoMaWHU: I did try casting it to `Urn` type in resource layer and pass it to Dao, but it didn't work out because Dao method requires `URN extends Urn` type.

I don't think we can cast a string to `URN` outside of Dao layer because generic type is only defined in Dao, also  we cannot use reflection because of Java Type Erasure.

shall we go with the implemented approach?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
JiaoMaWHU: discussed offline, we will go with current approach since it's simpler and does not have unchecked warnings.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
JiaoMaWHU: discussed offline, will go with the current approach since casting it to URN before Dao layer is impossible per my research

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1087,19 +1105,51 @@ public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTe
...
+   * @param mode backfill mode to scope the backfill process
+   * @param aspects set of aspects to backfill
+   * @param urns set of urns to backfill
+   * @return map of urn to their backfilled aspect values
+   */
+  @Nonnull
+  public Map<String, Set<String>> backfillMAE(@Nonnull BackfillMode mode, @Nullable Set<String> aspects,
+      @Nonnull Set<String> urns) {
+    // convert string to entity urn
+    if (_urnClass == null) {
```
yangyangv2, line 1138: can we comment on in which scenario, _urnClass can be null?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1087,19 +1105,51 @@ public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTe
...
+    final Set<URN> urnSet = urns.stream().map(x -> getUrnFromString(x, _urnClass)).collect(Collectors.toSet());
+
+    // convert string to aspect class
+    Set<Class<? extends RecordTemplate>> aspectSet = null;
+    if (aspects != null) {
+      aspectSet = aspects.stream().map(ModelUtils::getAspectClass).collect(Collectors.toSet());
+    }
+
+    // call type specific backfill method and transform results to string map
+    return transformBackfillResultsToStringMap(backfill(mode, aspectSet, urnSet));
```
yangyangv2, line 1150: can backfill take null aspectSet? Should we make it @Nonnull to for more clarity?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,119 @@
...
+    final BackfillMode backfillMode = ALLOWED_INGESTION_BACKFILL_BIMAP.get(ingestionMode);
+    if (backfillMode == null) {
+      return backfillResults;
+    }
+
+    // Group requests by entity type
+    final Map<String, List<BackfillItem>> entitytoRequestsMap = new HashMap<>();
+    backfillRequests.forEach(request -> {
+      try {
+        final String entity = Urn.createFromString(request.getUrn()).getEntityType();
```
yangyangv2, line 68: nit: can we name it as entityType for better readability

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
```diff
@@ -0,0 +1,119 @@
...
+      } catch (URISyntaxException e) {
+        log.warn("Failed casting string to Urn, request: " + request, e);
+      }
+    });
+
+    // for each entity, backfill MAE for each urn in parallel
+    for (String entity : entitytoRequestsMap.keySet()) {
+      final Optional<BaseLocalDAO<? extends UnionTemplate, ? extends Urn>> dao = getLocalDaoByEntity(entity);
+      if (!dao.isPresent()) {
+        log.warn("LocalDAO not found for entity: " + entity);
```
yangyangv2, line 79: nit: "for entity type: " + entityType

restli-resources/src/main/java/com/linkedin/metadata/restli/dao/DefaultLocalDaoRegistryImpl.java
```diff
@@ -0,0 +1,59 @@
...
+   * by the URN class of each BaseLocalDAO. If any inconsistency is detected (i.e., the key does not match the expected
+   * entity type from the DAO's URN class), an IllegalArgumentException is thrown.
+   *
+   * @param entityToLocalDaoMap A non-null map associating entity type strings with their
+   *                            corresponding BaseLocalDAO instances.
+   * @throws IllegalArgumentException if there is a mismatch between an entity type key and its
+   *                                  expected entity type from the DAO's URN class.
+   */
+  public DefaultLocalDaoRegistryImpl(@Nonnull Map<String, BaseLocalDAO<? extends UnionTemplate, ? extends Urn>> entityToLocalDaoMap) {
+    entityToLocalDaoMap.forEach((key, value) -> {
```
yangyangv2, line 35: if this is created by a factory, can we make the initialization into a init() method rather than in the constructor? 

https://medium.com/@luketong/best-practices-for-constructor-methods-in-java-8fa799dd6f8b

restli-resources/src/main/java/com/linkedin/metadata/restli/dao/LocalDaoRegistry.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public interface LocalDaoRegistry {
+
+  /**
+   * Returns the {@link BaseLocalDAO} registered on the given entity type.
+   *
+   * @param entity the entity type string
+   * @return the {@link BaseLocalDAO} registered on the given entity type, or null if no such registration.
+   */
+  BaseLocalDAO<? extends UnionTemplate, ? extends Urn> getLocalDaoByEntity(@Nonnull String entity);
```
yangyangv2, line 20: nit: let's use entityType universally in this PR. entity is amubigious

JiaoMaWHU: done

restli-resources/src/main/java/com/linkedin/metadata/restli/dao/DefaultLocalDaoRegistryImpl.java
JiaoMaWHU: good suggestion! will update

JiaoMaWHU:
merge for now, will confirm with @zhixuanjia offline for any other concern and address it as fast follow

PR #342, opened 2024-01-18
(feat) support clear relationship by source entity urn

yangyangv2:
## Checklist

This feature is to allow clear relationships from a given source entity. Context for this feature is to solve Issue 2 in this doc:

https://docs.google.com/document/d/1WwXdos5BR7cxpDzxpArofRgj0_cdD9QGmjiZj7pQd14/edit#heading=h.v2d5gj2wsv95

- [] Docs related to the changes have been added/updated (if applicable)

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
@@ -5,7 +5,6 @@ import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.metadata.dao.internal.BaseGraphWriterDAO;
 import java.util.List;
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import lombok.Value;
 
 
@@ -15,22 +14,16 @@ import lombok.Value;
 public abstract class BaseLocalRelationshipBuilder<ASPECT extends RecordTemplate> {
 
   private final Class<ASPECT> _aspectClass;
-  private final Class<? extends RecordTemplate>[] _relationshipClasses;
 
   @Value
   public static class LocalRelationshipUpdates {
     List<? extends RecordTemplate> relationships;
+    Class<? extends RecordTemplate>[] relationshipClasses;
     BaseGraphWriterDAO.RemovalOption removalOption;
   }
 
   public BaseLocalRelationshipBuilder(@Nonnull Class<ASPECT> aspectClass) {
-    this(aspectClass, null);
-  }
-
-  public BaseLocalRelationshipBuilder(@Nonnull Class<ASPECT> aspectClass,
-      @Nonnull Class<? extends RecordTemplate>[] relationshipClasses) {
     _aspectClass = aspectClass;
-    _relationshipClasses = relationshipClasses;
   }
 
   /**
@@ -46,12 +39,4 @@ public abstract class BaseLocalRelationshipBuilder<ASPECT extends RecordTemplate
    */
   @Nonnull
   public abstract <URN extends Urn> List<LocalRelationshipUpdates> buildRelationships(@Nonnull URN urn, @Nonnull ASPECT aspect);
-
-  /**
-   * Returns a list of relationship classes used in this builder.
-   */
-  @Nullable
-  public Class<? extends RecordTemplate>[] getSupportedRelationshipClasses() {
-    return _relationshipClasses;
-  }
 }
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
@@ -5,6 +5,7 @@ import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.metadata.dao.internal.BaseGraphWriterDAO;
 import java.util.List;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import lombok.Value;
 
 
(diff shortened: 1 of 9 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -55,6 +56,28 @@ public <ASPECT extends RecordTemplate> void processLocalRelationshipUpdates(
...
+   * @param supportedRelationshipClasses relationship classes needs to be cleared
+   * @param <RELATIONSHIP> relationship model
+   */
+  @Transactional
+  public <RELATIONSHIP extends RecordTemplate> void clearRelationshipsBySource(@Nonnull Urn sourceUrn,
+      @Nullable Class<? extends RecordTemplate>[] supportedRelationshipClasses) {
+    if (supportedRelationshipClasses == null || supportedRelationshipClasses.length == 0) {
+      return;
+    }
+    for (Class<? extends RecordTemplate> relationshipClass : supportedRelationshipClasses) {
```
zhixuanjia, line 71: can we just pass the relationship class as a param to the method?

dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
```diff
@@ -18,6 +18,7 @@ public abstract class BaseLocalRelationshipBuilder<ASPECT extends RecordTemplate
   @Value
   public static class LocalRelationshipUpdates {
     List<? extends RecordTemplate> relationships;
+    Class<? extends RecordTemplate>[] relationshipClasses;
```
zhixuanjia, line 21: I think there is potentially mismatch between relationships and relationshipClasses.

do you feel this is better?
```
@Value
public static class LocalRelationshipUpdate<RELATIONSHIP extends RecordTemplate> {
  RELATIONSHIP relationship;
  Class<? extends RecordTemplate> relationshipClass;
}

@Value
public static class LocalRelationshipUpdates {
  List<LocalRelationshipUpdate> updates;
  BaseGraphWriterDAO.RemovalOption removalOption;
}
```
Downside is this will be a breaking change but user can found out at build time and fix forward.

yangyangv2: I can see the benefit of this, but 1) it will cause more DAO API updates, 2) it still doesn't solve the issue if `updates.isEmpty()`, then where to get the relationshipClass to update the relationship table.

Would prefer to stay with the current implementation and give the Builder owner the flexibility, responsiblity to set the correct relationship table to clean up.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -47,11 +47,47 @@ public EbeanLocalRelationshipWriterDAO(EbeanServer server) {
...
+      }
+    }
+  }
+
+  /**
+   * This method is to serve for the purpose to clear all the relationships from a source entity urn.
+   * @param urn entity urn could be either source or destination, depends on the RemovalOption
+   * @param supportedRelationshipClasses relationship classes needs to be cleared
+   */
+  @Transactional
```
zhixuanjia, line 67: Transactional annotation is not needed as the parent method processLocalRelationshipUpdates already has it.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -47,11 +47,47 @@ public EbeanLocalRelationshipWriterDAO(EbeanServer server) {
...
+      } else {
+        addRelationships(relationshipUpdate.getRelationships(), relationshipUpdate.getRemovalOption());
+      }
+    }
+  }
+
+  /**
+   * This method is to serve for the purpose to clear all the relationships from a source entity urn.
+   * @param urn entity urn could be either source or destination, depends on the RemovalOption
+   * @param supportedRelationshipClasses relationship classes needs to be cleared
```
zhixuanjia, line 65: `relationship classes needs to be cleared` seems doesn't make sense, what does it mean to clear a class.

dao-api/src/main/java/com/linkedin/metadata/dao/builder/BaseLocalRelationshipBuilder.java
zhixuanjia: sounds good.

zhixuanjia (approved):

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
yangyangv2: what about ``` relationship that needs to be cleared```?

zhixuanjia (approved):
thanks for addressing the comment.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
jphui, line 1: Is there a reason you didn't override
```
  @Override
  public <RELATIONSHIP extends RecordTemplate> void removeRelationships(@Nonnull List<RELATIONSHIP> relationships) {
    throw new UnsupportedOperationException("Local relationship only supports adding relationships.");
  }
```
and created the "clearRelationships" method?

yangyangv2: The behaviors are different. 

- Remove means given a list of "specific relationship targets", please remove them.
- Clear means, I don't know what "specific relationship targets" are, but I'll give you the entity reference, please remove all the relationships.

PR #350, opened 2024-02-05
feat(annotations): Add Search Index annotations

jphui:
# Checklist

# Overview
[[META-19971]](https://jira01.corp.linkedin.com:8443/browse/META-19971)

This PR introduces the `@gma.search.index` annotation that can be applied to fields in Aspect model PDL's to support No Code Gen Search Onboarding.

## E2E Experience
See [search.md](https://github.com/linkedin/datahub-gma/tree/jhui/searchAnnotation/docs/annotations/search.md).

## Important Docs
* [RFC - No Code Gen Search Onboarding](https://docs.google.com/document/d/1QgSnP8td8yWft_E-KBxt4wmYg2yWKReuH-JtCWa-4DY/edit)

# Implementation Approach

> The high-level idea for this PR was to _**keep as much of the existing parsing behavior and infrastructure as possible**_. My mindset was to reduce "how much new stuff I added" and instead try to _build off of and extend_ existing functionality.

## Structure

This plays out practically by keeping the core courier `GmaAnnotation` and first adding an `optional SearchAnnotation`.

### `SearchAnnotation` ([link](https://github.com/linkedin/datahub-gma/blob/jhui/searchAnnotation/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/SearchAnnotation.pdl))
is introduced to serve as a container for Search-related annotation metadata. For now, it only contains a mapping of _field names_ -> _IndexAnnotations_:
```
index: optional map[string, array[IndexAnnotation]]
```
... but can be easily extended in the future to incorporate new top-level or field-level annotations.

**NOTE**: _GmaAnnotation_ is used to hold parsed annotation metadata across more than just _record_ [schemas](https://linkedin.github.io/rest.li/pdl_schema) (it's used at least for _typerefs_ when parsing the AspectUnion PDL's), so _SearchAnnotation's_ `index` map will only be populated for _record_ types since it's the only type with "fields".

### `IndexAnnotation` ([link](https://github.com/linkedin/datahub-gma/blob/jhui/searchAnnotation/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/IndexAnnotation.pdl))
is introduced to serve as a container for field-level indexing metadata. For now, it only contains one field:
```
urn: optional string
```
... but can easily be extended in the future to incorporate field-level customer preferences for indexing.

## Flow of parsing logic

> The overall idea for the new logic is to:
> 1. do all the original parsing as-it-was
> 2. parse the fields and "fill out GmaAnnotation" with the new annotation metadata

Unfortunately, the original parsing methodology relied heavily on the _singular_ and _top-level_ nature of GMA annotations, meaning it was a simple "extract and cast" process. However, both of those assumptions are broken with this new annotation since it can _appear multiple times_ and is at the _field level_. Thus, the introduced logic is not as straightforward.

In [GmaAnnotationParser.java](https://github.com/linkedin/datahub-gma/blob/jhui/searchAnnotation/gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java):

1. abstract existing `parse()` functionality --> _`parseTopLevelAnnotations()`_
2. create a new `parseSearchIndexFields()` that iterates through the fields and returns a `field to annotations map`
3. refactor `parse()` to combine (1) and (2) with pseudocode that looks something like:
```
parse():
    x = originalParsing()
    y = newFieldParsing()
    if (y has stuff)
        x.append(y)
    return x
```

# Testing

In [GmaAnnotationParserTest.java](https://github.com/linkedin/datahub-gma/blob/jhui/searchAnnotation/gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java), I add a new test [_parseAspectWithOnlySearchIndexAnnotations()_](https://github.com/linkedin/datahub-gma/blob/641bc55e16f66227518f9de024b4fffab462b7f7/gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java#L84) that checks to make sure the _@gma.search.index_ annotations in [_SearchAnnotatedBarAspect.pdl_](https://github.com/linkedin/datahub-gma/blob/jhui/searchAnnotation/gradle-plugins/metadata-annotations-test-models/src/main/pegasus/com/linkedin/testing/SearchAnnotatedAspectBar.pdl) are recognized.

* single, multiple, and no _@gma.search.index_ annotation cases are tested.

* Note that this Aspect model does not have the _@gma.aspect.entities_ annotation, yet it still passes the test. See _What checks should be done?_ below to read discussion about whether this is "acceptable".

Backwards compatibility with existing tests was also ensured.

# Open Discussion / Considerations

### _What checks should be done?_

In short, should we prevent certain _@gma.search.index_ annotations from happening, and if so, should it be at this -- the _datahub-gma_ -- level?

Checks like:
* the urn's in the _@gma.search.index_ annotation have to be inside the _@gma.search.entities_ annotation as well
* can only annotate simple field types (see RFC's _Limitations_ section)

_**Current answer**_: **None**. We either defer checking to the layer above _datahub-gma_ (the Code Gen plugin perhaps) or simply take a _"Don't cry about it"_ mindset: unsupported annotations are just ignored, no errors needed.

### _Should non-annotation metadata be parsed?_

Since we are introducing a "`Map< Field, Annotations >`", we have inherently already broken a little bit into the realm of _"non-annotation metadata"_.

The question is, should we also extend _GmaAnnotation_ to be able to hold metadata like _field fypes_? Or, do we keep it as Annotation-centric as possible?

_**Current answer**_: Keep it Annotation-centric. Field types can be handled elsewhere. We're modifying the _Annotation_Parser after all.

fieldtypes? (extend semantics to be "metadata parsing") --> could use a single GmaAnnotation object to generate everything?

# Useful Reading

* https://linkedin.github.io/rest.li/pdl_schema


gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/IndexAnnotation.pdl
```diff
@@ -0,0 +1,11 @@
+namespace com.linkedin.metadata.annotations
+
+/**
+ * Annotation on a field in an Aspect model indicating the field's Search configuration.
+ */
+record IndexAnnotation {
```
derekpham, line 6: not sure if I understand this yet. Shouldn't the field in `IndexAnnotation` be the name of the field instead of the urn?

Would you mind adding description to the PR on how users should expect to use the annotation and how it will be represented through your modeling?

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
```diff
@@ -34,11 +39,36 @@ public GmaAnnotationParseException(String message) {
...
    */
   public Optional<GmaAnnotation> parse(@Nonnull DataSchema schema) {
+    Optional<GmaAnnotation> gmaAnnotation = parseTopLevelAnnotations(schema);
+
+    // if annotation ecosystem grows, can abstract the following into a subroutine
+    final IndexAnnotationArrayMap searchIndexFields = parseSearchIndexFields(schema);
+    if (!searchIndexFields.isEmpty()) {
+      // we need to augment GmaAnnotation with SearchIndex metadata
+      if (!gmaAnnotation.isPresent()) {
+        // no top-level annotations, so we need to create one and fill out SearchIndex metadata
```
derekpham, line 54: do we need this comment?

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
```diff
@@ -34,11 +39,36 @@ public GmaAnnotationParseException(String message) {
...
+        gmaAnnotation = Optional.of(new GmaAnnotation().setSearch(new SearchAnnotation().setIndex(searchIndexFields)));
+      } else {
+        // yes top-level annotations, so we need to just fill out SearchIndex metadata
```
derekpham, line 57: same here

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
```diff
@@ -34,11 +39,36 @@ public GmaAnnotationParseException(String message) {
...
   public Optional<GmaAnnotation> parse(@Nonnull DataSchema schema) {
+    Optional<GmaAnnotation> gmaAnnotation = parseTopLevelAnnotations(schema);
+
+    // if annotation ecosystem grows, can abstract the following into a subroutine
+    final IndexAnnotationArrayMap searchIndexFields = parseSearchIndexFields(schema);
+    if (!searchIndexFields.isEmpty()) {
+      // we need to augment GmaAnnotation with SearchIndex metadata
+      if (!gmaAnnotation.isPresent()) {
+        // no top-level annotations, so we need to create one and fill out SearchIndex metadata
+        gmaAnnotation = Optional.of(new GmaAnnotation().setSearch(new SearchAnnotation().setIndex(searchIndexFields)));
```
derekpham, line 55: can we just return here? `return Optional.of(...)`

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
```diff
@@ -66,4 +96,39 @@ public Optional<GmaAnnotation> parse(@Nonnull DataSchema schema) {
 
     return Optional.of(gmaAnnotation);
   }
+
+  /**
+   * Obtains the {@code @gma} search annotations from the schema.
+   * Currently does not perform any validation checks.
+   */
+  private @Nonnull IndexAnnotationArrayMap parseSearchIndexFields(@Nonnull DataSchema schema) {
+    Map<String, IndexAnnotationArray> javaMap = new HashMap<>();
```
derekpham, line 105: any reason why you name it `javaMap`?

derekpham: how about `fieldNameToEntityUrnClassNames`

gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java
```diff
@@ -76,4 +79,21 @@ public void parseCommonAspectAllowedWithNullAllowList() {
         new AspectEntityAnnotationArray(new AspectEntityAnnotation().setUrn("com.linkedin.testing.FooUrn"),
             new AspectEntityAnnotation().setUrn("com.linkedin.testing.BarUrn")))));
   }
+
+  @Test
+  public void parseAspectWithOnlySearchIndexAnnotations() {
+    final Optional<GmaAnnotation> gma =
+            new GmaAnnotationParser().parse((RecordDataSchema) DataTemplateUtil.getSchema(SearchAnnotatedAspectBar.class));
+     assertThat(gma).contains(new GmaAnnotation().setSearch(
```
derekpham, line 87: `containsExactly`?

```
List foo = {"a", "b", "c"}

assertThat(foo).contains("a", "b") -> pass
assertTath(foo).containsExactly("a", "b") -> fail
```

jphui: Since we're calling `.contains()` on an Optional, the passed object will be checked for equality. I can clarify if needed but the check that is performed is in effect a "containsExactly"

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
jphui: Keeping comments just for extra extra clarity.

gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java
derekpham: ah right

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
derekpham: i feel like calling `setSearch` makes your intention clear enough. Up to you tho

gradle-plugins/metadata-annotations-lib/src/main/java/com/linkedin/metadata/annotations/GmaAnnotationParser.java
```diff
@@ -34,11 +39,36 @@ public GmaAnnotationParseException(String message) {
...
    *
-   * @throws GmaAnnotationParseException if the provided {@code @gma} annotation does not match the schema.
+   * @throws GmaAnnotationParseException upon failure.
    */
   public Optional<GmaAnnotation> parse(@Nonnull DataSchema schema) {
+    Optional<GmaAnnotation> gmaAnnotation = parseTopLevelAnnotations(schema);
+
+    // if annotation ecosystem grows, can abstract the following into a subroutine
+    final IndexAnnotationArrayMap searchIndexFields = parseSearchIndexFields(schema);
+    if (!searchIndexFields.isEmpty()) {
```
zhixuanjia, line 51: does `!searchIndexFields.isEmpty()` implies gmaAnnotation.isPresent() is true? In other words, do we still need to check on gmaAnnotation.isPresent()?

jphui: Not necessarily, there could be accidental use cases where the customer annotates a new Aspect with _@gma.search.index_ without any of the normal _@gma.aspect.entity_ or other annotations. In such a case, we want to avoid both Crashing and Not Returning _any_ annotations (since that would be incorrect parsing logic).

PR #344, opened 2024-01-26
feat: Add BaseVersionedAspect.pdl

schuangv2:
## Summary
Adding BaseVersionedAspect.pdl. It can be included by other aspects to add semantic versioning to the aspect. 

## Details
**Context**: The Datahub backend can be updated with Kafka Metadata Change Events (MCEs). However MCEs with outdated metadata may be received by the backend, eg a developer testing an older version of a project. Thus we want to (1) add semantic versioning to aspects and (2) add logic for datahub to process MCEs based on the version of the MCE and aspect in backend. This PR covers (1).

**Internal design doc**: https://docs.google.com/document/d/1Von4_q_CtvGrdMDfWHHVCcqmGyFT9CMlmXoeZAO59Oo/edit?usp=sharing

## Testing
* **Build test**: Tested with `./gradlew build` as part of github checks

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Tests for the changes have been added/updated (if applicable)
- [x] Docs related to the changes have been added/updated (if applicable)

PR #352, opened 2024-02-06
feat: Aspect Version Ingestion Logic

schuangv2:
## Summary
Adding aspect version ingestion logic. 

## Details
**Context**: The Datahub backend can be updated with Kafka Metadata Change Events (MCEs). However MCEs with outdated metadata may be received by the backend, eg a developer testing an older version of a project. Thus we want to (1) add optional semantic versioning to aspects and (2) add logic for datahub to process MCEs based on the version of the MCE and aspect in backend. (1) was covered in #344. This PR covers (2).

BUG=[META-20008](https://jira01.corp.linkedin.com:8443/browse/META-20008)

**Internal design docs**: 
- https://docs.google.com/document/d/1Von4_q_CtvGrdMDfWHHVCcqmGyFT9CMlmXoeZAO59Oo/edit?usp=sharing
- https://docs.google.com/document/d/18dV-DAwZvwA0ND0V5vPC-dmTMABi6EHOAqRUj8ig8a4/edit#heading=h.ho6frsjj5w78

## Testing
* **Build test**: Tested with `./gradlew build`.
* **Unit tests**: Added and passed unit tests.

## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -449,9 +450,11 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
       }
     }
 
-    // Skip saving if there's no actual change
-    if ((oldValue == null && newValue == null) || oldValue != null && newValue != null
-        && equalityTester.equals(oldValue, newValue)) {
+    // Skip saving for the following scenarios
+    if ((oldValue == null && newValue == null) // values are null
+        || aspectVersionComparator(newValue, oldValue) < 0 // newValue ver < oldValue ver
+        || oldValue != null && newValue != null && equalityTester.equals(oldValue, newValue) // values are equal
```
edwinalu, line 456: Would the semantic version be part of the value (oldValue and newValue)?

The logic is correct, but could you please put parentheses around the condition in line 356 to avoid any confusion, and also in case of changes later?

edwinalu (approved):
Thanks for adding the ingestion logic for semantic versioning!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: Thanks @edwinalu. 
_Would the semantic version be part of the value (oldValue and newValue)?_
oldValue/newValue are records that may or may not contain the field baseSemanticVersion. The aspectVersionComparator() accounts for both possibilities.

_The logic is correct, but could you please put parentheses around the condition in line 356 to avoid any confusion, and also in case of changes later?_
Good point, I've added the parantheses.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
+    DataMap newVerMap = newValue != null ? newValue.data().getDataMap("baseSemanticVersion") : null;
+    DataMap oldVerMap = oldValue != null ? oldValue.data().getDataMap("baseSemanticVersion") : null;
+
+    if (newVerMap == null && oldVerMap == null) {
+      return 0;
+    } else if (newVerMap == null && oldVerMap != null) {
+      return -1;
+    } else if (newVerMap != null && oldVerMap == null) {
+      return 1;
+    } else { //newVerMap != null && oldVerMap != null)
```
RealChrisL, line 1432: remove the right parenthesis in the comment?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
+      // Translate baseSemanticVersion into array [major, minor, patch]
+      int[] newVerArr = { newVerMap.getInteger("major").intValue(), newVerMap.getInteger("minor").intValue(),
```
RealChrisL, line 1434: Please put final as possible in variables.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: @RealChrisL thanks for the catch! Fixed.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: @RealChrisL thanks for the tip, I updated the code.

RealChrisL (approved):

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
     return mapToReturn;
   }
+
+  /** Aspect Version Comparator.
+   * @param newValue - Aspect that may have baseSemanticVersion field (by including BaseVersionedAspect).
+   * @param oldValue - Aspect that may have baseSemanticVersion field (by including BaseVersionedAspect).
+   * @return Return integer (-1, 0, 1) depending on if newValue version is (lesser than, equal to, greater than)
+   *     oldValue version.
+   */
+  protected int aspectVersionComparator(@Nonnull RecordTemplate newValue, @Nonnull RecordTemplate oldValue) {
```
yangyangv2, line 1420: the @Nonnull both `newValue` and `oldValue` is not guaranteed, right?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
+    // Attempt to extract baseSemanticVersion from incoming aspects
+    // If aspect or version does not exist, set version as lowest ranking (null)
+    final DataMap newVerMap = newValue != null ? newValue.data().getDataMap("baseSemanticVersion") : null;
```
yangyangv2, line 1423: you will need to handle the case either newValue or oldValue is null

yangyangv2, line 1423: Can we make "baseSemanticVersion" a constant?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
+    final DataMap oldVerMap = oldValue != null ? oldValue.data().getDataMap("baseSemanticVersion") : null;
+
```
yangyangv2, line 1425: please add comments to explain the scenarios

1. if neither new nor old has version, consider as equal
2. if only new has version, it will override the old value regardless equal or not
3. if old value has version, but new doesn't have version, it will ignore the new ingestion.
4. version comparison logic ..

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -274,6 +277,34 @@ public void testMAEEmissionOnValueChange() throws URISyntaxException {
     verifyNoMoreInteractions(_mockEventProducer);
   }
 
+  @Test(description = "Test MAE emission triggered by incoming aspects with higher versions")
```
yangyangv2, line 280: Thanks for writing these cases! 

Please also add two cases for either newValue and oldValue is null.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
yangyangv2: This would cause assertionFailure when `newValue != null` and `oldValue == null` 

 ---- For all fresh ingestions.

yangyangv2 (changes requested):
aspectVersionComparator() needs to handle case

newValue != null and oldValue == null or
newValue == null and oldValue =! null

The current implementation will break all the fresh ingestions

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: Thanks @yangyangv2. I have removed the @nonNull check. The first few lines in aspectVersionComparator handles null values.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: Thanks @yangyangv2 . After I removed `@Nonnull`, the code does handle null inputs. To elaborate,

- If any of the values are null, the ternary logic will keep them as null
- Logic in the rest of the comparator treats null as lowest ranking. This is one standard way of setting up a comparator. 
- With this logic, we will get the following behavior that matches our design:

| newValue | oldValue | newVerMap | oldVerMap | Comparator output | Behavior |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Null | Not Null | Null | Not Null | -1 | Skip write |
| Not Null  | Null  | Not Null | Null | 1 | Do not skip write  |

I made this clearer by adding comments and a aspectVersionSkipWrite() function.

schuangv2: @yangyangv2  thanks, I have made the change.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: thanks @yangyangv2 . I have added more comments to explain the scenarios.

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
schuangv2: Thanks @yangyangv2 . I can't add null cases here because _dummyLocalDAO.add() does not accept null inputs. I have added null tests to testAspectVersionSkipWrite().

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1410,45 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
+    if (newVerMap == null && oldVerMap == null) {
+      return 0;
+    } else if (newVerMap == null && oldVerMap != null) {
+      return -1;
+    } else if (newVerMap != null && oldVerMap == null) {
+      return 1;
+    } else { //newVerMap != null && oldVerMap != null
+      // Translate baseSemanticVersion into array [major, minor, patch]
+      final int[] newVerArr = { newVerMap.getInteger("major").intValue(), newVerMap.getInteger("minor").intValue(),
```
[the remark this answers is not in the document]
schuangv2: @yangyangv2 thanks, I have made the changes.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: @yangyangv2 Also added function aspectVersionSkipWrite() and tests to explicity call out the behavior and make it easier to understand.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -449,9 +458,11 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
       }
     }
 
-    // Skip saving if there's no actual change
-    if ((oldValue == null && newValue == null) || oldValue != null && newValue != null
-        && equalityTester.equals(oldValue, newValue)) {
+    // Skip saving for the following scenarios
+    if ((oldValue == null && newValue == null) // values are null
+        || (aspectVersionSkipWrite(newValue, oldValue)) // newValue ver < oldValue ver
```
yangyangv2, line 463: Thanks for addressing bug. A second thought on this behavior:

if the incoming version is greater, should we run the update regardless the equality check? or we only accept ingestion when the version is higher?

Given: new (1.0, status = true), old (0.9, status = true)

should the ending state be:

A: new (1.0, status = true)
B: old (0.9, status = true)

@edwinalu , @edwinalu , @jsdonn

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1407,4 +1418,64 @@ protected Map<String, Set<String>> transformBackfillResultsToStringMap(
...
     return mapToReturn;
   }
+
+  /** Aspect Version Comparator.
+   * @param newValue - Aspect that may have baseSemanticVersion field (by including BaseVersionedAspect).
+   * @param oldValue - Aspect that may have baseSemanticVersion field (by including BaseVersionedAspect).
+   * @return Return integer (-1, 0, 1) depending on if newValue version is (lesser than, equal to, greater than)
+   *     oldValue version.
+   */
+  protected int aspectVersionComparator(RecordTemplate newValue, RecordTemplate oldValue) {
```
yangyangv2, line 1428: nit: put a @Nullable to both param

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
edwinalu: Yes, the higher version should get persisted, and in @yangyangv2 's example, the end state should be A: new (1.0, status = true)

yangyangv2: this logic is essentially saying, 

if the new version is smaller, the ingestion will be dropped
if the new version is larger or equal, it will do the equality check

or should this be

if the new version is larger, ignore the equal check and do the update
if the new version is equal, then do the equality check.

schuangv2: @yangyangv2 @edwinalu thanks for the comments. 
@edwinalu I just added a test and confirmed that the higher version will be persisted, even if values are equal. The reason this happens is that the version is included in the equality check. Eg if we bump the version without changing anything else, equality check will return false and new version will be written.

schuangv2: @yangyangv2 I think we are almost on the same page. The logic is 

- if the new version is smaller, the ingestion will be dropped
- if the new version is equal, it will do the equality check
- if new version is larger, it will fail equality check. Ingestion will occur.

edwinalu: Thanks for confirming and adding the test case!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
schuangv2: @yangyangv2 thanks, I've added the change.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
yangyangv2: With this logic the outcome will be B: old (0.9, status = true). 

because  1.0 > 0.9, so `(aspectVersionSkipWrite(newValue, oldValue))` will be `false` and the equity check will be invoked. Since both have `status = true` then the new ingestion (1.0, status = true) will be dropped and 

B: old (0.9, status = true) 

will be the outcome.

yangyangv2: Nvm, synced offline and the equalityTester will capture the version difference. 

I'll let this PR pass due to its urgency. But MG team should review the equalityTest whether to exclude auxiliary metadata in the future we will have other control metadata such as writeMode. We shouldn't let equalityTest be polluted with these control attributes.

@jsdonn @zhixuanjia

yangyangv2 (approved):
Thanks for addressing all the comments!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: I wonder if we could have the use case-specific comparison logic on the producer side. In this example, if SI can detect the current version we have stored in db (through get API call), and they can add the OVERRIDE ingestion mode if the comparison returns true.

PR #354, opened 2024-02-07
feat: enable listUrns paginated by last urn in old schema mode

JiaoMaWHU:
## Summary
- current `listUrns` methods either do not support old schema or only support paginated by offset.
- this PR implemented a method for pagination by last urn in old schema, and added in the `listUrn` with index filters method.
- Note that it can only be triggered if the provided filters are null.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -36,6 +36,7 @@ import com.linkedin.metadata.query.IndexValue;
 import com.linkedin.metadata.query.ListResultMetadata;
 import io.ebean.DuplicateKeyException;
 import io.ebean.EbeanServer;
+import io.ebean.ExpressionList;
 import io.ebean.PagedList;
 import io.ebean.Query;
 import io.ebean.SqlUpdate;
@@ -1099,6 +1100,39 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     return toListResult(urns, null, pagedList, start);
   }
 
+  /*
+    * List all URNs of the entity registered in the DAO, paginated by last urn. Expect duplicate urns in the result.
+    *
+    * This method is old schema mode use only. Seek alternatives in other schemas.
+   */
+  @Deprecated
+  protected List<URN> listUrnsPaginatedByLastUrn(@Nullable URN lastUrn, int pageSize) {
+    if (_schemaConfig != SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("this method is only allowed in OLD_SCHEMA_ONLY mode");
+    }
+
+    final String entityType = ModelUtils.getEntityTypeFromUrnClass(_urnClass);
+
+    ExpressionList<EbeanMetadataAspect> query = _server.find(EbeanMetadataAspect.class)
+        .select(KEY_ID)
+        .where()
+        .startsWith(URN_COLUMN, "urn:li:" + entityType + ":")
+        .eq(VERSION_COLUMN, LATEST_VERSION)
+        .ne(METADATA_COLUMN, DELETED_VALUE);
+
+    if (lastUrn != null) {
+      query = query.gt(URN_COLUMN, String.valueOf(lastUrn));
+    }
+
+    final PagedList<EbeanMetadataAspect> pagedList = query
+        .setMaxRows(pageSize)
+        .orderBy()
+        .asc(URN_COLUMN)
+        .findPagedList();
+
+    return pagedList.getList().stream().map(entry -> getUrn(entry.getKey().getUrn())).collect(Collectors.toList());
+  }
+
   @Nonnull
   <ASPECT extends RecordTemplate> ListResult<ASPECT> getListResult(@Nonnull Class<ASPECT> aspectClass,
       @Nonnull PagedList<EbeanMetadataAspect> pagedList, int start) {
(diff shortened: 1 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1401,6 +1435,9 @@ static <ASPECT extends RecordTemplate> String getFieldColumn(@Nonnull String pat
    *
    * <p>NOTE: Currently this works for upto 10 filter conditions.
    *
+   * <p>This method can only be used in old schema mode if the provided indexFilter and indexSortCriterion are null. In old
+   * schema mode, expect duplicate urns in the result.
```
zhixuanjia, line 1439: Why we don't deduplicate the urns? Can't think of any use cases for duplicate urns.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1099,6 +1100,39 @@ public <ASPECT extends RecordTemplate> ListResult<URN> listUrns(@Nonnull Class<A
...
+    if (_schemaConfig != SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("this method is only allowed in OLD_SCHEMA_ONLY mode");
+    }
+
+    final String entityType = ModelUtils.getEntityTypeFromUrnClass(_urnClass);
+
+    ExpressionList<EbeanMetadataAspect> query = _server.find(EbeanMetadataAspect.class)
+        .select(KEY_ID)
+        .where()
+        .startsWith(URN_COLUMN, "urn:li:" + entityType + ":")
```
zhixuanjia, line 1119: what does line do? `.startsWith(URN_COLUMN, "urn:li:" + entityType + ":")`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
JiaoMaWHU: here are what I have tried to make it return distinct urns:

1. set distinct
```
.select(KEY_ID)
.setDistinct(true)
.where()
```
however it didn't work, the returned keys are null due to unknown reasons

2. add distinct in the return list.
this worked, but it could make the size of the list less than the pageSize, making the downstream hard to determine when should a pagination stop

do you have any suggestions?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
JiaoMaWHU: it filters for the keys of a specific entity

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
zhixuanjia: How about manually assemble the SQL.  something like `_server.createSqlQuery("SELECT distinct(urn) ...").findList()`

JiaoMaWHU: sure, let me try that

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -1469,6 +1470,49 @@ public void testListUrns() {
     assertEquals(results.getValues().get(2), makeFooUrn(2));
   }
 
+  @Test
+  public void testListUrnsPaginatedByLastUrn() {
```
zhixuanjia, line 1474: Can we has a test case that covers duplicate urns case?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1099,6 +1100,39 @@ public <ASPECT extends RecordTemplate> ListResult<URN> listUrns(@Nonnull Class<A
...
+    if (_schemaConfig != SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("this method is only allowed in OLD_SCHEMA_ONLY mode");
+    }
+
+    final String query = getDistinctUrnsOfEntitySqlQuery(lastUrn, pageSize);
+    final List<SqlRow> sqlRows = _server.createSqlQuery(query).setFirstRow(0).findList();
+    return sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString(URN_COLUMN))).collect(Collectors.toList());
+  }
+
+  private String getDistinctUrnsOfEntitySqlQuery(URN lastUrn, int pageSize) {
```
zhixuanjia, line 1119: move this to SQLStatementUtils?

zhixuanjia (approved):
LGTM!

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
JiaoMaWHU: yes it is covered in https://github.com/linkedin/datahub-gma/pull/354/files#diff-2f5183652d8c3cb8bedbfeca755324b04385b3ac03050c28880d6341574575adR1506

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
JiaoMaWHU: would incline to keep it here so that such a legacy query don't confuse user, or be accidentally reused

PR #356, opened 2024-02-13
fix: toLowerCase when checking column existance

zhixuanjia:
MySQL columns are case insensitive. So `abc` and `aBC` are essentially same columns.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -547,12 +547,12 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
       final List<SqlRow> rows = _server.createSqlQuery(SQLStatementUtils.getAllColumnForTable(tableName)).findList();
       Set<String> columns = new HashSet<>();
       for (SqlRow row : rows) {
-        columns.add(row.getString("COLUMN_NAME"));
+        columns.add(row.getString("COLUMN_NAME").toLowerCase());
       }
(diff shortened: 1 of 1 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -547,12 +547,12 @@ public boolean checkColumnExists(@Nonnull String tableName, @Nonnull String colu
       final List<SqlRow> rows = _server.createSqlQuery(SQLStatementUtils.getAllColumnForTable(tableName)).findList();
       Set<String> columns = new HashSet<>();
       for (SqlRow row : rows) {
-        columns.add(row.getString("COLUMN_NAME"));
+        columns.add(row.getString("COLUMN_NAME").toLowerCase());
```
JiaoMaWHU, line 550: this change looks good to me. But want to confirm the rc, is the understanding correct that the column in aim db is lowercased, but the restli request contains camel case column?

zhixuanjia: so the column in AIM db is `i_publishrequest$modeldeploymentgroup` but in the request, we pass in `i_publishrequest$modelDeploymentGroup` The uppercase D and G makes it mismatch.

However, if you SQL select instead. 
```
SELECT ... WHERE COLUMN_NAME = 'i_publishrequest$modeldeploymentgroup';

SELECT ... WHERE COLUMN_NAME = 'i_publishrequest$modelDeploymentGroup';
``` 
Both SQL statements work. That's why previous impl works.

JiaoMaWHU (approved):
thanks for fixing this issue!

PR #349, opened 2024-02-02
feat: Create BaseIngestionMode.pdl

schuangv2:
## Summary
Adding BaseIngesitonMode.pdl. This record has a writeMode attribute will declare how the aspect metadata change events (MCEs) should be written to the backend.

## Details
For starters, we will allow MCEs to declare the following write modes:
- DEFAULT: Default write mode. Same behavior as not setting writeMode.
- OVERRIDE: Override checks that may skip metadata write (eg equality check).

## Testing
- **Build test**: `./gradlew build` ran successfully

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Tests for the changes have been added/updated (if applicable)
- [x] Docs related to the changes have been added/updated (if applicable)

PR #357, opened 2024-02-13
feat(ingestion): add IngestionMode to MCE, allow DAO to process LIVE_OVERRIDE events

jsdonn:
Redesigns the solution that PR #349 (which added the ingestion modes to aspects) was meant to address. Instead of having ingestion mode controls on the aspect-level (i.e. allowing aspects to extend a BaseIngestionMode model), have ingestion mode controls on the MCE level. Also is cleaner since we can reuse the existing IngestionMode introduced as part of the MAE/MCE backfill work. 

Behavior:
- If IngestionMode = LIVE (or if not specified in the MCE), nothing changes. Same behavior as before where no-change operations are skipped.
- If IngestionMode = LIVE_OVERRIDE, always update the database with the incoming changes.
  - If incoming metadata = current metadata, still do the upsert operation (resulting in just a timestamp update). DON'T EMIT MAE in this case.
  - If incoming metadata != current metadata, proceed as normal i.e. upsert into db and emit MAE.

It's worth noting that the LIVE_OVERRIDE write mode will always take precedence over the semantic version comparison logic added in PR #352 . This means that even if the version check fails (newVer < oldVer), we do not skip the ingestion event.

In the future, we should have 3 ingestion mode controls:
1) On the MCE-level
2) In an aspect's annotations (e.g. for garbage collector use case)
3) In the GMS-specific configs

Order of precedence will be 3 > 2 > 1. For now, this PR introduces 1).

Next steps are:
1) Bump metadata-models version with latest datahub-gma
2) Bump wherehows-samza (mce-v5-consumer) with latest metadata-models
3) Update all GMS which use ingestWithTracking with the latest API changes 
  - metric-gms, metadata-store, corp-identity-gms, job-gms, service-gms
4) Update mce-v5-consumer to check if MCE has an IngestionMode attached. If so, call ingestWithTracking with the IngestionMode param.

## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -458,21 +478,30 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
+
     // Save the newValue as the latest version
     long largestVersion = saveLatest(urn, aspectClass, oldValue, oldAuditStamp, newValue, auditStamp, latest.isSoftDeleted, trackingContext);
 
     // Apply retention policy
     applyRetention(urn, aspectClass, getRetention(aspectClass), largestVersion);
 
-    return new AddResult<>(oldValue, newValue, aspectClass);
+    // don't emit MAE if we are force updating metadata whose value does not change (i.e. timestamp-only update)
+    return new AddResult<>(oldValue, newValue, aspectClass, ingestionMode == IngestionMode.LIVE_OVERRIDE && oldAndNewEqual);
```
zhixuanjia, line 504: why we want to control the MAE behavior here? 

MAE should be emitted should be purely based on EqaulityTester.equals(oldVal, newVal).

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -661,6 +697,16 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, @Nonnull Cla
     return add(urn, new AspectUpdateLambda<>(aspectClass, updateLambda), auditStamp, maxTransactionRetry, trackingContext);
   }
 
+  /**
+   * Same as above {@link #add(Urn, Class, Function, AuditStamp, int)} but with a write mode parameter.
+   */
```
zhixuanjia, line 702: remove `above` , method orders might change.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -747,26 +793,29 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, @Nonnull Cla
...
   }
 
   /**
    * Same as above {@link #add(Urn, RecordTemplate, AuditStamp)} but with tracking context.
    */
   @Nonnull
   public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, @Nonnull ASPECT newValue,
-      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext) {
-    return add(urn, (Class<ASPECT>) newValue.getClass(), ignored -> newValue, auditStamp, trackingContext);
+      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext, @Nullable IngestionMode ingestionMode) {
```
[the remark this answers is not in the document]
jsdonn: I checked that this add() method is only used in a few places:
image
I changed all 4 non-test usages. Is this still a concern? There should not be any external users afaik since this method includes IngestionTrackingContext as a param (and I don't think any customers have adopted this)

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionMode.pdl
```diff
@@ -10,6 +10,13 @@ enum IngestionMode {
    */
   LIVE
 
+  /**
+   * Live ingestion with override toggled on. Intended to be used with Metadata Change Events (MCEs).
+   * The metadata will be ingested irrespective of whether or not the old and new values are the same.
```
zhixuanjia, line 15: not just equality checks right? it will also skip semantic versioning check.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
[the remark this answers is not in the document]
jsdonn: I also want to avoid keep adding add() methods. there are already so many in this class.

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionMode.pdl
```diff
@@ -10,6 +10,13 @@ enum IngestionMode {
...
+   * A Metadata Audit Event (MAE) will be emitted only if the old and new values are different.
```
zhixuanjia, line 16: no need to mention MAE, how to emit MAE can be controlled separately. e.g. there is an existing flag that can `alwaysEmitMAE`

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionMode.pdl
```diff
@@ -10,6 +10,13 @@ enum IngestionMode {
...
+   */
+  LIVE_OVERRIDE
```
zhixuanjia, line 18: this will be converted to Avro schema right? In avro schema, new enum has to be added at the last element.

gradle-plugins/metadata-events-generator-lib/src/main/resources/AspectUnionEvent.rythm
```diff
@@ -32,6 +33,11 @@ record MCE@(eventSpec.getShortTyperefName()) {
    */
   ingestionTrackingContext: optional union[null, IngestionTrackingContext] = null
 
+  /**
+   * Type of the ingestion. Allow null for backward compatibility. Null will be treated as live ingestion.
+   */
+  ingestionMode: optional union[null, IngestionMode] = null
```
zhixuanjia, line 39: ingestion mode for union event should be added into each aspect, not at the top level.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jsdonn: Oh you're right. I didn't realize there is a check in unwrappAddResult for oldValue != null. I will remove.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -130,9 +139,19 @@ public static class AspectUpdateLambda<ASPECT extends RecordTemplate> {
...
+    @NonNull
+    IngestionMode ingestionMode;
+
     AspectUpdateLambda(ASPECT value) {
       this.aspectClass = (Class<ASPECT>) value.getClass();
       this.updateLambda = (ignored) -> value;
+      this.ingestionMode = IngestionMode.LIVE;
+    }
+
+    AspectUpdateLambda(@NonNull Class<ASPECT> aspectClass, @NonNull Function<Optional<ASPECT>, ASPECT> updateLambda) {
```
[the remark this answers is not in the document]
jsdonn: used here - https://jarvis.corp.linkedin.com/codesearch/result/?name=BaseLocalDAO.java&path=datahub-gma%2Fdao-api%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fmetadata%2Fdao&reponame=linkedin%2Fdatahub-gma#647

only 3 args

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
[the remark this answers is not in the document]
yangyangv2: what if you have another ingestion param to add in the future? 

If add() method is only used internally, we want to extend this time, how about 

"for future extensibility consideration, better to have a Wrapper class such as `IngestionContext`, or reuse the existing IngestionTrackingContext,"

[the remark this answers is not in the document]
jsdonn: I added an IngestionParams which includes IngestionMode instead.

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionParams.pdl
```diff
@@ -0,0 +1,11 @@
+namespace com.linkedin.metadata.events
```
zhixuanjia, line 1: suggest to move it to different packages, something like `com.linkedin.metadata.internals`. It's not used in `events`, only used in DAO layer.

Or maybe instead of using PDL, just create a plain java class?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -402,7 +440,8 @@ public void testAlwaysFalseEqualityTester() {
     }
 
     verify(_mockProducer, times(1)).produceMetadataAuditEvent(urn, null, foo1);
-    verify(_mockProducer, times(1)).produceMetadataAuditEvent(urn, foo1, foo2);
+    // TODO - fix this test, MAE is not emitted because foo1 == foo2.
```
zhixuanjia, line 443: why AlwaysFalseEqualityTester will have foo1 == foo2. Shouldn't it always return foo1 != foo2 instead?

jsdonn: fixed this. it's because before I just used oldValue.equals(newValue). now I am using the equality tester

dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionParams.pdl
jsdonn: Created com.linkedin.metadata.internal package.

zhixuanjia (approved):
thanks for addressing the comments.

PR #382, opened 2024-05-29
Introduce BaseAspectV3Resource

zhixuanjia:
## Context
Aspect-routing is an essential feature provided by Metadata Graph. As part of the simplify metadata onboarding initiative, we will need to make this feature easy to enable. Current process (_go/mg/aspectrouting, a 7-page doc on how to enable aspect-routing_) for enabling aspect-routing has many moving pieces and requires code changes as different places.

## Solution
Inspired by existing `BaseEntityAgnosticResource`, we introduce `BaseAspectV3Resource` in this PR. The new BaseAspectV3Resource has mostly same set of methods as its previous version BaseAspectV2Resource. The major difference is that BaseAspectV3Resource is **entity agnostic**. User only needs to specify the routing aspect when implementing it. 

Using `AccoutableOwnership` aspect as an example, the class signature will be

from 
`public class GrpcMethodAccountableOwnershipResource extends BaseAspectV2Resource<GrpcMethodUrn, GrpcMethodAspect, AccountableOwnership>`
to 
`public class AccountableOwnershipResource extends BaseAspectV3Resource<AccountableOwnership>`

## Why this solution
The major benefit of this solution is when a new entity is onboarded, user will _not_ need to go to aspect GMS e.g. ownership-gms to implement new clients and new resources etc. The AccountableOwnershipResource hosted in ownership-gms should be able read / write AccountableOwnership not no matter which entity the aspect is attached to.

## Testing Done
Unit test

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
@@ -2,6 +2,7 @@ package com.linkedin.metadata.dao;
 
 import com.google.common.io.Resources;
 import com.linkedin.common.AuditStamp;
+import com.linkedin.common.urn.Urn;
 import com.linkedin.metadata.dao.localrelationship.SampleLocalRelationshipRegistryImpl;
 import com.linkedin.metadata.dao.urnpath.EmptyPathExtractor;
 import com.linkedin.metadata.dao.utils.BarUrnPathExtractor;
@@ -83,7 +84,7 @@ public class EbeanLocalAccessTest {
     _ebeanLocalAccessBar = new EbeanLocalAccess<>(_server, EmbeddedMariaInstance.SERVER_CONFIG_MAP.get(_server.getName()),
         BarUrn.class, new BarUrnPathExtractor(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
     _ebeanLocalAccessBurger = new EbeanLocalAccess<>(_server, EmbeddedMariaInstance.SERVER_CONFIG_MAP.get(_server.getName()),
-        BurgerUrn.class, new EmptyPathExtractor<>(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
+        BurgerUrn.class, new EmptyPathExtractor(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
     _ebeanLocalAccessFoo.setLocalRelationshipBuilderRegistry(new SampleLocalRelationshipRegistryImpl());
     _now = System.currentTimeMillis();
   }
@@ -115,7 +116,7 @@ public class EbeanLocalAccessTest {
     // Given: metadata_entity_foo table with fooUrns from 0 ~ 99
 
     FooUrn fooUrn = makeFooUrn(0);
-    AspectKey<FooUrn, AspectFoo> aspectKey = new AspectKey(AspectFoo.class, fooUrn, 0L);
+    AspectKey<AspectFoo> aspectKey = new AspectKey(AspectFoo.class, fooUrn, 0L);
 
     // When get AspectFoo from urn:li:foo:0
     List<EbeanMetadataAspect> ebeanMetadataAspectList =
@@ -135,7 +136,7 @@ public class EbeanLocalAccessTest {
 
     // When get AspectFoo from urn:li:foo:9999 (does not exist)
     FooUrn nonExistFooUrn = makeFooUrn(9999);
-    AspectKey<FooUrn, AspectFoo> nonExistKey = new AspectKey(AspectFoo.class, nonExistFooUrn, 0L);
+    AspectKey<AspectFoo> nonExistKey = new AspectKey(AspectFoo.class, nonExistFooUrn, 0L);
     ebeanMetadataAspectList = _ebeanLocalAccessFoo.batchGetUnion(Collections.singletonList(nonExistKey), 1000, 0, false);
 
     // Expect: get AspectFoo from urn:li:foo:9999 returns empty result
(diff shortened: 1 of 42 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectV3Resource.java
jphui, line 1: I hate to be that person but I think one of my `nits` would be getting rid of the `V3` and naming the resource `BaseEntityAgnosticAspectResource` so that immediately it's obvious what it is.

Even with V2 I wouldn't know off the top of my head how it's different.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
jphui, line 1: I actually didn't even know this existed... do you know if this resource would have been another approach to SMO: "no entity resource classes, etc."?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
jphui, line 1: Any TLDR on why BaseEntityResource needs updating too? I thought the change was restricted to just an Aspect level (sub)resource.

jphui (commented):
Thanks for doing this, I don't recall this being a part of any OKR so this is a big Craftsmanship win 🙌 

A few things:

1. Is there any doc or ticket this can be linked to? If my previous statement is correct perhaps this was just pickup work too.
2. What is the relationship between `BaseEntityAgnosticResource` and the one introduced here? Are they orthogonal ideas or should they be used together (?)

Small nit, in PR description, if you can put some more context to:
> the class signature will be like public AccountableOwnershipResource extends BaseAspectV3Resource<AccountableOwnership>

aka:
"""
from
```
public AccountableOwnershipResource extends BaseAspectV2Resource<Dataset, AccountableOwnership>
```
to
```
public AccountableOwnershipResource extends BaseAspectV3Resource<AccountableOwnership>
```
"""

just to help with clarity and the value-add here.

3. I'm now thinking about the implications on "`Aspect Resource Reuse`"... would it be correct to say that:
```
previously, users needed to create an "Entity-specific, Aspect-specific" resource when routing.

now, a single Routed aspect resource can be reused by multiple entities in the same GMS?
```

(4) Now also wondering the implications on reusing Aspect Resources entirely. As of now I think things like `Status.java` and stuff needs to exist 1 per entity even outside the Routing domain...

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -83,7 +84,7 @@ public void init() {
     _ebeanLocalAccessBar = new EbeanLocalAccess<>(_server, EmbeddedMariaInstance.SERVER_CONFIG_MAP.get(_server.getName()),
         BarUrn.class, new BarUrnPathExtractor(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
     _ebeanLocalAccessBurger = new EbeanLocalAccess<>(_server, EmbeddedMariaInstance.SERVER_CONFIG_MAP.get(_server.getName()),
-        BurgerUrn.class, new EmptyPathExtractor<>(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
+        BurgerUrn.class, new EmptyPathExtractor(), _ebeanConfig.isNonDollarVirtualColumnsEnabled());
```
jsdonn, line 87: small note - this will break all downstreams that need to do a metadata-models version bump. like AIM for example - https://jarvis.corp.linkedin.com/codesearch/result/?name=BaseLocalDaoFactory.java&path=ai-metadata-gms%2Fai-metadata-factories%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fproml%2Fmetadata%2Ffactory&reponame=linkedin-multiproduct%2Fai-metadata-gms#72

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectV3Resource.java
```diff
@@ -0,0 +1,162 @@
...
+   */
+  @Nonnull
+  protected abstract LocalDaoRegistry getLocalDaoRegistry();
+
+  /**
+   * Returns a {@link BaseRestliAuditor} for this resource.
+   */
+  @Nonnull
+  protected BaseRestliAuditor getAuditor() {
+    return DUMMY_AUDITOR;
```
jsdonn, line 61: I know you carried this over from v2 but are we supposed to update this at some point to not be a dummy value?

restli-resources/src/main/java/com/linkedin/metadata/restli/dao/DefaultLocalDaoRegistryImpl.java
```diff
@@ -57,13 +57,10 @@ public static DefaultLocalDaoRegistryImpl init(
   }
```
jsdonn, line 57: regarding the init method above ^ where and how is DefaultLocalDaoRegistryImpl::init called? 

Can we keep it simpler for the user and just accept a list of entity types? and then we can construct the DAOs for them so they don't have to construct 1 DAO per entity.

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
jsdonn: What is the point of this resource? Doesn't it serve a similar purpose as BaseAspectV3Resource? Both are entity agnostic resources. Is there a way to reuse the existing one?

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectV3Resource.java
zhixuanjia: Great idea!

restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityAgnosticResource.java
zhixuanjia: to @jphui 
Yep, this is the exact “Generic API” approach we discussed during SMO design phase. 

to @jsdonn 
So short answer, no they don't serve a similar purpose even they are all generic API. The existing `BaseEntityAgnosticResource` is created as part of platform consistency project and its APIs are for entity level operations and should be used in "main" GMS such as TMS or service-gms.

The `BaseAspectV3Resource ` in this PR is to create only one resource for a single aspect and its APIs are for aspect level operations and should be used in "routed aspect" GMS such as ownership-gms.

PR #383, opened 2024-06-05
[SMO] Generic DAO and its serving API

zhixuanjia:
## Summary
First PR of a series PRs to introduce Generic DAO and its corresponding service APIs.

There has been previous attempt to implement a generic interface https://github.com/linkedin/datahub-gma/pull/325. But this attempt only covers the rest APIs. Under the hood, it still requires user to create many type-bounded DAOs, which means we still needs code changes for new entity e.g. write DAO factory, register the DAO etc...

This PR attempts to have a grand rewrite starting from the very bottom layer, which remove all the type bounds. All the type-casting is left to client to perform. Metadata is ingested as "raw" and served as "raw" in terms of JSON serialized strings.

## Testing Done
Unit tests

## Checklist


JiaoMaWHU (commented):
what is the "client" here that is supposed to cast string <-> strong typed metadata?

after this pr, will the metadata producers still be required to create strong typed models for MCEs?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Save metadata into database.
+   */
+  private void saveLatest(@Nonnull String urn, @Nonnull Class aspectClass, @Nonnull RecordTemplate newValue,
+      @Nullable RecordTemplate currentValue, @Nonnull AuditStamp newAuditStamp, @Nullable AuditStamp currentAuditStamp) {
+    validateUrn(urn);
+
+    // Save oldValue as the largest version + 1
+    long largestVersion = getNextVersion(urn, aspectClass);
+    System.out.println("largest versio:" + largestVersion);
```
yangyangv2, line 115: Make it a debug log? and also include the urn/aspect information.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
```diff
@@ -0,0 +1,242 @@
...
+    if (metadata == null || metadata.getMetadata() == null) {
+      return Optional.empty();
+    }
+
+    final ExtraInfo extraInfo = toExtraInfo(metadata);
+    return Optional.of(new GenericLocalDAO.MetadataWithExtraInfo(metadata.getMetadata(), extraInfo));
+  }
+
+  /**
+   * Save metadata into database.
```
yangyangv2, line 107: can we add a high-level description about the persistence logic?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
```diff
@@ -0,0 +1,242 @@
...
+
+      // update latest version
+      updateWithOptimisticLocking(urn, newValue, aspectClass, newAuditStamp, 0, new Timestamp(currentAuditStamp.getTime()));
+    } else {
+      // When for fresh ingestion or with changeLog disabled
+      insert(urn, newValue, aspectClass, newAuditStamp, 0);
+    }
+  }
+
+  private long getNextVersion(@Nonnull String urn, @Nonnull Class aspectClass) {
```
yangyangv2, line 128: do we still need this in new schema?

yangyangv2 (commented):
Can you elaborate what is the ending goal for this refactor? considering

1. MMG will still need strong typed APIs, however
2. Graph query, aspect query does require entity agnostic access

Is the goal to unify the bottle logic for the two cases above?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  /**
+   * Save the metadata into database.
+   * @param urn The identifier of the entity which the metadata is associated with.
+   * @param aspectClass The aspect class for the metadata.
+   * @param metadata The metadata serialized as JSON string.
+   * @param auditStamp audit stamp containing information on who and when the metadata is saved.
+   */
+  public void save(@Nonnull String urn, @Nonnull Class aspectClass, @Nonnull String metadata, @Nonnull AuditStamp auditStamp) {
```
yangyangv2, line 67: let's keep the Urn type for better type constrain?? what's the consideration of using a string?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
zhixuanjia: This EbeanGenericLocalDAO is specifically for old schema format. For new schema, the impl will be `EbeanGenericLocalAccess.java` which I will implement in later PR.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
zhixuanjia: sounds good. Added some descriptions.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
zhixuanjia: sure, updated.

zhixuanjia:
> Can you elaborate what is the ending goal for this refactor? considering
> 
> 1. MMG will still need strong typed APIs, however
> 2. Graph query, aspect query does require entity agnostic access
> 
> Is the goal to unify the bottle logic for the two cases above?

Hi @yangyangv2 ,

The immediate issue that this PR want to solve is aspect routing. Current aspect routing hooks too many dependencies on the aspect GMS e.g. ownership-gms. Just a few examples:
1. Even with code-gen, we still have to make code changes in other teams' MP. This makes the boundary unclear.
2. We have to make sure the aspect GMS is deployed first with new APIs. Otherwise, main GMS will throw errors.
3. We have to import new clients in the main GMS whenever there is a new entity.

With the generic DAO / APIs in this PR, the aspect GMS e.g. ownership-gms can host a set of generic APIs and _require no code change for new entity / aspect_. Regarding the type casting, this can happen in the BaseAspectRoutingResource class. 

As for future use cases, I think the type-casting can be a dedicated layer in the mid-tier of MMG. When new entity onboarded, an entry can be added in this layer. But no other places require code change since they are all generic.  A entity-specific resource / DAO class can still be created in case users needs customized logic.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanGenericLocalDAO.java
zhixuanjia: Using string is more generic and also I think generic Urn will go away in Proto? I don't have strong opinion on this. We can use Urn type and switch back to String should be easy.

yangyangv2: Let's stay with string for now. The statement of deprecating Urn in Proto concerns me. And we can come back to different type laster, it's an API mostly used internally.

yangyangv2:
> > Can you elaborate what is the ending goal for this refactor? considering
> > 
> > 1. MMG will still need strong typed APIs, however
> > 2. Graph query, aspect query does require entity agnostic access
> > 
> > Is the goal to unify the bottle logic for the two cases above?
> 
> Hi @yangyangv2 ,
> 
> The immediate issue this PR want to solve is aspect routing. Current aspect routing hooks too many dependencies on the aspect GMS e.g. ownership-gms. Just a few examples:
> 
> 1. Even with code-gen, we still have to make code changes in other teams' MP. This makes the boundary unclear.
> 2. We have to make sure the aspect GMS is deployed first with new APIs. Otherwise, main GMS will throw errors.
> 3. We have to import new clients in the main GMS whenever there is a new entity.
> 
> With the generic DAO / APIs in this PR, the aspect GMS e.g. ownership-gms can host a set of generic APIs and _require no code change for new entity / aspect_. Regarding the type casting, this can happen in the BaseAspectRoutingResource class.
> 
> As for future use cases, I think the type-casting can be a dedicated layer in the mid-tier of MMG. When new entity onboarded, an entry can be added in this layer. But no other places require code change since they are all generic. A entity-specific resource / DAO class can still be created in case users needs customized logic.

Synced offline. Generally aligned the GenericDAO will help with the status quo. Once moving to MMG, the future trend is still strongly-typed. GenericDAO may still have its place in the low-level logic. Just for wide-use and or extension of GenericDAO should  be well considered to avoid with the Code-Gens in MMG.

yangyangv2 (approved):
LGTM, thanks for the discussion and addressing the comments.

PR #389, opened 2024-07-12
feat: add new findRelationships interface and unit test placeholder

ybz1013:
## Summary
new findRelationships API, which doesn't require entity snapshot, is needed to support non-MG entities. This PR adds API signature and unit test placeholder for review purpose.

The old API only uses snapshot to extract the entity type and find the db table. We can just provide that entity urn in the function call.

## Context
The current findRelationships API in EbeanLocalRelationshipQueryDAO requires that the source, destination, and relationship type parameters are of type RecordTemplate (i.e. Snapshots in this case). Considering that we want to support relationships between both MG entities and non-MG entities (for the crew use case), we need to define a new findRelationships API in EbeanLocalRelationshipQueryDAO which does impose this restriction on the parameters.

## Testing Done
mint build

## JIRA
META-20529

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -165,23 +165,25 @@ public class EbeanLocalRelationshipQueryDAO {
   }
 
   /**
-   * Finds a list of relationships of a specific type based on the given filters.
-   * Source and destination entities must have tables registered under metadata_entity_sourceEntityType/destinationEntityType in db.
+   * Finds a list of relationships of a specific type based on the given filters if applicable.
    *
-   * @param sourceEntityType type name of the source entity to query
-   * @param sourceEntityFilter the filter to apply to the source entity when querying
-   * @param destinationEntityType type name of the destination entity to query
-   * @param destinationEntityFilter the filter to apply to the destination entity when querying
+   * @param sourceEntityUrn urn of the source entity to query
+   * @param sourceEntityFilter the filter to apply to the source entity when querying (not applicable to non-MG entities)
+   * @param destinationEntityUrn urn of the destination entity to query. If relationship is OwnedBy, this is crew/ldap.
+   * @param destinationEntityFilter the filter to apply to the destination entity when querying (not applicable to non-MG entities)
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -164,6 +164,27 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
         .collect(Collectors.toList());
   }
 
+  /**
+   * Finds a list of relationships of a specific type based on the given filters.
+   * Source and destination entities must have tables registered under metadata_entity_sourceEntityType/destinationEntityType in db.
```
jsdonn, line 169: So this line is not true anymore. For example, Crew is not an entity but we are allowing it to be either the source or destination of a relationship. So Crew will not have entity table associated with it (and thus we cannot do filtering on any Crew attributes).

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -164,6 +164,29 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+   * Finds a list of relationships of a specific type based on the given filters if applicable.
+   *
+   * @param sourceEntityUrn urn of the source entity to query
+   * @param sourceEntityFilter the filter to apply to the source entity when querying (not applicable to non-MG entities)
+   * @param destinationEntityUrn urn of the destination entity to query. If relationship is OwnedBy, this is crew/ldap.
+   * @param destinationEntityFilter the filter to apply to the destination entity when querying (not applicable to non-MG entities)
+   * @param relationshipType the type of relationship to query
+   * @param relationshipFilter the filter to apply to relationship when querying
+   * @param offset the offset query should start at. Ignored if set to a negative value.
+   * @param count the maximum number of entities to return. Ignored if set to a non-positive value.   * @return A list of relationship records.
```
jsdonn, line 177: we should probably set a default for count and then put it in the javadocs but you can do that when you implement this method

jsdonn (approved):
method signature LGTM

PR #390, opened 2024-07-16
feat: add new findRelationships API and unit tests

ybz1013:
## Summary
Implemented new findRelationships API that doesn't require snapshot.class as input AND supports non-MG entity (only crew for now).

Note that now entity as passed in as urn, and we will extract the entity type directly from it, which is not type-safe as snapshot. So extra validation is needed to check if an entity is actually MG. Two options I considered, one is to check if snapshot.class exists, and second is check the table name in db. This PR is following the latter.

Another change is nullability of entity and filter, the old API requires filters to be nonnull, but it still allows empty filter arrays. So I made it nullable, and also, entity and filter should both exists for the sql query to work.

## Details
change of old code:
1. rearranged param order for buildFindRelationshipSQL
2. modified buildFindRelationshipSQL input nullability and refactored logic for handling non-mg entity filter

new code:
1. implemented new findRelationships which takes entity urn instead of snapshot
2. new validation for entity urn and filter
4. logic to check if an entity has table in db by sql query
5. OwnedBy.pdl and table creation logic for local access for unit testing
6. unit tests that covers old MG entity only use case, and new crew entity use case

## Testing Done
./gradlew build
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -1,5 +1,6 @@
 package com.linkedin.metadata.dao;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.data.template.UnionTemplate;
 import com.linkedin.metadata.dao.utils.ClassUtils;
@@ -24,13 +25,20 @@ import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
 import org.javatuples.Pair;
+
+import static com.linkedin.metadata.dao.utils.SQLSchemaUtils.*;
+
 
 /**
  * An Ebean implementation of {@link BaseQueryDAO} backed by local relationship tables.
  */
 @Slf4j
 public class EbeanLocalRelationshipQueryDAO {
+  public static final String URN_CREW_PATTERN = "urn:li:crew:?[a-zA-Z0-9]*";
+  public static final String URN_ENTITY_PATTERN =
+      "urn:li:[a-zA-Z0-9]+:?\\(?[a-zA-Z0-9]*\\)?";
   private final EbeanServer _server;
   private final MultiHopsTraversalSqlGenerator _sqlGenerator;
 
@@ -151,13 +159,8 @@ public class EbeanLocalRelationshipQueryDAO {
 
     final String relationshipTableName = SQLSchemaUtils.getRelationshipTableName(relationshipType);
 
-    final String sql = buildFindRelationshipSQL(
-        destTableName,
-        sourceTableName,
-        relationshipTableName,
-        sourceEntityFilter,
-        destinationEntityFilter,
-        relationshipFilter);
+    final String sql = buildFindRelationshipSQL(relationshipTableName, relationshipFilter, sourceTableName,
+        sourceEntityFilter, destTableName, destinationEntityFilter);
 
     return _server.createSqlQuery(sql).findList().stream()
         .map(row -> RecordUtils.toRecordTemplate(relationshipType, row.getString("metadata")))
@@ -174,7 +177,8 @@ public class EbeanLocalRelationshipQueryDAO {
    * @param relationshipType the type of relationship to query
    * @param relationshipFilter the filter to apply to relationship when querying
    * @param offset the offset query should start at. Ignored if set to a negative value.
-   * @param count the maximum number of entities to return. Ignored if set to a non-positive value.   * @return A list of relationship records.
+   * @param count the maximum number of entities to return. Ignored if set to a non-positive value.
+   * @return A list of relationship records.
    */
   @Nonnull
   public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationships(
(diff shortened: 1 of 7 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -24,13 +25,20 @@
...
 
+import static com.linkedin.metadata.dao.utils.SQLSchemaUtils.*;
+
+
 /**
  * An Ebean implementation of {@link BaseQueryDAO} backed by local relationship tables.
  */
 @Slf4j
 public class EbeanLocalRelationshipQueryDAO {
+  public static final String URN_CREW_PATTERN = "urn:li:crew:?[a-zA-Z0-9]*";
```
jsdonn, line 39: We shouldn't have specific entities in here. Then if we want to onboard new entities, we have to modify this file every time which is not efficient. I would recommend a different way to detect entities vs non-entities.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -184,7 +188,135 @@ public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationship
...
+   * 2. if entity urn is null or empty, then filter should be emtpy.
+   * If any of above is violated, throw IllegalArgumentException.
+   * Categorize:
+   * 1. If entityUrn is null or empty, return NULL.
+   * 2. If entityUrn is a crew entity urn, return CREW.
+   * 3. If entityUrn is an MG entity urn, return MG_ENTITY.
+   *
+   * @return EntityType of the entityUrn
+   */
+  private EntityType validateAndCategorizeEntityUrn(@Nullable LocalRelationshipFilter filter, @Nullable String entityUrn) {
```
jsdonn, line 251: Can we do this differently? 

You can query for existing tables in the db and extract the registered entities from the table names. I'd recommend we run this query once on service start up and then cache the results in a map or list to reuse later.

https://stackoverflow.com/questions/3913620/get-all-table-names-of-a-particular-database-by-sql-query

jsdonn: i.e. 

1) run `SELECT TABLE_NAME  FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE' AND TABLE_SCHEMA=DATABASE();`.

2) filter for all tables that start with `metadata_entity_*`

3) extract the entity type from each table name and save to list

jsdonn (commented):
some initial comments

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -256,39 +388,70 @@ private <SNAPSHOT extends RecordTemplate> SNAPSHOT constructSnapshot(@Nonnull fi
...
+      @Nullable final LocalRelationshipFilter crewEntityFilter) {
 
     StringBuilder sqlBuilder = new StringBuilder();
     sqlBuilder.append("SELECT rt.* FROM ").append(relationshipTableName).append(" rt ");
 
-    if (destTableName != null) {
+    List<Pair<LocalRelationshipFilter, String>> filters = new ArrayList<>();
+
+    // destTableName and destinationEntityFilter should always come in pair. Since we are joining and filtering on the same table.
+    if (destTableName != null && destinationEntityFilter != null) {
```
jsdonn, line 420: We can have dest type (destTableName) but without any filters

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -256,39 +388,70 @@ private <SNAPSHOT extends RecordTemplate> SNAPSHOT constructSnapshot(@Nonnull fi
...
    */
   @Nonnull
-  private String buildFindRelationshipSQL(@Nullable final String destTableName, @Nullable final String sourceTableName,
-      @Nonnull final String relationshipTableName, @Nonnull final LocalRelationshipFilter sourceEntityFilter,
-      @Nonnull final LocalRelationshipFilter destinationEntityFilter, @Nonnull final LocalRelationshipFilter relationshipFilter) {
+  private String buildFindRelationshipSQL(
+      @Nonnull final String relationshipTableName, @Nonnull final LocalRelationshipFilter relationshipFilter,
+      @Nullable final String sourceTableName, @Nullable final LocalRelationshipFilter sourceEntityFilter,
+      @Nullable final String destTableName, @Nullable final LocalRelationshipFilter destinationEntityFilter,
+      @Nullable final LocalRelationshipFilter crewEntityFilter) {
```
jsdonn, line 412: Remove entity-specific stuff

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EntityType.java
```diff
@@ -0,0 +1,7 @@
+package com.linkedin.metadata.dao;
+
+public enum EntityType {
```
jsdonn, line 3: If you still decide to use an enum, we can just define it in the EbeanLocalRelationshipQueryDAO unless you think it will be used in other places.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -184,7 +188,135 @@ public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationship
       int offset, int count) {
     // NOTE: additional validation for sourceEntityUrn and sourceEntityUrn first.
     // for non-MG entities, filters need to be null or ignored.
-    throw new RuntimeException("findRelationships is not implemented.");
+    EntityType sourceEntityType = validateAndCategorizeEntityUrn(sourceEntityFilter, sourceEntityUrn);
+    EntityType destEntityType = validateAndCategorizeEntityUrn(destinationEntityFilter, destinationEntityUrn);
+    validateEntityFilter(relationshipFilter, relationshipType);
```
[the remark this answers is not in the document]
ybz1013: I copied this from the old findRelationships API. I think its a type. Fixed for old and new APIs.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -256,39 +345,77 @@ private <SNAPSHOT extends RecordTemplate> SNAPSHOT constructSnapshot(@Nonnull fi
...
+
     if (destTableName != null) {
       sqlBuilder.append("INNER JOIN ").append(destTableName).append(" dt ON dt.urn=rt.destination ");
+
+      if (destinationEntityFilter != null) {
+        filters.add(new Pair<>(destinationEntityFilter, "dt"));
+      }
+    } else if (destinationEntityFilter != null) {
+      // non-mg entity case, applying dest filter on relationship table
+      filters.add(new Pair<>(destinationEntityFilter, "rt"));
```
jsdonn, line 382: Wait can you explain this? why are we applying the destination entity filter on the relationship table?

jsdonn (commented):
Lemme take a look at unit tests after you address these initial comments

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: since the destTableName is null, in this case the destination entity filter at this point will only have an non-mg entity urn. We will use that urn as a condition in the relationship table.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -174,17 +184,65 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+   */
+  @VisibleForTesting
+  protected boolean isMgEntityUrn(@Nonnull Urn entityUrn) {
+    // there is some race condition, the local relationship db might not be ready when EbeanLocalRelationshipQueryDAO inits.
+    // so we can't init the _mgEntityTypeNameSet in constructor.
+    if (_mgEntityTypeNameSet == null) {
+      initMgEntityTypeNameSet();
+    }
+
+    assert _mgEntityTypeNameSet != null;
```
jsdonn, line 231: I don't think we need this.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -174,17 +184,65 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+    return _mgEntityTypeNameSet.contains(entityUrn.getEntityType());
```
jsdonn, line 232: Normalize? I think everything in the set will be lowercase like `datasetinstance` but getEntityType might return `DatasetInstance` or `datasetInstance`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -256,39 +360,86 @@ private <SNAPSHOT extends RecordTemplate> SNAPSHOT constructSnapshot(@Nonnull fi
...
-        new Pair<>(sourceEntityFilter, "st"),
-        new Pair<>(destinationEntityFilter, "dt"),
-        new Pair<>(relationshipFilter, "rt"));
+        filters.toArray(new Pair[filters.size()]));
 
     if (whereClause != null) {
       sqlBuilder.append(" AND ").append(whereClause);
     }
+
+    if (limit > 0) {
```
jsdonn, line 424: hopefully limit is more than 0 or else it will return 0 results lol

jsdonn (commented):
good tests. can you please add 1 more test where you try to apply a non-urn filter on non-mg entity use case? it should fail with validation exception

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: forgot to remove it

jsdonn (approved):
LGTM, thanks for addressing all the comments, good work!

PR #391, opened 2024-07-24
feat(local relationship): support local relationship tables in old schema

jsdonn:
## Summary
As part of the graph query service, we want to support local relationship tables in all GMSes, regardless of which database schema mode they are on (i.e. NEW_SCHEMA_ONLY vs OLD_SCHEMA_ONLY). This requires adding support for local relationship ingestion and querying in the old schema.

### Ingestion
Before: local relationship ingestion was bundled with new schema entity table ingestion in EbeanLocalAccess. While bundled in the same `add` method, there is no technical dependency between the two. 

After: local relationship ingestion has been moved to `EbeanLocalDAO::saveLatest`. The logic is the same:
1) check if the incoming update is a soft deletion.
2) if soft deletion, check if the aspect to be soft-deleted has any relationships tied to it.
3) if yes, throw an error. No aspect soft deletion allowed if it is associated with local relationships.
4) ingest aspect as usual (into entity tables in new schema or metadata_aspect in old schema)
5) ingest local relationships using the same logic as before.

### Query
Before: querying for local relationships using the findRelationships API would join multiple tables together (relationship + entity tables) to satisfy any filter conditions. This logic was built specifically for the new schema.

After: EbeanLocalRelationshipQueryDAO now has a schemaConfig similar to EbeanLocalDAO, which can take on the values of NEW_SCHEMA_ONLY, DUAL_SCHEMA, and OLD_SCHEMA_ONLY. It can be set using `setSchemaConfig()`. findRelationships will use different SQL queries depending on the value of schemaConfig. In new schema mode, everything is the same as before.

In old schema mode, there are certain restrictions. Only 1 filter is allowed per source or destination entity. That filter, if it exists, MUST be on the urn field (and not the aspect field for example). Meaning - no joins are necessary; all information in the allowed query can be found just by looking at the local relationship table. This restriction is in place because filtering on metadata_aspect is not feasible since the `metadata` column is not indexed. Thus, the resulting SQL will look something like:
```
SELECT rt.* FROM metadata_relationship_<relationship_type> rt
WHERE deleted_ts IS NULL
{ (optional) AND rt.source = urn:li:<sourceType>:1 }
{ (optional) AND rt.destination = urn:li:<destinationType>:2 }
```

### Other Details
Introduced RelationshipSource enum which can have 2 values: `RELATIONSHIP_BUILDERS` and `ASPECT_METADATA`. This enum will be used to tell the DAO where to ingest relationship metadata from. Right now, all relationship data comes from looking up an Aspect class in a local relationship builder registry to retrieve the corresponding relationship builder. Then, the relationship is built using that relationship builder and ingested.

In the future, relationships will be stored directly in the aspect metadata so maintaining relationship builders and a registry will not be necessary. However, I will leave the implementation of supporting the `ASPECT_METADATA` enum for a later PR.

## Testing Done
Updated/refactored some unit tests.

## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -120,6 +133,16 @@ public boolean isChangeLogEnabled() {
     return _changeLogEnabled;
   }
 
+  /**
+   * Set where the relationships should be derived from during ingestion. Either from aspect models or from relationship
+   * builders. The default is relationship builders.
+   * @param relationshipSource {@link RelationshipSource ASPECT_METADATA or RELATIONSHIP_BUILDERS}
+   */
+  public void setRelationshipSource(RelationshipSource relationshipSource) {
```
yangyangv2, line 141: should this be determined at DAO construction time? and marked as final?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -570,6 +597,22 @@ protected <T> T runInTransactionWithRetry(@Nonnull Supplier<T> block, int maxTra
...
+      if (_relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS
+          && _localRelationshipBuilderRegistry != null
+          && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
+        throw new UnsupportedOperationException(
+            String.format("Aspect %s cannot be soft-deleted because it has a local relationship builder registered.",
+                aspectClass.getCanonicalName()));
+      }
+
+      if (_relationshipSource == RelationshipSource.ASPECT_METADATA) {
+        // not yet implemented
```
yangyangv2, line 612: would this break anything if we push the code?

yangyangv2, line 612: throw UnsupportedOperationException? and add a TODO?

to avoid incorrect setting on ASPECT_METADATA but forget to implement this.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -817,6 +864,22 @@ protected <ASPECT extends RecordTemplate> void insert(@Nonnull URN urn, @Nullabl
...
+    if (_relationshipSource == RelationshipSource.ASPECT_METADATA) {
+      // not yet implemented
+    } else if (_relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS) {
+      if (_localRelationshipBuilderRegistry != null && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
+        List<LocalRelationshipUpdates> localRelationshipUpdates =
+            _localRelationshipBuilderRegistry.getLocalRelationshipBuilder(aspect).buildRelationships(urn, aspect);
+        _localRelationshipWriterDAO.processLocalRelationshipUpdates(urn, localRelationshipUpdates);
+        return localRelationshipUpdates;
+      }
+    }
```
yangyangv2, line 879:
```
else {
 UnsupportedOperationException
}
```

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -961,10 +1024,12 @@ public void setQueryKeysCount(int keysCount) {
   }
 
   /**
-   * Set a local relationship builder registry.
+   * Provide a local relationship builder registry. Local relationships will be built based on the builders during data ingestion.
+   * @param localRelationshipBuilderRegistry All local relationship builders should be registered in this registry.
+   *                                         Can be set to null to turn off local relationship ingestion.
```
yangyangv2, line 1029: in which scenario we want to disable the local relationship? routing GMS?

this is little implicit operation control --- should we keep the registry nonNull but have an more explicit control to enable/disable local relationship ingestion?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -80,6 +87,9 @@ public EbeanLocalRelationshipQueryDAO(EbeanServer server) {
   @Nonnull
   public <SNAPSHOT extends RecordTemplate> List<SNAPSHOT> findEntities(@Nonnull Class<SNAPSHOT> snapshotClass,
       @Nonnull LocalRelationshipFilter filter, int offset, int count) {
+    if (_schemaConfig == EbeanLocalDAO.SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("findEntities is not supported in OLD_SCHEMA_MODE");
```
yangyangv2, line 91: should this be a Checked Exception to make sure the OLD_SCHEMA can properly handle it? or can we be sure OLD_SCHEMA will not run into this logic

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -104,6 +114,9 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
       @Nonnull Class<DEST_SNAPSHOT> destinationEntityClass, @Nonnull LocalRelationshipFilter destinationEntityFilter,
       @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter, int minHops,
       int maxHops, int offset, int count) {
+    if (_schemaConfig == EbeanLocalDAO.SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException("findEntities is not supported in OLD_SCHEMA_MODE");
```
yangyangv2, line 118: same as line 91

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -418,6 +422,34 @@ public static String whereClause(@Nonnull LocalRelationshipFilter filter,
     return String.join(" AND ", andClauses);
   }
 
+  /**
+   * Construct the where clause SQL from a filter when running in old schema mode. Assumes that all filters are applied on
+   * urn fields, thus only the relationship table needs to be queried.
```
yangyangv2, line 427: are  urn fields == either source or destination? can you clarify in the comment?

yangyangv2 (commented):
minor comments, thanks for the hard / great work!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: We already have way too many EbeanLocalDAO constructors. I will keep it as a setter method and mark in the comment that it should only be used during construction time in the dao factory.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: Added UnsupportedOperationException

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: done

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: Can you clarify what you mean by checked exception?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
jsdonn: Clarified `Urn fields refers to the source or destination urn in a relationship query.`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: We want to disable the local relationships when a relationship is not ready to be onboarded yet (like when the table has not been created yet). 

In the future, we might not need this; we can just assume that all relationships will be ingested. But for now, while we still have ES graph and MySQL local relationship tables, we only want to ingest the relationships that are actually onboarded to MySQL local relationships.

yangyangv2: makes sense, can you include a brief explanation in the comment?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
yangyangv2: the caller needs to explicitly catch and handle the exception.

jsdonn: added a `throws UnsupportedOperationException` to the method signature

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: added a `throws UnsupportedOperationException` to the method signature

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
yangyangv2: UnsupportedOperationException is a "RuntimeException" and it does not make caller to explicitly catch it. You can either switch to some exception that extends Exception instead of RuntimeException. 

My intention was a question more making the client to be cautious when using this findEntity methods as they may be running on old schema, otherwise they'll see this issue at runtime.. But if you think this risk is low, then you can stay with the Runtime Exception, but make the documentation clear.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
yangyangv2: same above

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: Understood. Changed to `OperationNotSupportedException` from javax which extends `Exception`. Also added to the javadocs.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: Changed to `OperationNotSupportedException` from javax which extends `Exception`. Also added to the javadocs.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: In general, I don't expect many non-MG users to use the localrelationship dao. As far as I know only the lineage accessor in TMS uses it and we have control over that.

yangyangv2 (approved):
LGTM, thanks for addressing the comments.

PR #410, opened 2024-08-29
Injecting PreIngestionLambda in BaseLocalDao

rakhiagr:
## Summary
Injecting PreIngestionLambda in BaseLocalDao
## Testing Done
./gradlew build
## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -1,6 +1,7 @@
 package com.linkedin.metadata.dao;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.Message;
 import com.linkedin.common.AuditStamp;
 import com.linkedin.common.urn.Urn;
 import com.linkedin.data.DataMap;
@@ -24,6 +25,8 @@ import com.linkedin.metadata.dao.equality.EqualityTester;
 import com.linkedin.metadata.dao.exception.ModelValidationException;
 import com.linkedin.metadata.dao.ingestion.BaseLambdaFunction;
 import com.linkedin.metadata.dao.ingestion.LambdaFunctionRegistry;
+import com.linkedin.metadata.dao.ingestion.PreIngestionAspectRegistry;
+import com.linkedin.metadata.dao.ingestion.RestliCompliantPreUpdateRoutingClient;
 import com.linkedin.metadata.dao.producer.BaseMetadataEventProducer;
 import com.linkedin.metadata.dao.producer.BaseTrackingMetadataEventProducer;
 import com.linkedin.metadata.dao.retention.IndefiniteRetention;
@@ -179,6 +182,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   protected UrnPathExtractor<URN> _urnPathExtractor;
 
   private LambdaFunctionRegistry _lambdaFunctionRegistry;
+  private PreIngestionAspectRegistry _preIngestionAspectRegistry;
 
   // Maps an aspect class to the corresponding retention policy
   private final Map<Class<? extends RecordTemplate>, Retention> _aspectRetentionMap = new HashMap<>();
@@ -210,6 +214,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   private boolean _emitAuditEvent = false;
 
   private Clock _clock = Clock.systemUTC();
+
+  private RestliCompliantPreUpdateRoutingClient _restliCompliantPreUpdateRoutingClient;
 
   /**
    * Constructor for BaseLocalDAO.
@@ -395,6 +401,20 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     _lambdaFunctionRegistry = lambdaFunctionRegistry;
   }
 
+  /**
+   * Set Pre Update Routing Client.
+   */
+  public void setPreUpdateRoutingClient(
+      @Nullable RestliCompliantPreUpdateRoutingClient<Message> restliCompliantPreUpdateRoutingClient) {
+    _restliCompliantPreUpdateRoutingClient = restliCompliantPreUpdateRoutingClient;
+  }
+
+  /**
+   * Set Pre Ingestion Aspect Registry.
+   */
+  public void setPreIngestionAspectRegistry(@Nullable PreIngestionAspectRegistry preIngestionAspectRegistry) {
+    _preIngestionAspectRegistry = preIngestionAspectRegistry;
+  }
   /**
    * Enables or disables atomic updates of multiple aspects.
    */
@@ -744,8 +764,21 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdateLambda<ASPECT> updateLambda,
       @Nonnull AuditStamp auditStamp, int maxTransactionRetry, @Nullable IngestionTrackingContext trackingContext) {
     checkValidAspect(updateLambda.getAspectClass());
-
-    final AddResult<ASPECT> result = runInTransactionWithRetry(() -> aspectUpdateHelper(urn, updateLambda, auditStamp, trackingContext),
+    AspectUpdateLambda<ASPECT> effectiveUpdateLambda;
+    if (_preIngestionAspectRegistry != null
+        && _preIngestionAspectRegistry.isRegistered(updateLambda.getAspectClass())) {
+      Message updatedAspect = _restliCompliantPreUpdateRoutingClient.routingLambda(
+          _restliCompliantPreUpdateRoutingClient.convertUrnToMessage(urn),
+          _restliCompliantPreUpdateRoutingClient.convertAspectToMessage(
+              updateLambda.getUpdateLambda().apply(Optional.empty())));
+      ASPECT convertedAspect = (ASPECT) _restliCompliantPreUpdateRoutingClient.convertAspectFromMessage(updatedAspect);
+      effectiveUpdateLambda = new AspectUpdateLambda<>((Class<ASPECT>) convertedAspect.getClass(), // Aspect class
+          ignored -> convertedAspect // Lambda that returns the converted aspect
+      );
+    } else {
+      effectiveUpdateLambda = updateLambda;
+    }
+    final AddResult<ASPECT> result = runInTransactionWithRetry(() -> aspectUpdateHelper(urn, effectiveUpdateLambda, auditStamp, trackingContext),
         maxTransactionRetry);
 
     return unwrapAddResult(urn, result, auditStamp, trackingContext);
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreIngestionAspectRegistry.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreIngestionAspectRegistry.java
@@ -1,0 +1,9 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.linkedin.data.template.RecordTemplate;
+import javax.annotation.Nonnull;
+
+
+public interface PreIngestionAspectRegistry {
+  <ASPECT extends RecordTemplate> boolean isRegistered(@Nonnull final Class<ASPECT> aspectClass);
+}
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/RestliCompliantPreUpdateRoutingClient.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/RestliCompliantPreUpdateRoutingClient.java
@@ -1,0 +1,15 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.google.protobuf.Message;
+
+import com.linkedin.common.urn.Urn;
+import com.linkedin.data.template.RecordTemplate;
+
+
+public interface RestliCompliantPreUpdateRoutingClient<ASPECT extends Message> extends PreUpdateRoutingClient {
+  Message convertUrnToMessage(Urn pegasusUrn);
+
+  ASPECT convertAspectToMessage(RecordTemplate pegasusAspect);
+
+  RecordTemplate convertAspectFromMessage(ASPECT messageAspect);
+}
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreUpdateRoutingClient.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreUpdateRoutingClient.java
@@ -13,7 +13,6 @@ public interface PreUpdateRoutingClient<ASPECT extends Message> {
    * @param urn the urn of the asset
    * @param aspect the aspect to be updated
    * @return the updated aspect
-   * @throws Exception if the routing fails
    */
-  ASPECT routingLambda(Message urn, ASPECT aspect) throws Exception;
+  ASPECT routingLambda(Message urn, ASPECT aspect);
 }


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -744,8 +764,21 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
   public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdateLambda<ASPECT> updateLambda,
       @Nonnull AuditStamp auditStamp, int maxTransactionRetry, @Nullable IngestionTrackingContext trackingContext) {
     checkValidAspect(updateLambda.getAspectClass());
-
-    final AddResult<ASPECT> result = runInTransactionWithRetry(() -> aspectUpdateHelper(urn, updateLambda, auditStamp, trackingContext),
+    AspectUpdateLambda<ASPECT> effectiveUpdateLambda;
+    if (_preIngestionAspectRegistry != null
+        && _preIngestionAspectRegistry.isRegistered(updateLambda.getAspectClass())) {
+      Message updatedAspect = _restliCompliantPreUpdateRoutingClient.routingLambda(
```
JiaoMaWHU, line 770: `PreIngestionAspectRegistry` can take multiple aspects as the input

`PreUpdateRoutingClient` is tied to a specific aspect but here applied for all aspects

I think you would need to follow this [piece](https://jarvis.corp.linkedin.com/codesearch/result/?name=BaseLocalDAO.java&path=datahub-gma%2Fdao-api%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fmetadata%2Fdao&reponame=linkedin%2Fdatahub-gma#1597) to resolve the gap

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreIngestionAspectRegistry.java
```diff
@@ -0,0 +1,14 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.linkedin.data.template.RecordTemplate;
+import javax.annotation.Nonnull;
+
+/**
+ * A registry maintains which aspect needs PreIngestionRouting.
+ */
+public interface PreIngestionAspectRegistry {
```
JiaoMaWHU, line 9: I guess this would restli specific? do you agree?

if so let's name this as `RestliPreIngestionAspectRegistry`, you would need a new registry class for grpc

if you could think of a way to consolidate restli and grpc into one single class, that would be perfect

rakhiagr: I could do something like 
```
public interface PreIngestionAspectRegistry {
    /**
     * Check if routing is registered for an aspect.
     */
    <ASPECT> boolean isRegistered(Class<ASPECT> aspectClass);
}
```
and then have specific interfaces for Restli and GRPC
```
public interface RestliPreIngestionAspectRegistry extends PreIngestionAspectRegistry {
    @Override
    <ASPECT extends RecordTemplate> boolean isRegistered(Class<ASPECT> aspectClass);
}
```
```
public interface GrpcPreIngestionAspectRegistry extends PreIngestionAspectRegistry {
    @Override
    <ASPECT extends Message> boolean isRegistered(Class<ASPECT> aspectClass);
}
```

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/RestliCompliantPreUpdateRoutingClient.java
```diff
@@ -0,0 +1,24 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.google.protobuf.Message;
+
+import com.linkedin.common.urn.Urn;
+import com.linkedin.data.template.RecordTemplate;
+
+
+public interface RestliCompliantPreUpdateRoutingClient<ASPECT extends Message> extends PreUpdateRoutingClient {
```
JiaoMaWHU, line 9: curious, do you need to pass ASPECT to `PreUpdateRoutingClient` in the inheritance?

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreIngestionAspectRegistry.java
JiaoMaWHU: would PreIngestionAspectRegistry work? 

will `<ASPECT extends RecordTemplate> boolean isRegistered(Class<ASPECT> aspectClass);` override `<ASPECT> boolean isRegistered(Class<ASPECT> aspectClass);`

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
rakhiagr: Am I missing something?

PreUpdateRoutingClient.routingLamda() can take one aspect at a time. And so is `PreIngestionAspectRegistry`.

JiaoMaWHU: the generic type in your PreIngestionAspectRegistry is at method level

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/PreIngestionAspectRegistry.java
rakhiagr: Good point. I did some search, and I don't think it would work as "For overriding methods, the method signature, including generic type parameters, must match exactly between the interface and its implementations"

```
public interface PreIngestionAspectRegistry {
    boolean isRegistered(Class<?> aspectClass);
}
```

This could be another way I can think of. Or we can keep it simple and have 2 separate `RestliPreIngestionAspectRegistry` and `GrpcPreIngestionAspectRegistry`

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/RestliCompliantPreUpdateRoutingClient.java
rakhiagr: That's what ChatGPT told me. 
"
If you’re extending from a generic interface, you should indeed pass the generic type parameter ASPECT in the inheritance.

all generic types are specified in both the parent and extending interfaces to maintain type safety and consistency across implementations.
"

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -211,6 +215,8 @@ public static class AspectUpdateLambda<ASPECT extends RecordTemplate> {
 
   private Clock _clock = Clock.systemUTC();
 
+  private RestliCompliantPreUpdateRoutingClient _restliCompliantPreUpdateRoutingClient;
```
JiaoMaWHU, line 218: this is not needed, right?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1606,4 +1626,24 @@ protected <ASPECT extends RecordTemplate> ASPECT updatePreIngestionLambdas(@Nonn
...
+  /**
+   * Route the aspect update lambda to the appropriate pre-ingestion aspect service.
+   */
+  protected <ASPECT extends RecordTemplate> AspectUpdateLambda<ASPECT> preUpdateLambda(URN urn,
+      AspectUpdateLambda<ASPECT> updateLambda) {
+    if (_restliPreIngestionAspectRegistry != null && _restliPreIngestionAspectRegistry.isRegistered(
+        updateLambda.getAspectClass())) {
+      RestliCompliantPreUpdateRoutingClient client =
+          _restliPreIngestionAspectRegistry.getPreIngestionRouting(updateLambda.getUpdateLambda().apply(Optional.empty()));
+      Message updatedAspect = client.routingLambda(_restliCompliantPreUpdateRoutingClient.convertUrnToMessage(urn),
```
JiaoMaWHU, line 1639: change the use of _restliCompliantPreUpdateRoutingClient to client?

JiaoMaWHU (commented):
you will need to update this part as well: https://github.com/linkedin/datahub-gma/blob/9dcb6aa05e583c81ffd4c88e38c1e5605245cd53/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectRoutingResource.java#L207

current change in the BaseLocalDao only covers the non-routing part

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -395,6 +399,15 @@ public void setLambdaFunctionRegistry(@Nullable LambdaFunctionRegistry lambdaFun
     _lambdaFunctionRegistry = lambdaFunctionRegistry;
   }
 
+  /**
+   * Set pre ingestion aspect registry.
+   */
+  public void setRestliPreIngestionAspectRegistry(
+      @Nullable RestliPreUpdateAspectRegistry restliPreUpdateAspectRegistry) {
```
JiaoMaWHU, line 406: why `@Nullable`?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -745,17 +758,18 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
...
-      runInTransactionWithRetry(() -> aspectUpdateHelper(urn, updateLambda, auditStamp, trackingContext),
-          maxTransactionRetry);
-    }
-    final AddResult<ASPECT> result = runInTransactionWithRetry(() -> aspectUpdateHelper(urn,
-        new AspectUpdateLambda<>(aspectClass, updateLambda.getUpdateLambda(),
-            updateLambda.getIngestionParams().setTestMode(false)), auditStamp, trackingContext), maxTransactionRetry);
-    return unwrapAddResult(urn, result, auditStamp, trackingContext);
+      final Class<ASPECT> aspectClass = updateLambda.getAspectClass();
+      checkValidAspect(aspectClass);
+      AspectUpdateLambda<ASPECT> effectiveUpdateLambda = preUpdateRouting(urn, updateLambda);
```
JiaoMaWHU, line 763: `preIngestionRoutedUpdateLambda` or `updateLambdaAfterPreIngestionRouting`?

JiaoMaWHU, line 763: Let's move this line to https://github.com/linkedin/datahub-gma/blob/be56ee0ccc227cda1a40edcf81225a60731f26c3/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java#L826

the `preUpdateRouting` helper method here assumes all lambda functions are `ignore -> newValue`, therefore could potentially overwrite lambdas with other content 

in addition, to avoid any inconsistency of the lambda between your helper and the add method
- option 1. add comment to reminds devs of updating `preUpdateRouting` as well if they update `ignore -> newValue`. if you can think of better options, that's also ok
- option 2. update it to
```
PreUpdateRoutedValue preUpdateRoutedValue = helper(newValue)
return add(urn, (Class<ASPECT>) newValue.getClass(), ignored -> PreUpdateRoutedValue, auditStamp, trackingContext, nonNullIngestionParams);
```

option 2 seems better, if you could think of even better options, that's also ok

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/RestliCompliantPreUpdateRoutingClient.java
```diff
@@ -0,0 +1,24 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.google.protobuf.Message;
+
+import com.linkedin.common.urn.Urn;
+import com.linkedin.data.template.RecordTemplate;
+
+
```
JiaoMaWHU, line 8: java doc, please specify what is the difference between this class and the parent interface, so that users understand which class to use in different use cases

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
rakhiagr: To indicate that the restliPreUpdateAspectRegistry parameter can be null.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
rakhiagr: Good point, going with option 2 here.

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -650,4 +651,31 @@ public void testPreIngestionLambda() throws URISyntaxException {
...
+    _dummyLocalDAO.setRestliPreIngestionAspectRegistry(new SamplePreUpdateAspectRegistryImpl());
+    expectGetLatest(urn, AspectFoo.class,
+        Arrays.asList(makeAspectEntry(null, null), makeAspectEntry(foo, _dummyAuditStamp)));
+
+    _dummyLocalDAO.add(urn, foo, _dummyAuditStamp);
+
+    verify(_mockEventProducer, times(1)).produceMetadataAuditEvent(urn, null, bar);
+    verify(_mockEventProducer, times(1)).produceAspectSpecificMetadataAuditEvent(urn, null, bar, _dummyAuditStamp, IngestionMode.LIVE);
+    verifyNoMoreInteractions(_mockEventProducer);
+  }
```
JiaoMaWHU, line 680: missing a test case where `RestliPreIngestionAspectRegistry` is injected but the aspect to ingest is not registered

rakhiagr: Added.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -833,7 +848,8 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, @Nonnull ASP
     final IngestionParams nonNullIngestionParams =
         ingestionParams == null || !ingestionParams.hasTestMode() ? new IngestionParams().setIngestionMode(
             IngestionMode.LIVE).setTestMode(false) : ingestionParams;
-    return add(urn, (Class<ASPECT>) newValue.getClass(), ignored -> newValue, auditStamp, trackingContext, nonNullIngestionParams);
+    ASPECT effectiveUpdateLambda = preUpdateRouting(urn, newValue);
```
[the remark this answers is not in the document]
rakhiagr: Updated

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
rakhiagr: So basically there are 2 approaches here
**Option 1:**
```
private RestliPreUpdateAspectRegistry _restliPreUpdateAspectRegistry;
  
public void setRestliPreIngestionAspectRegistry(
    @Nullable RestliPreUpdateAspectRegistry restliPreUpdateAspectRegistry) {
    _restliPreUpdateAspectRegistry = restliPreUpdateAspectRegistry;
}
```
**Option 2:**
```
@Nonnull
private RestliPreUpdateAspectRegistry _restliPreUpdateAspectRegistry = null;
  
public void setRestliPreIngestionAspectRegistry(
    RestliPreUpdateAspectRegistry restliPreUpdateAspectRegistry) {
    _restliPreUpdateAspectRegistry = restliPreUpdateAspectRegistry;
}
```

I feel both the approaches have their pros and cons. Option 1 is better if you anticipate the registry might not always be set (or set later). Anyway we have proper null checks in the codebase. Also, for option 2, the initialization to null goes against the intent of @Nonnull, which can cause issues during runtime if the field is accessed before the setter is called.

Even _lambdaFunctionRegistry is set using Nullable. 
https://github.com/linkedin/datahub-gma/blob/3c10b71586494796dc31f3ef976cb91bf97561e2/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java#L394

Let me know if you think otherwise.

JiaoMaWHU: I mean
```
private RestliPreUpdateAspectRegistry _restliPreUpdateAspectRegistry = null;
  
public void setRestliPreIngestionAspectRegistry(
    @Nonnull RestliPreUpdateAspectRegistry restliPreUpdateAspectRegistry) {
    _restliPreUpdateAspectRegistry = restliPreUpdateAspectRegistry;
}
```

JiaoMaWHU (approved):

PR #425, opened 2024-09-12
fix(findRelationshipsV2): change source/dest params from Urn to String

jsdonn:
## Summary
There are cases when a user doesn't need to provide an urn for the destination. For example, they only specify the destination type (like "dataset") but without also providing a filter on the UrnField (like urn = "urn:li:dataset:foo"). The current implementation does not support this as it expects either an Urn or null for the source and destination entities. If null, there's no way to tell the API which type of destination to look for.

## Testing Done
adjusted unit tests. ./gradlew build
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
@@ -356,7 +356,7 @@ public class EbeanLocalRelationshipQueryDAOTest {
     _localRelationshipQueryDAO.setSchemaConfig(schemaConfig);
 
     List<ReportsTo> reportsToAlice = _localRelationshipQueryDAO.findRelationshipsV2(
-        null, null, fooEntityUrn, destFilter,
+        null, null, "foo", destFilter,
         ReportsTo.class, new LocalRelationshipFilter().setCriteria(new LocalRelationshipCriterionArray()).setDirection(RelationshipDirection.UNDIRECTED),
         -1, -1);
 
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -526,7 +526,7 @@ OwnedBy.class, new LocalRelationshipFilter().setCriteria(new LocalRelationshipCr
     LocalRelationshipFilter filterUrn2 = new LocalRelationshipFilter().setCriteria(new LocalRelationshipCriterionArray(filterUrnCriterion2));
 
     // test owned by of crew2 can be found
-    List<OwnedBy> ownedByCrew2 = _localRelationshipQueryDAO.findRelationshipsV2(null, null, crewEntityUrn, filterUrn2,
+    List<OwnedBy> ownedByCrew2 = _localRelationshipQueryDAO.findRelationshipsV2(null, null, "crew", filterUrn2,
```
ybz1013, line 529: for the 3rd arg, are we passing in crew or leave it null? 
If it doesn't change the implementation, can we leave it null? So in global API I wont have to extract it from urn. Thanks

ybz1013 (commented):
one question otherwise LGTM

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
jsdonn: The previous filter validation logic stated that "if entity urn is null, the filter must also be null. We should continue to enforce this for the case where the asset is a registered MG asset type.

So in the GQS, we need to pass in any non-null and non-empty string as this param if we want the query to find non-MG asset types.

ybz1013 (approved):

PR #433, opened 2024-09-23
Read column from aspect-alias instead of column annotation - part 1

yangyangv2:
## Summary

Read column from aspect-alias instead of column annotation - part 1

This is Part 1 of the work, which takes additional parameter (assetType aka entityType) into the getAspectColumn method.

Part 2 of the work will be implementing a map which retrieve aspect-alias by assetType and aspect class

## Testing Done

./gradlew build

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #436, opened 2024-09-28
read column from aspect alias part 2

yangyangv2:
## Summary

this is the 2nd part of PR: https://github.com/linkedin/datahub-gma/pull/433.

## Implementation Details

it is using a reflection to load all the asset classes with an lazy loading cache to map the requested `aspect` to the `aspect_alias`, which is the DB column stored on the database.

TODO

1. The reflection of loading all the assets can be moved to Li's internal code, will need horizontal update on the GMSes
2. Need a comprehensive unit test on metadata-models to ensure the calculated aspect_alias names are matching the "column" annotation.

## Testing Done

./gradlew build

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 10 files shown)


jphui (approved):
lgtm, curious if this will have to be re-implemented when Proto models become our SOT... from my understanding it looks like you're using the Proto->Pegasus classes right now rather than the Proto directly.

yangyangv2:
> lgtm, curious if this will have to be re-implemented when Proto models become our SOT... from my understanding it looks like you're using the Proto->Pegasus classes right now rather than the Proto directly.

the DAO will stay with  pegasus for a while. once we fully switch to Porto (in the future), it will be a full re-write. Not in the current Porto migration scope

PR #442, opened 2024-10-01
Add RampedResourceImpl Lix to set New MG Kernal as default

jphui:
## Summary

title

NOTE that I'm bumping the version to `1.0.*`, feel like since this changes the default state to `RAMPED` this is called for?

## Testing Done

./gradlew. build

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #446, opened 2024-10-08
Consolidate resource-lix into datahub-gma

yangyangv2:
## Summary

1. Change default READ API behavior to Model 2.0
2. Consolidate resource-lix implementation into datahub-gma
3. add & fix unit tests

## Testing Done

./gradlew build

## Cleanup Steps

1. [P0] bump datahub-gms in metadata-models
2. [P0] bump corp-identity GMS or metric GMS with the following cleanup to make sure no regression
4. [P0] make sure AIM has getResource() to return LegacyResourceImpl(), so that AIM will be safe to bump metadata-models which has default READ API behavior changed to ramped. 
5. [P0] bump metadata-models on GMSes which needs to onboard new entities (e.g. service-gms, datahub-gms). So that new entities can get onboarded without using RampedResourceImpl()
6. [P1] bump metadata-models from all GMS with the cleanup
            - remove the ResourceLixFactory from each GMS
            - remove the _resourceLix, getResourceLix() on each resource endpoint.

## Checklist

### DIFF
--- a/restli-resources/src/test/java/com/linkedin/metadata/restli/BaseEntityResourceTest.java
+++ b/restli-resources/src/test/java/com/linkedin/metadata/restli/BaseEntityResourceTest.java
@@ -24,6 +24,7 @@ import com.linkedin.metadata.query.IndexGroupByCriterion;
 import com.linkedin.metadata.query.IndexSortCriterion;
 import com.linkedin.metadata.query.MapMetadata;
 import com.linkedin.metadata.query.SortOrder;
+import com.linkedin.metadata.restli.lix.LegacyResourceImpl;
 import com.linkedin.metadata.restli.lix.ResourceLix;
 import com.linkedin.parseq.BaseEngineTest;
 import com.linkedin.restli.common.ComplexResourceKey;
@@ -62,7 +63,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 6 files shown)


restli-resources/src/test/java/com/linkedin/metadata/restli/BaseEntityResourceTest.java
```diff
@@ -82,6 +82,11 @@ class TestResource extends
                      BaseEntityResource<ComplexResourceKey<EntityKey, EmptyRecord>, EntityValue, FooUrn, EntitySnapshot,
                          EntityAspectUnion, InternalEntitySnapshot, InternalEntityAspectUnion, EntityAsset> {
 
+    @Override
+    protected ResourceLix getResourceLix() {
+      return new LegacyResourceImpl();
+    }
+
```
jphui, line 89: When I was working on my version of this PR -- https://github.com/linkedin/datahub-gma/pull/442 -- about 11 or so of the tests were failing, the only difference is that you switched out 

```
    @Override
    protected ResourceLix getResourceLix() {
      return new ResourceLix() {
        @Override
        public boolean testGet(@Nonnull String urn, @Nonnull String entityType) {
          return true;
        }
...
```
with 

```
@Override
    protected ResourceLix getResourceLix() {
      return new LegacyResourceImpl();
    }
```

How is it possible that changing the Lix behavior doesn't require modifying tests?

jphui (approved):
LGTM, though

restli-resources/src/test/java/com/linkedin/metadata/restli/BaseEntityResourceTest.java
yangyangv2: I made the two mock resource TestResource and InternalTestResource to behave exactly the same as before.

- Previously default is legacy, then the getResource is made on InternalTestResource to return RampedResourceImpl
- Now, the default is ramped, then I move the getResource to TestResource to return LegacyResourceImpl

PR #452, opened 2024-10-18
feat(relationship): add support for extracting relationships from aspect metadata

jsdonn:
## Summary
Implement the relationship ingestion logic when the dao is configurated to get relationships from aspect metadata instead of by using relationship builders.

This configuration can be set by calling `EbeanLocalDAO::setRelationshipSource` with enum value `ASPECT_METADATA` (as opposed to the default `RELATIONSHIP_BUILDERS`). Note that this configuration will be applied to ALL aspects that the DAO processes (i.e. all aspects belonging to a single entity type).

Also add support to the EbeanDAOUtils::extractRelationshipsFromAspect method for the case where the relationship field is a single value (not a list).

I plan to implement a delete (soft delete) relationships API in a later PR.

## Testing Done
added a unit test for the relationship ingestion. also added the single relationship value case to EbeanDAOUtilsTest.
## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -388,8 +389,16 @@ public static <RELATIONSHIP extends RecordTemplate, ASPECT extends RecordTemplat
       try {
         Method getMethod = clazz.getMethod("get" + StringUtils.capitalize(fieldName)); // getFieldName
         Object obj = getMethod.invoke(aspect); // invoke getFieldName
-        // all relationship fields will be represented as Arrays so filter out any non-lists, empty lists, and lists that don't contain RecordTemplates
-        if (!(obj instanceof List) || ((List) obj).isEmpty() || !(((List) obj).get(0) instanceof RecordTemplate)) {
+        // all relationship fields will be represented as either singleton fields or Arrays so filter out any non-RecordTemplates,
```
hongliu2024, line 392: Question about this `extractRelationshipsFromAspect` method in general:
an aspect can have a status field to indicate that it's soft deleted. Should we check that and exclude relationships derived from soft deleted aspects?

hongliu2024: nvm. Saw your comment in EbeanLocalDAO 
```
  // TODO: not yet implemented -> add support for removing relationships when the aspect is to be soft-deleted
```

jsdonn: > Should we check that and exclude relationships derived from soft deleted aspects?

Just want to clarify that `status = removed` is slightly different than aspect being soft-deleted from the storage perspective. `status = removed` is still metadata. maybe DH interprets removed status as soft-deleted in their application, but from the MG point of view, `status = removed` is just like any other piece of metadata like `owner = jdonn`.

But you still bring up a valid question on actually soft-deleted aspects, how do we handle the relationships then? Well by definition when we get an aspect that contains relationships, it is not a soft-deleted aspect. Since the way we soft-delete aspects now is by ingesting an aspect with "null" value. But let's say we want to soft-delete an aspect which has relationships that have been ingested into the relationship tables. Should we also mark the corresponding relationships as deleted? I think yes since we are not treating relationship tables as just a reflection of whatever relationships are passed in through aspects (without any modification).

ybz1013 (approved):
lgtm

PR #455, opened 2024-10-21
Included Entity type in the routing map

rakhiagr:
## Summary
Every dao has the registry and lambda logic, and the aspect routing may only be needed for specific asset, so we can reduce the number of calls by filtering out using entity and aspect to check whether the routing client exists.
## Testing Done
./gradlew build
Unit test
## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/AspectCallbackMapKey.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/AspectCallbackMapKey.java
@@ -3,8 +3,47 @@ package com.linkedin.metadata.dao.ingestion;
 import com.linkedin.data.template.RecordTemplate;
 import lombok.Data;
 
+
+/**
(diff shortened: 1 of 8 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/AspectCallbackMapKey.java
```diff
@@ -0,0 +1,10 @@
+package com.linkedin.metadata.dao.ingestion;
+
+import com.linkedin.data.template.RecordTemplate;
+import lombok.Data;
+
+@Data
+public class AspectCallbackMapKey {
```
yihzhou, line 7: You need to rewrite the hashcode() and equals() for this class so the new aspect + entityType always get the same value in Hashmap. Example [here](https://www.techiedelight.com/use-object-as-key-hashmap-hashset-java/).

yihzhou (approved):
LGTM!

PR #456, opened 2024-10-23
soft delete relationships when aspects are soft deleted (ASPECT_METADATA only)

jsdonn:
## Summary
Previously, there was a gap when using relationship builders where an aspect that was registered with a relationship builder could not be soft deleted since there was no agreed-upon way to remove relationships associated with that aspect.

Now, when deriving relationships directly from aspect metadata, we can handle soft deletion gracefully. When the source aspect is marked as soft deleted, also mark any of the relationships from the old aspect value (existing in the db) as deleted in the corresponding relationship tables.

Note on the large number of files changed in this PR. I had to add a new test aspect model which required making many minor changes to other unit tests. No need to closely review those minor changes.

## Testing Done
added a unit test
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -611,10 +611,11 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
                 aspectClass.getCanonicalName()));
       }
 
-      if (_relationshipSource == RelationshipSource.ASPECT_METADATA) {
-        // TODO: not yet implemented -> add support for removing relationships when the aspect is to be soft-deleted
-        throw new UnsupportedOperationException("This method has not been implemented yet to support the "
-            + "ASPECT_METADATA RelationshipSource type yet.");
+      if (_relationshipSource == RelationshipSource.ASPECT_METADATA && oldValue != null) {
+        List<RecordTemplate> relationships = extractRelationshipsFromAspect(oldValue).stream()
+            .flatMap(List::stream)
+            .collect(Collectors.toList());
+        _localRelationshipWriterDAO.removeRelationships(relationships);
       }
     }
 
@@ -887,6 +888,9 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
    */
   public <ASPECT extends RecordTemplate, RELATIONSHIP extends RecordTemplate> List<LocalRelationshipUpdates> addRelationshipsIfAny(
       @Nonnull URN urn, @Nullable ASPECT aspect, @Nonnull Class<ASPECT> aspectClass, boolean isTestMode) {
+    if (aspect == null) {
+      return Collections.emptyList();
+    }
     List<LocalRelationshipUpdates> localRelationshipUpdates = Collections.emptyList();
     if (_relationshipSource == RelationshipSource.ASPECT_METADATA) {
       List<List<RELATIONSHIP>> allRelationships = EBeanDAOUtils.extractRelationshipsFromAspect(aspect);
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -601,22 +601,13 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
       @Nonnull AuditStamp newAuditStamp, boolean isSoftDeleted, @Nullable IngestionTrackingContext trackingContext,
       boolean isTestMode) {
     // First, check that if the aspect is going to be soft-deleted that it does not have any relationships derived from it.
-    // We currently don't support soft-deleting aspects from which local relationships are derived from.
-    if (newValue == null) {
-      if (_relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS
+    // We currently don't support soft-deleting aspects from which local relationships are derived via relationship builders.
+    if (newValue == null && _relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS
           && _localRelationshipBuilderRegistry != null
           && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
-        throw new UnsupportedOperationException(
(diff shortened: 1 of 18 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -611,10 +611,11 @@ protected <ASPECT extends RecordTemplate> long saveLatest(@Nonnull URN urn, @Non
                 aspectClass.getCanonicalName()));
       }
 
-      if (_relationshipSource == RelationshipSource.ASPECT_METADATA) {
-        // TODO: not yet implemented -> add support for removing relationships when the aspect is to be soft-deleted
-        throw new UnsupportedOperationException("This method has not been implemented yet to support the "
```
yangyangv2, line 616: just curious, did we run into this exception when someone's trying to soft delete the relationship?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -102,7 +103,13 @@ public <RELATIONSHIP extends RecordTemplate> void addRelationships(@Nonnull List
 
   @Override
   public <RELATIONSHIP extends RecordTemplate> void removeRelationships(@Nonnull List<RELATIONSHIP> relationships) {
-    throw new UnsupportedOperationException("Local relationship only supports adding relationships.");
+    for (RELATIONSHIP relationship : relationships) {
+      _server.createSqlUpdate(SQLStatementUtils.deleteLocalRelationshipSQL(SQLSchemaUtils.getRelationshipTableName(relationship),
+              RemovalOption.REMOVE_ALL_EDGES_FROM_SOURCE_TO_DESTINATION))
+          .setParameter(CommonColumnName.SOURCE, getSourceUrnFromRelationship(relationship).toString())
```
yangyangv2, line 109: does the new relationship model has (requires) source urn?

yangyangv2: or this implementation is only for legacy relationship models

jsdonn: Yes, we're going to keep the source urn in each 2.0 relationship model. You suggested:
```
message Relationship {
    string source = 1;
    oneof relatedTo {
        proto.com.linkedin.metadata.relationship.DownstreamOf = 2;
        proto.com.linkedin.metadata.relationship.OwnedBy = 3;
    }
}
```
and
```
message OwnedBy {
    oneof destination {
        proto.com.linkedin.common.CorpuserUrn = 1;
        proto.com.linkedin.devx.CrewUrn = 2;
    }
    Owner owner = 3;
}
```

But we think that from the it should be better from the modeling perspective to see both the source and destination in the relationship model like:
```
message OwnedBy {
    oneof source {
        proto.com.linkedin.common.DatasetUrn = 1;
    }
    oneof destination {
        proto.com.linkedin.common.CorpuserUrn = 2;
        proto.com.linkedin.devx.CrewUrn = 3;
    }
    Owner owner = 4;
}
```
and then it can be omitted from the top-level Relationship model.
```
message Relationship {
    oneof relatedTo {
        proto.com.linkedin.metadata.relationship.DownstreamOf = 1;
        proto.com.linkedin.metadata.relationship.OwnedBy = 2;
    }
}
```

jsdonn: Let me know your thoughts or if you have a strong opinion on this

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
jsdonn: No, this is under the `if (_relationshipSource == RelationshipSource.ASPECT_METADATA)` branch, which no one is using right now.

jsdonn: Although, there is a similar exception for relationship builders which probably we ran into a few times
```
      if (_relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS
          && _localRelationshipBuilderRegistry != null
          && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
        throw new UnsupportedOperationException(
            String.format("Aspect %s cannot be soft-deleted because it has a local relationship builder registered.",
                aspectClass.getCanonicalName()));
      }
```

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -611,10 +611,11 @@ protected <ASPECT extends RecordTemplate> long saveLatest(@Nonnull URN urn, @Non
...
-            + "ASPECT_METADATA RelationshipSource type yet.");
+      if (_relationshipSource == RelationshipSource.ASPECT_METADATA && oldValue != null) {
```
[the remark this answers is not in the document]
jsdonn: Thanks for the suggestion. Reordered to be 2 -> 1/3 (either or)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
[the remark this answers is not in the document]
jsdonn: Thanks for working on this. Please remember that we will probably need to handle both 1.0 and 2.0 models simultaneously during the migration phase.

hongliu2024 (approved):
Thanks for addressing my comment. LGTM

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
yangyangv2: The problem for having the source urn on the OwnedBy is it is a duplicated information with the host asset and you may lose the model constrains

what if 
```
DatasetAsset {
   urn: DatasetUrn ,

   ownership OwnedBy {
       source: CropUserUrn, ---> this is supposed to be DatasetUrn. 
       destination: DatasetUrn
   } 
}
```

PR #460, opened 2024-10-24
Fix: aspectUpdateHelper result is null due to isSkipProcessing enabled so unwrapAddResult() hits NPE

yihzhou:
## Summary
`aspectUpdateHelper` allows to return NULL when `isSkipProcessing` enabled so there is no real DAO save happen.
However the `unwrapAddResult()`doesn't expect result as NULL so it hits NPE. Also both MAE producing and post update hook should not happen because no real DAO happens.
Thus add NULL for return. 

## Testing Done
./gradlew idea

## Checklist


rakhiagr (approved):
LGTM

PR #462, opened 2024-10-28
support relationship v2

hongliu2024:
## Summary
This PR is to ensure the EbeanLocalWriterDAO can support both relationship V1 and relationship V2.
The changes are in Schema Validation for Relationship v2, RecordUtils, ModelUtils, and GraphUtils.

Question though is `checkSameUrn` logic in GraphUtils.java. They are related to the removal option, and for some options, they check for source Urn. But in Relationship model v2, there is no source field. Should it just skip the check for source field for V2, or should it look at the belonging aspect to find the source Urn? I feel looking at belonging aspect can have problems too because one relationship can be included in multiple different aspects.

## Testing Done
Unit test cases added.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/utils/GraphUtils.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/utils/GraphUtils.java
@@ -7,7 +7,6 @@ import java.util.List;
 import javax.annotation.Nonnull;
 
 import static com.linkedin.metadata.dao.utils.ModelUtils.*;
-import static com.linkedin.metadata.dao.utils.RecordUtils.*;
 
 
 public class GraphUtils {
@@ -25,6 +24,7 @@ public class GraphUtils {
       return;
     }
 
+    // ToDo: how to handle this for Relationship V2?
     final Urn sourceUrn = getSourceUrnFromRelationship(relationships.get(0));
     final Urn destinationUrn = getDestinationUrnFromRelationship(relationships.get(0));
 
@@ -41,7 +41,7 @@ public class GraphUtils {
   private static void checkSameUrn(@Nonnull List<? extends RecordTemplate> records, @Nonnull String field,
       @Nonnull Urn compare) {
     for (RecordTemplate relation : records) {
-      if (!compare.equals(getRecordTemplateField(relation, field, Urn.class))) {
+      if (!compare.equals(ModelUtils.getUrnFromRelationship(relation, field))) {
         throw new IllegalArgumentException("Records have different " + field + " urn");
       }
     }
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
@@ -44,9 +44,10 @@ public class ModelUtils {
 
   private static final ClassLoader CLASS_LOADER = DummySnapshot.class.getClassLoader();
   private static final String ASPECTS_FIELD = "aspects";
+  private static final String DESTINATION_FIELD = "destination";
   private static final String FIELD_FIELD_PREFIX = "FIELD_";
-  private static final String MEMBER_FIELD_PREFIX = "MEMBER_";
   private static final String METADATA_AUDIT_EVENT_PREFIX = "METADATA_AUDIT_EVENT";
+  private static final String SOURCE_FIELD = "source";
   private static final String URN_FIELD = "urn";
 
   private ModelUtils() {
@@ -293,7 +294,9 @@ public class ModelUtils {
   private static <RELATIONSHIP extends RecordTemplate> Urn getUrnFromRelationship(@Nonnull RELATIONSHIP relationship,
       @Nonnull String fieldName) {
     RelationshipValidator.validateRelationshipSchema(relationship.getClass());
-    final Urn urn = RecordUtils.getRecordTemplateField(relationship, fieldName, urnClassForRelationship(relationship.getClass(), fieldName));
+    boolean isRelationshipInV2 = isRelationshipInV2(relationship);
+    final Urn urn = RecordUtils.getRecordTemplateField(relationship, fieldName,
+        urnClassForRelationship(relationship.getClass(), fieldName, isRelationshipInV2));
     if (urn == null) {
       ValidationUtils.throwNullFieldException(URN_FIELD);
     }
@@ -306,7 +309,12 @@ public class ModelUtils {
   @Nonnull
   public static <RELATIONSHIP extends RecordTemplate> Urn getSourceUrnFromRelationship(
       @Nonnull RELATIONSHIP relationship) {
-    return getUrnFromRelationship(relationship, "source");
+    //ToDo: remove this if statement after all relationships are in Relationship model V2.
+    if (!isRelationshipInV2(relationship)) {
+      return getUrnFromRelationship(relationship, SOURCE_FIELD);
+    }
+
+    return getUrnFromRelationship(relationship, extractFieldNameFromUnionField(relationship, SOURCE_FIELD));
   }
 
   /**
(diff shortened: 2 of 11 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -846,6 +905,31 @@ public static boolean isCommonAspect(@Nonnull Class<? extends RecordTemplate> cl
...
+
+  /**
+   * Check if a given relationship schema is in MG model v2 or not.
+   * Returns TRUE if source and destination fields are of union type.
+   * @param schema schema of a valid relationship model defined in com.linkedin.metadata.relationship
+   * @return boolean. True if the relationship is in MG model V2.
+   */
+  static boolean isRelationshipInV2(@Nonnull RecordDataSchema schema) {
+    // check the data type of the source and destination fields in schema and see if it's a union type
+    return schema.getFields().stream().anyMatch(
```
jsdonn, line 926: Yang is saying we should instead keep the `source` in the base RelationshipV2.proto model. and then specific types of relationships will have 'destination' field only. Check: https://github.com/linkedin/datahub-gma/pull/456#discussion_r1815818392

What do you think? The issue with keeping "source" in the relationship model, which is part of the aspect model now is that the source could be different from the asset urn of the aspect.

validators/src/main/java/com/linkedin/metadata/validator/RelationshipValidator.java
```diff
@@ -43,23 +47,24 @@ public static void validateRelationshipSchema(@Nonnull RecordDataSchema schema)
...
+          className, DESTINATION_FIELD);
     }
 
-    ValidationUtils.fieldsUsingInvalidType(schema, ValidationUtils.PRIMITIVE_TYPES).forEach(field -> {
-      ValidationUtils.invalidSchema("Relationship '%s' contains a field '%s' that makes use of a disallowed type '%s'.",
-          className, field.getName(), field.getType().getType());
-    });
-
-    validatePairings(schema);
+    // Relationship V2 uses AuditStamp class, instead of primitive types
```
jsdonn, line 61: IIRC we also need to replace usages of com.linkedin.common.Time with int64. Does this PR cover that?

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -753,6 +789,29 @@ private static Class<? extends Urn> urnClassForField(@Nonnull Class<? extends Re
...
+          .getField(fieldName)
+          .getType()
+          .getProperties()
+          .get("java")).getString("class");
+    } else {
+      urnClassName =((DataMap)  ValidationUtils.getRecordSchema(recordClass)
+          .getField(fieldName)
+          .getType()
+          .getMembers()
+          .get(0)
```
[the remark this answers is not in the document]
hongliu2024: Removed this logic.

validators/src/main/java/com/linkedin/metadata/validator/RelationshipValidator.java
hongliu2024: I'm thinking to use [AuditStamp](https://jarvis.corp.linkedin.com/codesearch/result/?name=AuditStamp.proto&path=models%2Fmodels%2Fsrc%2Fmain%2FdataProto%2Fproto%2Fcom%2Flinkedin%2Fcommon&reponame=linkedin-multiproduct%2Fmodels) class to replace the AuditStampActor and AuditStampTimeAt

Example [TestRelationship.proto](https://jarvis.corp.linkedin.com/codesearch/result?name=TestRelationship.proto&path=metadata-models%2Fmetadata-models%2Fsrc%2Fmain%2FdataProto%2Fproto%2Fcom%2Flinkedin%2Fmetadata%2Frelationship&reponame=linkedin-multiproduct%2Fmetadata-models)

Please let me know wdyt

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
hongliu2024: as discussed offline, I removed the source field from the v2 model

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -315,7 +323,26 @@ public static <RELATIONSHIP extends RecordTemplate> Urn getSourceUrnFromRelation
...
+   *   source: optional union[
+   *     sourceDemoAsset: DemoAssetUrn
+   *   ]
+   */
+  public static <RELATIONSHIP extends RecordTemplate> String extractFieldNameFromUnionField(RELATIONSHIP relationship, String fieldName) {
+    final DataMap dataMap = RecordUtils.getRecordTemplateField(relationship, fieldName, DataMap.class);
+    if (dataMap == null) {
+      ValidationUtils.throwNullFieldException(fieldName);
+    }
+    return dataMap.keySet().iterator().next();
```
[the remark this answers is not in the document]
hongliu2024: this field is translated from proto `oneOf`, which should have only 1 item.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -290,10 +291,21 @@ public static <ENTITY extends RecordTemplate> Urn getUrnFromEntity(@Nonnull ENTI
...
+    boolean relationshipInV2 = isRelationshipInV2(relationship.getClass());
+    // validate relationship schema
+    RelationshipValidator.validateRelationshipSchema(relationship.getClass(), relationshipInV2);
+
+    // get Urn field from the relationship
+    final Urn urn;
+    if (!relationshipInV2) {
+      // if V1, get the urn from the given field name, "source" or "destination"
+      urn = RecordUtils.getRecordTemplateField(relationship, fieldName, urnClassForRelationship(relationship.getClass(), fieldName));
+    } else {
```
jsdonn, line 305: should we validate that fieldName = 'destination' in this case?

dao-api/src/main/java/com/linkedin/metadata/dao/utils/GraphUtils.java
```diff
@@ -25,6 +24,7 @@ public static void checkSameUrn(@Nonnull final List<? extends RecordTemplate> re
       return;
     }
 
+    // ToDo: how to handle this for Relationship V2?
```
jsdonn, line 27: > Should it just skip the check for source field for V2, or should it look at the belonging aspect to find the source Urn? I feel looking at belonging aspect can have problems too because one relationship can be included in multiple different aspects.

We can just check during runtime if the source urn in the relationship(s) match the asset urn being ingested. I will add it in my followup PR.

testing/test-models/src/main/pegasus/com/linkedin/testing/AssetWithRelationship.pdl
```diff
@@ -0,0 +1,18 @@
...
+record AssetWithRelationship {
+  /**
+   * For unit tests
+   */
+  urn: optional FooUrn
+
+  /**
+   * For unit tests
+   */
+  aspect_with_relationship: AspectWithRelationship
```
jsdonn, line 17: nit: use camelcase

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
```diff
@@ -306,7 +318,11 @@ private static <RELATIONSHIP extends RecordTemplate> Urn getUrnFromRelationship(
   @Nonnull
   public static <RELATIONSHIP extends RecordTemplate> Urn getSourceUrnFromRelationship(
```
ybz1013, line 319: EbeanLocalRelationshipWriterDAO.addRelationshipGroup invokes this for each relationship groups when create sql updates. 

Since here it will throw exceptions, how are the relationship updates are fired?

ybz1013 (commented):
One question

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
hongliu2024: updated.

testing/test-models/src/main/pegasus/com/linkedin/testing/AssetWithRelationship.pdl
hongliu2024: updated

dao-api/src/main/java/com/linkedin/metadata/dao/utils/ModelUtils.java
hongliu2024: That's a great question. I was trying to handle it but it needs to come from aspect. 
@jsdonn - I'd like to discuss about this.
For now I think it's ok to push this PR out since V1 isn't impacted by this.

jsdonn: As discussed offline, Hong will help add logic to propagate the source urn passed in during ingestion to the EbeanLocalRelationshipWriterDAO

jsdonn (approved):
LGTM thanks for doing this. Let's address the concerns around not including the source field in relationship 2.0 models in a fast-follow PR

PR #463, opened 2024-10-29
feat(relationship 2.0): use REMOVE_ALL_EDGES_FROM_SOURCE by default for ingesting 2.0 relationships

jsdonn:
## Summary
In Model 1.0, when new relationships are ingested, existing relationships are handled (i.e. removed) according to user-defined removal options. There are 4 possible options:
1) REMOVE_NONE
2) REMOVE_ALL_EDGES_FROM_SOURCE
3) REMOVE_ALL_EDGES_TO_DESTINATION
4) REMOVE_EDGES_FROM_SOURCE_TO_DESTINATION

1. has no use cases and will no longer be supported.
2. will be the default
3. will no longer be supported. in 2.0, relationships must always be modeled such that the asset they are derived from is the source in the relationship.
4. has use cases in AIM but moving forward, it will be up to the user to replicate these now "non-default" behaviors via pre-ingestion lambdas. Essentially, whatever the user provides to us in the aspect model, we will ingest using REMOVE_ALL_EDGES_FROM_SOURCE behavior.

## Testing Done
adjusted some unit tests
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -896,8 +896,13 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
       List<List<RELATIONSHIP>> allRelationships = EBeanDAOUtils.extractRelationshipsFromAspect(aspect);
       localRelationshipUpdates = allRelationships.stream()
           .filter(relationships -> !relationships.isEmpty()) // ensure at least 1 relationship in sublist to avoid index out of bounds
-          .map(relationships -> new BaseLocalRelationshipBuilder.LocalRelationshipUpdates(
-              relationships, relationships.get(0).getClass(), BaseGraphWriterDAO.RemovalOption.REMOVE_NONE))
+          .map(relationships -> {
+            for (RELATIONSHIP relationship : relationships) {
+              validateRelationshipSource(urn, relationship); // validate that all relationships' sources are equal to the given asset urn
+            }
+            return new BaseLocalRelationshipBuilder.LocalRelationshipUpdates(
+              relationships, relationships.get(0).getClass(), BaseGraphWriterDAO.RemovalOption.REMOVE_ALL_EDGES_FROM_SOURCE);
+          })
           .collect(Collectors.toList());
     } else if (_relationshipSource == RelationshipSource.RELATIONSHIP_BUILDERS) {
       if (_localRelationshipBuilderRegistry != null && _localRelationshipBuilderRegistry.isRegistered(aspectClass)) {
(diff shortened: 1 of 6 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -375,6 +376,21 @@ public static LocalRelationshipCriterion buildRelationshipFieldCriterion(LocalRe
...
 
+  /**
+   * Validate that the source of a relationship matches the asset urn that the relationship was derived from. Relationships
+   * must be defined such that the source is always the asset urn of the aspect the relationship is included in.
+   * @param assetUrn urn of the asset to be ingested
+   * @param relationship relationship derived from the aspect to be ingested
+   * @throws IllegalArgumentException if the asset urn and the relationship source urn don't match
+   */
+  public static <URN, RELATIONSHIP extends RecordTemplate> void validateRelationshipSource(URN assetUrn, RELATIONSHIP relationship) {
+    Urn sourceUrn = getSourceUrnFromRelationship(relationship);
```
yangyangv2, line 387: do we still require the relationship to define the source?

jsdonn: Removed. the source urn that will be used will be the asset urn during ingestion.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -896,8 +895,8 @@ public <ASPECT extends RecordTemplate, RELATIONSHIP extends RecordTemplate> List
       List<List<RELATIONSHIP>> allRelationships = EBeanDAOUtils.extractRelationshipsFromAspect(aspect);
       localRelationshipUpdates = allRelationships.stream()
           .filter(relationships -> !relationships.isEmpty()) // ensure at least 1 relationship in sublist to avoid index out of bounds
-          .map(relationships -> new BaseLocalRelationshipBuilder.LocalRelationshipUpdates(
-              relationships, relationships.get(0).getClass(), BaseGraphWriterDAO.RemovalOption.REMOVE_NONE))
+          .map(relationships -> new LocalRelationshipUpdates(
+            relationships, relationships.get(0).getClass(), BaseGraphWriterDAO.RemovalOption.REMOVE_ALL_EDGES_FROM_SOURCE))
```
hongliu2024, line 899: will this code have any impact to the existing V1 relationships, before they are fully migrated to V2?

jsdonn: This code is all in the if block
```
if (_relationshipSource == RelationshipSource.ASPECT_METADATA)
```
so it will only affect the v2 models. @ybz1013 is working to make a "dual-write" mode which will write both v1 and v2 relationships (into different tables).

jsdonn: He will also add a validation to make sure that the v1 relationships (derived from relationship builder) == v2 relationships (extracted from aspect) while in dual-write mode so that we can avoid data inconsistencies during the transition phase.

hongliu2024: Sounds good. Thanks for confirming

hongliu2024 (approved):
LGTM

PR #470, opened 2024-11-07
fix: make unit test use v2 + fix removeRelationships logic

ybz1013:
## Summary
I noticed the unit tests that were supposed to extract relationship from aspect were using V1 relationship model, which has the source urn in it unlike V2. So the logic tested was not really for relationship V2.

1. added BelongsToV2 model that doesn't have source urn and has union as destination (how v2 should be
2. changed aspect related unit tests to use BelongsToV2 
3. fixed removeRelationships logic to be able to handle V2 case by passing down the source urn

## Testing Done
./gradlew build

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
@@ -64,7 +64,8 @@ import com.linkedin.testing.FooSnapshot;
 import com.linkedin.testing.MixedRecord;
 import com.linkedin.testing.localrelationship.AspectFooBar;
 import com.linkedin.testing.localrelationship.BelongsTo;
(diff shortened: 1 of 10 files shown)


jsdonn (approved):
Thanks for proactively fixing this

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -3132,6 +3133,7 @@ public void testAddWithLocalRelationshipBuilder() throws URISyntaxException {
     assertEquals(aspects.size(), 1);
   }
 
+  // TODO: fix this
```
hongliu2024, line 3136: what to fix here? Can you please add some details?

ybz1013: forgot to remove

PR #472, opened 2024-11-08
Change the API we use to fill in the default value

ZihanLi58:
## Summary
Change the API we use to fill in the default value
The original way of fill in default value will fill in all the field with default value even they are not required. which means as long as the field is optional, we will fill in the default value for it. The  behavior is conflict with the way we do compare using DataTemplateUtil.areEqual(), which only fill in the default value for required field. We do compare using DataTemplateUtil.areEqual() to compare the existing value and new value to avoid duplicate write to database, so the original approach will make us to write to db for every request even it's essentially the same. 

In the new approach, we will use ValidateDataAgainstSchema.validate(model, new ValidationOptions(RequiredMode.FIXUP_ABSENT_WITH_DEFAULT, CoercionMode.NORMAL,
            UnrecognizedFieldMode.DISALLOW)); to do the default fill in, which is the same API they use in DataTemplateUtil.areEqual(). 
This approach will make sure 
1. We don't write the same data to db multiple times.
2. We only fill in default value when it's a required field, which save some space on db as well
3. We leverage the existing check, so no much overhead will be added there

Along with the change, I also remove _modelValidationOnWrite and force it to always do the default value fill in. I have verified that no one is calling enableModelValidationOnWrite, so the behavior should remain unchanged. 
## Testing Done
Unit test. 
Deploy to local and verify it works
## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -206,8 +206,6 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   private final Map<Class<? extends RecordTemplate>, EqualityTester<? extends RecordTemplate>>
       _aspectEqualityTesterMap = new ConcurrentHashMap<>();
 
-  private boolean _modelValidationOnWrite = true;
-
   // Always emit MAE on every update regardless if there's any actual change in value
   private boolean _alwaysEmitAuditEvent = false;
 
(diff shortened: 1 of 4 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1550,9 +1539,9 @@ public long newNumericId() {
   /**
    * Validates a model against its schema.
    */
-  protected void validateAgainstSchema(@Nonnull RecordTemplate model) {
+  public static void validateAgainstSchemaAndFillinDefault(@Nonnull RecordTemplate model) {
```
yangyangv2, line 1542: can we add a new test case to capture the change of this behavior

ZihanLi58: Since there is no e2e test, the unit test I changed actually mock the write behavior. And it test the default value fill in. Let me know if you think more test needed

yihzhou (approved):

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
yangyangv2: I'm thinking the unit test testToAndFromJson().

Can we make cases that it's serializing a record template with empty default and optional default, so future developers are clear with the expected behaviors

PR #475, opened 2024-11-13
cleanup: Add no-internalModels-flag variants to methods used by GMS's

jphui:
## Summary

Add no-flag variants to methods so that consumer GMS's don't need to call "lix logic", which is confusing since there is no Lix active any more.

Non-TMS Codesearch: https://jarvis.corp.linkedin.com/codesearch/results/?query=mp%3A*-gms%20getResourceLix

TMS Codesearch: https://jarvis.corp.linkedin.com/codesearch/results?query=mp%3Ametadata-store+getResourceLix&current=1&nresults=10

## Testing Done

n/a

## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
@@ -1076,11 +1076,44 @@ public abstract class BaseEntityResource<
   }
 
   /**
+   * Same as the above but doesn't require the callee to provide the 'isInternalModelsEnabled' flag.
+   *
+   * @param urns collection of urns
+   * @param aspectClasses set of aspect classes
+   * @return All {@link VALUE} objects keyed by {@link URN} obtained from DB
+   */
+  @Nonnull
+  protected Map<URN, VALUE> getInternal(@Nonnull Collection<URN> urns,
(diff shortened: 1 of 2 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/BaseEntityResource.java
```diff
@@ -1089,6 +1107,21 @@ protected Map<URN, VALUE> getInternalNonEmpty(@Nonnull Collection<URN> urns,
                 : toValue(newSnapshot(e.getKey(), e.getValue()))));
   }
 
+  /**
+   * Same as the above but doesn't require the callee to provide the 'isInternalModelsEnabled' flag.
+   */
+  @Nonnull
+  protected Map<URN, VALUE> getInternalNonEmpty(@Nonnull Collection<URN> urns,
+      @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses) {
+    final boolean isInternalModelsEnabled = getResourceLix().testGetAll(_urnClass.getSimpleName());
```
yangyangv2, line 1116: same here, unless only getAll is calling this method. But there's no restriction for future cases

yangyangv2 (commented):
what is the motivation on this clean up? 

If getResourceLix() is no longer used, better to cleanup in the original methods rather than introducing new ones.

PR #476, opened 2024-11-13
fix(relationship 2.0): handle the case where aspects can have multiple fields with the same relationship type

jsdonn:
## Summary
Currently, these is a "bug" where aspects that have multiple relationship fields of the same relationship type will not have all of them persisted properly in the DB. This is because the current logic for extracting relationships from aspect fields keeps all the extracted relationships separate by the field it was derived from. Then, the relationships are ingested sequentially by each field name group. Relationships in 2.0 are ingested using REMOVE_ALL_EDGES_FROM_SOURCE so any earlier relationships will get overwritten by later relationships if the later ones are of the same type as the earlier ones. 

The new logic will group them by relationship class. Each relationship class will be processed sequentially. This way, there is never any overlap of types between different relationship groups. So any deletions will only occur on previously existing relationships, not relationships added earlier in the current ingestion.
## Testing Done
Updated unit tests
## Checklist


yangyangv2 (approved):
LGTM, thanks for making the quick fix!

PR #477, opened 2024-11-13
fix: getAspectClassNames to handle union alias

ybz1013:
## Summary
Small fix of getAspectClassNames to handle union alias, to unblock EbeanLocalRelationshipQueryDAO.findEntities. (related to recent change in #436 

The buggy behavior was that when aliases exists in the union type, this function will return the alias instead of the canonical name.

For example:
```
typeref EntityUnionAlias = union[
  foo: EntityFoo,
  bar: EntityBar
]
```
old code will return foo when we actually expects `com.linkedin.testing.EntityFoo`

## Testing Done
added unit test
./graldew build

## Checklist


jsdonn (approved):
Thanks for proactively fixing the bug

PR #481, opened 2024-11-27
Fix silent failure when ingesting an Asset with an unrecognized field

jphui:
## Summary

BUG=[META-21242](https://jira01.corp.linkedin.com:8443/browse/META-21242)

This PR adds Restli model validation to the Asset ingestion endpoints to throw errors in these otherwise silent-failure cases: 

1. curli calls with fields that are simply not a part of the model, these should error out!
2. (java) client calls with a newer bump to metadata-models that contain NEW fields that are not recognized by the server, these should also error out!

## Testing Done

Added unit testing for both:

1. parent utility class
3. BaseEntityResource, which is where ingestion is invoked at the API layer

## Checklist


validators/src/main/java/com/linkedin/metadata/validator/ValidationUtils.java
```diff
@@ -209,4 +215,23 @@ private static DataSchema.Type getFieldOrArrayItemType(@Nonnull RecordDataSchema
...
+   * fields are all valid fields of the model.
+   *
+   * <p>This is copied from BaseLocalDAO, which uses this for Aspect-level ingestion validation. This is meant for
+   * Asset-level validation.</p>
+   *
+   * <p>Reference ticket: META-21242</p>
+   */
+  public static void validateAgainstSchema(@Nonnull RecordTemplate model) {
+    ValidationResult result = ValidateDataAgainstSchema.validate(model,
+        new ValidationOptions(RequiredMode.CAN_BE_ABSENT_IF_HAS_DEFAULT, CoercionMode.NORMAL,
```
yangyangv2, line 230: can ValidationOptions be a constant? or it has to be instantiated every time?

validators/src/main/java/com/linkedin/metadata/validator/ValidationUtils.java
```diff
@@ -209,4 +215,23 @@ private static DataSchema.Type getFieldOrArrayItemType(@Nonnull RecordDataSchema
...
+            UnrecognizedFieldMode.DISALLOW));
+    if (!result.isValid()) {
+      invalidSchema(result.getMessages().toString());
```
yangyangv2, line 233: using `String.valueOf(result.getMessages())` to avoid NPE?

yangyangv2 (commented):
Minor comments, thanks for making the improvements!

yangyangv2 (approved):
LGTM!

validators/src/main/java/com/linkedin/metadata/validator/ValidationUtils.java
```diff
@@ -46,6 +46,9 @@ public final class ValidationUtils {
 
   private static final Map<String, String> SOFT_DELETED_FIELD_METADATA = fieldMetadata(new SoftDeletedAspect().schema());
 
+  public static final ValidationOptions VALIDATION_OPTIONS =
```
yangyangv2, line 49: minor: make it private if not used in other class

PR #482, opened 2024-11-27
fix(relationship): fix bug in extractRelationshipFromAspect

jsdonn:
## Summary
There is a bug in metadata-store integration tests:
```
com.linkedin.metadata.dao.RestliClientStatusException: com.linkedin.restli.client.RestLiResponseException: Response status 500, serviceErrorMessage: java.lang.ClassCastException: class com.linkedin.common.EntityComplianceAnnotation cannot be cast to class java.util.List (com.linkedin.common.EntityComplianceAnnotation is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')

```

This is because RecordTemplates are first checked to see if they are relationships. If not, the previous logic will assume that they are lists. But this is not the case if they are other structs. The fix is to add a check to see if it is a list. If it's not, just return.
## Testing Done
update unit test
## Checklist


ybz1013 (approved):
Thanks for the fix

PR #484, opened 2024-12-09
fix(aim): make an exception for checkSameSourceUrn for AIM case

jsdonn:
## Summary
AIM integration tests are failing because of this validation check. They have a special case where the urn from which the relationship is derived (MlTrainedModel) is the destination of the Produces relationship. Allow this use case temporarily while AIM redesigns their models.
## Testing Done
mint build
## Checklist


ybz1013 (approved):
Might be better with some constants and sets for the entities if this gets more.

But for now, approving to unblock user.

samlovestech (approved):
LGTM, Thanks for the quick change.

PR #487, opened 2024-12-11
fix(backfill): ingest relationships + use createdon desc during entit…

jsdonn:
## Summary
job-gms uses (urn, aspect, version, createdon) as its primary key. The current backfillEntityTables logic reads a single row from metadata_aspect and inserts it into the entity table. This normally works since the primary key of all other metadata_aspect tables is (urn, aspect, version) so it will always only have 1 result. But with job-gms, there could be multiple version 0 rows so the current logic is indeterministic on which of the possible rows it returns. 

In this PR, update the query from the aspect table to order by createdon desc so that the most recently updated row is always returned in the case when there are multiple (urn, aspect, version) matches.

Also handle relationship ingestion (which was previously handled by _localAccess.add() but refactored into a separate method) during the entity table backfill. 
## Testing Done
update unit test for local relationships, local deployment of job-gms to confirm createdon desc query works.
## Checklist


rakhiagr (approved):
LGTM

PR #490, opened 2024-12-18
fix(new schema): fix duplicate key exception for aspect table

jsdonn:
## Summary
In new schema mode, the aspect table is not getting update with optimistic locking on the createdon column. As a result, the duplicate version 0 rows in job-gms are causing duplicate key exceptions even in new schema mode.

To fix, read the latest row from the aspect table, get the timestamp, and also perform an update using the old timestamp to match a single row in the aspect table.
## Testing Done
./gradlew build 
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -759,24 +759,18 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
   // metadata_aspect schema and we don't take this route here to keep this change backward compatible.
   private static final String OPTIMISTIC_LOCKING_UPDATE_SQL = "UPDATE metadata_aspect "
       + "SET urn = :urn, aspect = :aspect, version = :version, metadata = :metadata, createdOn = :createdOn, createdBy = :createdBy "
-      + "WHERE urn = :urn and aspect = :aspect and version = :version";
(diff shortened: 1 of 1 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -806,9 +800,11 @@ protected <ASPECT extends RecordTemplate> void updateWithOptimisticLocking(@Nonn
     final SqlUpdate oldSchemaSqlUpdate;
     if (_schemaConfig == SchemaConfig.NEW_SCHEMA_ONLY) {
       // In NEW_SCHEMA, the entity table is the SOT and getLatest (oldTimestamp) reads from the entity
-      // table. Therefore, we will apply compare-and-set with oldTimestamp on entity table (addWithOptimisticLocking)
-      // aspect table will apply regular update over (urn, aspect, version) primary key combination.
-      oldSchemaSqlUpdate = assembleOldSchemaSqlUpdate(aspect, null);
+      // table. Therefore, we will apply compare-and-set with oldTimestamp on entity table (addWithOptimisticLocking).
+      // We will also apply an optimistic locking update over (urn, aspect, version) primary key combination to avoid duplicate
+      // key exceptions when the primary key includes createdon.
+      EbeanMetadataAspect result = findLatestMetadataAspect(_server, urn, aspectClass);
```
rakhiagr, line 806: What if the urn and aspect does not exist? What does this return?

rakhiagr (approved):
Thanks for fixing this issue 🚢

PR #491, opened 2024-12-19
fix(backfill): update a_urn column during entity table backfill

jsdonn:
## Summary
Currently, backfillEntityTables API does not backfill the a_urn column. This is an oversight since we never expected to need to backfill this column and it cannot be backfilled like other aspect columns. Thus, the solution is to add it to the upsert SQL which is called in EbeanLocalAccess::add. 
## Testing Done
./gradlew build

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
@@ -58,6 +58,7 @@ public class SQLStatementUtils {
 
   // Note: include a_urn in the duplicate key update for backfilling the a_urn column in updateEntityTables() if the column was
   // added after data already exists in the entity table. Normally, a_urn never needs to get updated once it's set the first time.
+  // Note: this SQL should only be called if the urnExtraction flag is true (i.e. a_urn column exists)
   private static final String SQL_UPSERT_ASPECT_WITH_URN_TEMPLATE =
       "INSERT INTO %s (urn, a_urn, %s, lastmodifiedon, lastmodifiedby) VALUE (:urn, :a_urn, :metadata, :lastmodifiedon, :lastmodifiedby) "
           + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon, a_urn = :a_urn;";
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
@@ -56,9 +56,11 @@ public class SQLStatementUtils {
       "INSERT INTO %s (urn, %s, lastmodifiedon, lastmodifiedby) VALUE (:urn, :metadata, :lastmodifiedon, :lastmodifiedby) "
           + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon;";
 
+  // Note: include a_urn in the duplicate key update for backfilling the a_urn column in updateEntityTables() if the column was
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -56,9 +56,11 @@ public class SQLStatementUtils {
       "INSERT INTO %s (urn, %s, lastmodifiedon, lastmodifiedby) VALUE (:urn, :metadata, :lastmodifiedon, :lastmodifiedby) "
           + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon;";
 
+  // Note: include a_urn in the duplicate key update for backfilling the a_urn column in updateEntityTables() if the column was
+  // added after data already exists in the entity table. Normally, a_urn never needs to get updated once it's set the first time.
   private static final String SQL_UPSERT_ASPECT_WITH_URN_TEMPLATE =
       "INSERT INTO %s (urn, a_urn, %s, lastmodifiedon, lastmodifiedby) VALUE (:urn, :a_urn, :metadata, :lastmodifiedon, :lastmodifiedby) "
-          + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon;";
+          + "ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon, a_urn = :a_urn;";
```
yangyangv2, line 63: would this logic only be called when a_urn column presents? how to ensure this or can you clarify in the comment a_urn always exists when this is called and why?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
```diff
@@ -43,7 +43,7 @@ public void testCreateUpsertAspectSql() {
     String expectedSql =
         "INSERT INTO metadata_entity_foo (urn, a_urn, a_aspectfoo, lastmodifiedon, lastmodifiedby) VALUE (:urn, "
             + ":a_urn, :metadata, :lastmodifiedon, :lastmodifiedby) ON DUPLICATE KEY UPDATE a_aspectfoo = :metadata,"
-            + " lastmodifiedon = :lastmodifiedon;";
+            + " lastmodifiedon = :lastmodifiedon, a_urn = :a_urn;";
```
yangyangv2, line 46: do we have similar unit test that covers entity without `a_urn`?

jsdonn: It is right underneath. There is a flag passed into `createUpsertAspectSql()` which indicates if a_urn column exists or not. It is only controlled by the DAO code, not the user.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
jsdonn: I added a comment

jsdonn: Basically yes, there is a flag that is passed into all of the SQL statement methods that indicates whether the a_urn column exists. If it does, then use template 1, if not, use template 2

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
yangyangv2: sounds good, thanks for clarification.

yangyangv2 (approved):
LGTM

PR #496, opened 2025-02-07
(feat) Add deleteMany Dao method

jphui:
## Summary

In order to support the new [Delete Aspect API](https://docs.google.com/document/d/1kLnZNiWTxSsMIJP9sPT8Oc0gVGXau9JR5WkUy3ze0Wk/edit?tab=t.0), a `deleteMany()` method (with default-parameter variants) was added.

DeleteMany will be used to support the following spec:

```proto
/** Delete one or more Aspects for an URN */ 
rpc deleteAspects(DeleteAspectFromASSETRequest) returns (DeleteAspectFromASSETResponse) { option (.si.methodType) = DELETE; }

message DeleteAspectFromASSETRequest { 
/** Request context */ 
si.RequestContext context = 1; 
/** The URN of the asset for which aspects are being deleted. */ 
URN urn = 2; 
/** The list of aspect names to delete. */ 
repeated string aspectNames = 3; }
```

## Testing Done

Adjusted additional unit testing and added a new one.

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Docs related to the changes have been added/updated (if applicable)

PR #497, opened 2025-02-12
(feat) Add deleteWithReturn

jphui:
## Summary

This PR builds off of https://github.com/linkedin/datahub-gma/pull/496 by adding the ability for `deleteMany()` to return a Collection of the deleted Aspects' contents. This is necessary due to the finding [here](https://docs.google.com/document/d/1kLnZNiWTxSsMIJP9sPT8Oc0gVGXau9JR5WkUy3ze0Wk/edit?tab=t.0#bookmark=id.18k6qo9kpty6).
## Testing Done

added new unit testing :)

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -913,7 +913,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   /**
    * Deletes the latest version of an aspect for an entity and returns the ***old value***.
    *
-   * Note that if the aspect is already null or deleted, the return value will be null. Mechanistically, upon a normal
+   * <p>Note that if the aspect is already null or deleted, the return value will be null. Mechanistically, upon a normal
    * "LIVE" ingestion, the deletion operation is skipped altogether.
    */
   @Nullable
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -49,6 +49,7 @@ import com.linkedin.metadata.query.IndexGroupByCriterion;
 import com.linkedin.metadata.query.IndexSortCriterion;
 import java.sql.Timestamp;
 import java.time.Clock;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -659,7 +660,10 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
   private <ASPECT extends RecordTemplate> ASPECT_UNION unwrapAddResultToUnion(URN urn, AddResult<ASPECT> result,
       @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext) {
+    // handle post-update hooks and emit MAE + return the newValue
     ASPECT rawResult = unwrapAddResult(urn, result, auditStamp, trackingContext);
+
+    // package it into a union
     return ModelUtils.newEntityUnion(_aspectUnionClass, rawResult);
   }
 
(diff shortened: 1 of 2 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -817,53 +821,58 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
    * <p>The new aspect will have an automatically assigned version number, which is guaranteed to be positive and
    * monotonically increasing. Older versions of aspect will be purged automatically based on the retention setting.
    *
-   * <p>Note that we do not support Post-update hooks while soft deleting an aspect
+   * <p>Note that we do not currently support pre- or post- update hooks while soft deleting an aspect.
```
jphui, line 824: Updated this comment to be more accurate since pre-hooks are also not invoked during deletion.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -890,16 +899,30 @@ public <ASPECT extends RecordTemplate> void delete(@Nonnull URN urn, @Nonnull Cl
...
     checkValidAspect(aspectClass);
 
-    runInTransactionWithRetry(() -> {
+    final AddResult<ASPECT> result = runInTransactionWithRetry(() -> {
       final AspectEntry<ASPECT> latest = getLatest(urn, aspectClass, false);
       final IngestionParams ingestionParams = new IngestionParams().setIngestionMode(IngestionMode.LIVE);
       return addCommon(urn, latest, null, aspectClass, auditStamp, new DefaultEqualityTester<>(), trackingContext, ingestionParams);
     }, maxTransactionRetry);
 
+    return result.getOldValue();
```
yangyangv2, line 919: What if the oldValue is null? do we handle the null case?

If Null is handled, please mark the method return as @Nullable.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -2682,6 +2730,45 @@ public void testDeleteManyWithRelationshipRemoval() throws URISyntaxException {
     assertFalse(optionalAspect2.isPresent());
   }
 
+  @Test
+  public void testDeleteWithReturnOnNonexistentAsset() {
```
yangyangv2, line 2734: It's good you have unit test to cover the null case at deleteWithReturn level. How about the upper level method: deleteMany? we will expect to return a collection of null? or an empty collection?

yangyangv2 (commented):
few comments, clarify and ship.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: It can return `null`. I've updated the method signatures.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
jphui: In the next commit I'll add testing for the `deleteMany()` case here and just make clear what the expected behavior is (as-is).

jphui: **UPDATE**

An Aspect Union CANNOT have a 'null' as its member (ModelUtils / RecordUtils makes sure of this).

If we wanted "empty Aspect members" we could initiate this, but to me this is _**somewhat**_ potentially confusing since the implication here is that the "old value" of the deletion was an "empty aspect" as opposed to a _**NULL**_ aspect -- two separate concepts in some sense.

NULL makes more sense as a true-to-truth representation of the "old value" so I will leave it as null.

This means that if a deletion happens for an already-null or already-deleted Aspect, the returned Asset will have its member field set to NULL.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -890,16 +907,34 @@ public <ASPECT extends RecordTemplate> void delete(@Nonnull URN urn, @Nonnull Cl
...
   public <ASPECT extends RecordTemplate> void delete(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass,
       @Nonnull AuditStamp auditStamp, int maxTransactionRetry, @Nullable IngestionTrackingContext trackingContext) {
+    deleteWithReturn(urn, aspectClass, auditStamp, maxTransactionRetry, trackingContext);
+  }
+
+  /**
+   * Deletes the latest version of an aspect for an entity and returns the ***old value***.
+   *
+   * <p>Note that if the aspect is already null or deleted, the return value will be null. Mechanistically, upon a normal
+   * "LIVE" ingestion, the deletion operation is skipped altogether.
```
jphui, line 917: Note the documentation I added.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -817,53 +822,65 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
...
-   * <p>Note that we do not support Post-update hooks while soft deleting an aspect
+   * <p>Note that we do not currently support pre- or post- update hooks while soft deleting an aspect.
    *
    * @param urn urn the URN for the entity the aspects are attached to
    * @param aspectClasses Aspect Classes of the aspects being deleted, must be supported aspect types in {@code ASPECT_UNION}
    *                      Because Aspect Classes must be unique for a given Entity, we use a set to avoid duplicates.
    * @param auditStamp the audit stamp of this action
    * @param maxTransactionRetry maximum number of transaction retries before throwing an exception
+   * @return a collection of the deleted aspects (their value before deletion), each wrapped in an instance of {@link ASPECT_UNION}
+   *         If an aspect is already null or deleted, the return collection will not contain it whatsoever (it's "null").
```
jphui, line 833: More documentation.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: Both I believe. Both seem to be called in `aspectUpdateHelper()` and it is never called via the delete() pathway:

(1) Legacy hooks

https://github.com/linkedin/datahub-gma/blob/2de35c0187fd9545a3cf060d352a44ee78d32554/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java#L650

(2) Callback hooks 

https://github.com/linkedin/datahub-gma/blob/2de35c0187fd9545a3cf060d352a44ee78d32554/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java#L637

yangyangv2: You can keep the current implementation. Suggest to create a TODO with ticket to clarify this behavior in the callback documentation -- as an Aspect owner, if I request all the update to my asset to be notified through the callback, then what about DELETE (and CREATE as well)

NVM if it's already addressed.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
yangyangv2: NULL member field makes sense, thanks for digging into this.

just to make sure the comment and unit test cover this behavior.

thanks!

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
yangyangv2: do you have an unit test for this case?

jphui: Yes just added in the most recent commit :)

yangyangv2 (approved):
LGTM, thanks for addressing the comments and extended unit test cases!

PR #498, opened 2025-02-18
feat(assets): Add create API with insert to fail on duplicates

kotkar-pallavi:
## Summary

- Adding a new method to create new assets for [Create API](https://docs.google.com/document/d/1PpcxiFNHOLLTptxhP1l-TWzJ15cZXPiiZk04175lYm4/edit?usp=sharing).
- The current behavior is to create new assets as part of the upsert method. This new create method can be used to explicitly create assets. 
- Create method will throw an exception if we try to insert records with a duplicate primary key instead of doing an update.

## Testing Done
Local build and added unit tests

## Checklist

- [X] Links to related issues (if applicable)
- [X] Docs related to the changes have been added/updated (if applicable)

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -905,12 +905,10 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     aspectCreateLambdas.forEach(aspectCreateLambda -> checkValidAspect(aspectCreateLambda.getAspectClass()));
 
     // create aspects and process callbacks in a single transaction
-    final URN createdUrn = runInTransactionWithRetry(() -> {
+    return runInTransactionWithRetry(() -> {
       return createAspectsWithCallbacks(urn, aspectValues, aspectCreateLambdas, auditStamp, trackingContext);
       }, maxTransactionRetry
     );
-
-    return createdUrn;
   }
 
   /**
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -161,6 +161,28 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     }
   }
 
+  /**
+   * Immutable class to hold the details of a Create to an aspect.
+   *
+   * <p>This class allows the wildcard capture in {@link #create(Urn, List, AuditStamp, IngestionTrackingContext, IngestionParams)}</p>
+   *
+   * @param <ASPECT> the type of the aspect being updated
+   */
+  @AllArgsConstructor
+  @Value
+  public static class AspectCreateLambda<ASPECT extends RecordTemplate> {
+    @Nonnull
+    Class<ASPECT> aspectClass;
+
+    @Nonnull
+    IngestionParams ingestionParams;
+
+    public AspectCreateLambda(ASPECT value) {
+      this.aspectClass = (Class<ASPECT>) value.getClass();
+      ingestionParams = new IngestionParams().setIngestionMode(IngestionMode.LIVE);
+    }
+  }
+
   @Data
   @AllArgsConstructor
   protected static class AspectUpdateResult<ASPECT extends RecordTemplate> {
@@ -659,32 +681,49 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
         trackingContext, updateTuple.getIngestionParams());
   }
 
-  <ASPECT extends RecordTemplate> AddResult<ASPECT> createAspectWithCallbacks(@Nonnull URN urn,
-      @Nonnull ASPECT aspectValue, @Nonnull Class<ASPECT> aspectClass, @Nonnull AuditStamp auditStamp,
-      @Nullable IngestionTrackingContext trackingContext, @Nonnull IngestionParams ingestionParams) {
+  /**
+   * Adds a new version of an aspect for an entity and triggeers callbacks for all containing aspects if registered.
+   * @param urn the URN for the entity the aspect is attached to
+   * @param aspectValues the list new aspect value to be added to the asset being created
+   * @param aspectCreateLambdas the list of aspect create lambdas to be executed
+   * @param auditStamp the audit stamp for the operation
+   * @param trackingContext the tracking context for the operation
+   * @return the URN of the entity if the operation is successful, otherwise null
+   * @param <ASPECT_UNION> must be a valid aspect union type defined in com.linkedin.metadata.aspect
+   */
+  <ASPECT_UNION extends RecordTemplate> URN createAspectWithCallbacks(@Nonnull URN urn,
+      @Nonnull List<? extends RecordTemplate> aspectValues,
+      @Nonnull List<AspectCreateLambda<? extends RecordTemplate>> aspectCreateLambdas,
+      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext) {
 
-    AspectUpdateResult result = aspectCallbackHelper(urn, aspectValue, null, ingestionParams);
-    aspectValue = (ASPECT) result.getUpdatedAspect();
-    // skip the normal ingestion to the DAO
-    if (result.isSkipProcessing()) {
-      return null;
-    }
+    // process callbacks and check that all the aspects are valid
+    for (int i = 0; i < aspectValues.size(); i++) {
+      RecordTemplate aspectValue = aspectValues.get(i);
+      AspectCreateLambda<? extends RecordTemplate> createLambda = aspectCreateLambdas.get(i);
+      AspectUpdateResult result = aspectCallbackHelper(urn, aspectValue, null, createLambda.ingestionParams);
+      // skip the normal ingestion to the DAO
+      if (result.isSkipProcessing()) {
+        continue;
+      }
+      aspectValue = result.getUpdatedAspect();
 
-    checkValidAspect(aspectValue.getClass());
+      checkValidAspect(createLambda.aspectClass);
+      validateAgainstSchemaAndFillinDefault(aspectValue);
 
-    validateAgainstSchemaAndFillinDefault(aspectValue);
-
-    // Invoke pre-update hooks, if any
-    if (_aspectPreUpdateHooksMap.containsKey(aspectClass)) {
-      for (final BiConsumer<Urn, RecordTemplate> hook : _aspectPreUpdateHooksMap.get(aspectClass)) {
-        hook.accept(urn, aspectValue);
+      // Invoke pre-update hooks, if any
+      if (_aspectPreUpdateHooksMap.containsKey(createLambda.aspectClass)) {
+        for (final BiConsumer<Urn, RecordTemplate> hook : _aspectPreUpdateHooksMap.get(createLambda.aspectClass)) {
+          hook.accept(urn, aspectValue);
+        }
       }
     }
 
-    long numRows = createNewAspect(urn, aspectClass, aspectValue, auditStamp, trackingContext, ingestionParams.isTestMode());
-    log.debug("Created aspect {} for urn {} with {} rows", aspectClass.getSimpleName(), urn, numRows);
+    // check if any the aspects are in test mode
+    boolean isTestModeFalseForAll = aspectCreateLambdas.stream().filter(aspectCreateLambda -> aspectCreateLambda.getIngestionParams().isTestMode()).collect(
+          Collectors.toList()).isEmpty();
 
-    return new AddResult<>(null, aspectValue, aspectClass);
+    long numRows = createNewAspect(urn, aspectCreateLambdas, aspectValues, auditStamp, trackingContext, isTestModeFalseForAll);
+    return numRows > 0 ? urn : null;
   }
 
   private <ASPECT extends RecordTemplate> ASPECT_UNION unwrapAddResultToUnion(URN urn, AddResult<ASPECT> result,
@@ -844,19 +883,38 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
         : unwrapAddResult(urn, result, auditStamp, trackingContext));
   }
 
+  /**
+   * Add new aspects for an asset.
+   * @param urn the URN for the entity the aspects are attached to
+   * @param aspectValues the list of new aspect values to be added to the asset being created
+   * @param aspectCreateLambdas the list of aspect create lambdas to be executed
+   * @param auditStamp the audit stamp for the operation
+   * @param maxTransactionRetry the maximum number of times to retry the transaction
+   * @param trackingContext the tracking context for the operation
+   * @return the URN of the entity if the operation is successful, otherwise null
+   * @param <ASPECT> must be a supported aspect type in {@code ASPECT_UNION}
+   */
   @Nonnull
-  public <ASPECT extends RecordTemplate> ASPECT create(@Nonnull URN urn, @Nonnull ASPECT aspectValue,
-      @Nonnull Class<ASPECT> aspectClass, @Nonnull AuditStamp auditStamp, int maxTransactionRetry,
-      @Nullable IngestionTrackingContext trackingContext, @Nullable IngestionParams ingestionParams) {
+  public <ASPECT extends RecordTemplate> URN create(@Nonnull URN urn,
+      @Nonnull List<ASPECT> aspectValues,
+      @Nonnull List<AspectCreateLambda<? extends RecordTemplate>> aspectCreateLambdas,
+      @Nonnull AuditStamp auditStamp, int maxTransactionRetry,
+      @Nullable IngestionTrackingContext trackingContext) {
 
-    checkValidAspect(aspectClass);
+    // check that all the aspects are valid
+    aspectCreateLambdas.forEach(aspectCreateLambda -> checkValidAspect(aspectCreateLambda.getAspectClass()));
 
-    final AddResult<ASPECT> result = runInTransactionWithRetry(
-        () -> createAspectWithCallbacks(urn, aspectValue, aspectClass, auditStamp, trackingContext, ingestionParams),
-        maxTransactionRetry);
+    // create aspects and process callbacks in a single transaction
+    final List<AddResult<ASPECT>> results = runInTransactionWithRetry(() -> {
+      List<AddResult<ASPECT>> createdAspects = new ArrayList<>();
+      createAspectWithCallbacks(urn, aspectValues, aspectCreateLambdas, auditStamp, trackingContext);
+      return createdAspects;
+      }, maxTransactionRetry
+    );
 
-    // skip MAE producing and post update hook in test mode or if the result is null (no actual update with addCommon)
-    return ingestionParams.isTestMode() ? result.newValue : unwrapAddResult(urn, result, auditStamp, trackingContext);
+    results.stream().map(x -> unwrapAddResultToUnion(urn, x, auditStamp, trackingContext));
+
+    return urn;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


jphui (commented):
Solid PR, but a few high-level questions and potential gaps

1) `(Ebean)LocalAccess` is never directly invoked by the `metadata-graph-assets` service:

```java
    assetModelPegasusInterop.getPegasusAspectsFromAsset(asset).stream()
       .forEach(aspect -> dao.add((URN) pegasusUrn, aspect, auditStamp, pegasusTrackingContext, pegasusParams));
```

It only has access to `(Ebean)LocalDAO`, through which LocalAccess is invoked:

`EbeanLocalDAO.java`:

```java
  public <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
...
      ASPECT aspect = toRecordTemplate(aspectClass, result).orElse(null);
      _localAccess.add(urn, aspect, aspectClass, auditStamp, null, false);
...
```

I believe you will probably need to introduce a new set of methods inside `EbeanLocalDao.java` -- and / or perhaps `BaseLocalDao.java` -- in order to access the functionality here.

2) Design choice of adding a new method (done here in this PR) vs adding a flag for existing functionality

This is more of a question than a comment, but is there a reason that CREATE functionality is being added as a new method here as opposed to just adding a `boolean flag` or something to other existing `add()` methods?

At a cursory glance it _seems_ like other than the SQL statement used, the logic flow is basically identical.

Maybe you've thought through future augmentations / feature-adds that are needed and it needs to be separate or something similar?

3) Ingestion callback requirements?

Do you know if Ingestion Callback will be needed for CREATE? If you aren't sure what this is, @ZihanLi58 / I / Yihong can explain the concept -- and Zihan can probably elaborate on the need -- but basically this is something supported by Normal Ingestion at the moment and is implemented at the DAO (not Access) level.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -152,6 +152,59 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
...
+        .setParameter("lastmodifiedby", actor);
+
+    // If a non-default UrnPathExtractor is provided, the user MUST specify in their schema generation scripts
+    // 'ALTER TABLE <table> ADD COLUMN a_urn JSON'.
+    if (urnExtraction) {
+      sqlUpdate.setParameter("a_urn", toJsonString(urn));
+    }
+
+    // newValue is null if aspect is to be soft-deleted.
+    if (aspectValue == null) {
```
ZihanLi58, line 189: why do we support create null? use create to api to delete one aspect?

ZihanLi58 (commented):
Can you share more requirement about the create API? basically
is it for createAsset or createAspect?
If createAsset, then I think you might need to accept more aspects in one request?
If createAspect, then we probably only want to fail the request when the urn and the aspect are all existing? If urn exist but aspect does not, we should still create it for user?

ZihanLi58:
> Solid PR, but a few high-level questions and potential gaps
> 
> 1. `(Ebean)LocalAccess` is never directly invoked by the `metadata-graph-assets` service:
> 
> ```java
>     assetModelPegasusInterop.getPegasusAspectsFromAsset(asset).stream()
>        .forEach(aspect -> dao.add((URN) pegasusUrn, aspect, auditStamp, pegasusTrackingContext, pegasusParams));
> ```
> 
> It only has access to `(Ebean)LocalDAO`, through which LocalAccess is invoked:
> 
> `EbeanLocalDAO.java`:
> 
> ```java
>   public <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn, @Nonnull Class<ASPECT> aspectClass) {
> ...
>       ASPECT aspect = toRecordTemplate(aspectClass, result).orElse(null);
>       _localAccess.add(urn, aspect, aspectClass, auditStamp, null, false);
> ...
> ```
> 
> I believe you will probably need to introduce a new set of methods inside `EbeanLocalDao.java` -- and / or perhaps `BaseLocalDao.java` -- in order to access the functionality here.
> 
> 2. Design choice of adding a new method (done here in this PR) vs adding a flag for existing functionality
> 
> This is more of a question than a comment, but is there a reason that CREATE functionality is being added as a new method here as opposed to just adding a `boolean flag` or something to other existing `add()` methods?
> 
> At a cursory glance it _seems_ like other than the SQL statement used, the logic flow is basically identical.
> 
> Maybe you've thought through future augmentations / feature-adds that are needed and it needs to be separate or something similar?
> 
> 3. Ingestion callback requirements?
> 
> Do you know if Ingestion Callback will be needed for CREATE? If you aren't sure what this is, @ZihanLi58 / I / Yihong can explain the concept -- and Zihan can probably elaborate on the need -- but basically this is something supported by Normal Ingestion at the moment and is implemented at the DAO (not Access) level.

Based on the question I ask before, you can choose different design choice there. And to make sure the ingestion behavior is consistency, call back is required for create as well

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
kotkar-pallavi: removed the check. It is not applicable on create.

kotkar-pallavi:
@jphui 
1. I have added the method definition and iml in BaseLocalDao.java. Thanks for pointing out. 
2. The add method a lot of params not required for create. e.g.: the lambda function is not application on create functionality. I noticed the add method also contains a lot of legacy implementation leading to a very deep call stack. In order to keep the new impl clean and extendable for future improvements I decided to create a separate create method.
3. Added ingestion callbacks. 

@ZihanLi58 
The create API will be used to explicitly create a new asset. This is avoid the overloaded use of upsert method. We want to keep the interface similar to the existing upsert api, hence allowing create for single aspect instead of a list. Right now, upsert is done iteratively for each aspect. 
The create method will throw error if a URN already exists. it does not matter if it contains any assets or not. The user is expected to use the update api to add more aspects and these exceptions will be defined in the asset service. 
I added callbacks to create method and exposing it from the BaseLocalDao.java class. 

please review. thanks!

ZihanLi58:
@kotkar-pallavi Current update api actually accept asset as input, and it does the upsert for each aspect Iteratively. In your current approach, can the create user facing api take asset as input? If the asset contains multiple aspect, it seems will fail the request right as second create call for second aspect will fail since the first one already create the urn.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -659,6 +659,34 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> aspectUpdateHelper(URN
         trackingContext, updateTuple.getIngestionParams());
   }
 
+  <ASPECT extends RecordTemplate> AddResult<ASPECT> createAspectWithCallbacks(@Nonnull URN urn,
```
ZihanLi58, line 662: Can you add java doc for each of the new method you added?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -816,6 +844,21 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
...
+      @Nullable IngestionTrackingContext trackingContext, @Nullable IngestionParams ingestionParams) {
+
+    checkValidAspect(aspectClass);
+
+    final AddResult<ASPECT> result = runInTransactionWithRetry(
+        () -> createAspectWithCallbacks(urn, aspectValue, aspectClass, auditStamp, trackingContext, ingestionParams),
+        maxTransactionRetry);
+
+    // skip MAE producing and post update hook in test mode or if the result is null (no actual update with addCommon)
+    return ingestionParams.isTestMode() ? result.newValue : unwrapAddResult(urn, result, auditStamp, trackingContext);
```
jphui, line 859: IIUC, `CREATE` will return the NEW value of the Asset after it (successfully) runs?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -411,4 +412,26 @@ public void testCheckColumnExists() {
...
+  public void testCreateDuplicateAspect() {
+    FooUrn fooUrn = makeFooUrn(102);
+    AspectFoo aspectFoo = new AspectFoo().setValue("foo");
+    AuditStamp auditStamp = makeAuditStamp("actor", _now);
+    _ebeanLocalAccessFoo.create(fooUrn, aspectFoo, AspectFoo.class, auditStamp, null, false);
+    try {
+      _ebeanLocalAccessFoo.create(fooUrn, aspectFoo, AspectFoo.class, auditStamp, null, false);
+    } catch (DuplicateKeyException duplicateKeyException) {
+      assert (duplicateKeyException.getMessage().contains("Duplicate entry"));
+    }
```
jphui, line 435: Nice testing 😎

jphui (commented):
Other than Zihan's comment, lgtm from me

kotkar-pallavi:
> @kotkar-pallavi Current update api actually accept asset as input, and it does the upsert for each aspect Iteratively. In your current approach, can the create user facing api take asset as input? If the asset contains multiple aspect, it seems will fail the request right as second create call for second aspect will fail since the first one already create the urn.

@ZihanLi58  updated to create aspects with multiple aspects. it will be done in a single create statement that will throw exception if urn already exists. as discussed offline, doing a exists check followed by iterative create might lead to race condition. please re-review. thanks!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -63,6 +63,7 @@
 import javax.persistence.OptimisticLockException;
 import javax.persistence.RollbackException;
 import javax.persistence.Table;
+import lombok.NonNull;
```
yangyangv2, line 66: why using Lombok? instead of javax.annotation.Nonnull?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -152,6 +153,75 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
       return sqlUpdate.setParameter("metadata", metadata).execute();
   }
 
+  /**
+   * Create aspect from entity table.
```
yangyangv2, line 157: can include a comment in this method to include 

1.  conditions when this method is called: (a: all aspects have been pre-processed, b: ...)
2. does it guarantee no-partial success? 
3.  what happens then racing condition occurs

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -152,6 +153,75 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
...
+          .setCreatedfor(impersonator, SetMode.IGNORE_NULL);
+      if (ingestionTrackingContext != null) {
+        auditedAspect.setEmitTime(ingestionTrackingContext.getEmitTime(), SetMode.IGNORE_NULL);
+        auditedAspect.setEmitter(ingestionTrackingContext.getEmitter(), SetMode.IGNORE_NULL);
+      }
+      auditedAspects.add(toJsonString(auditedAspect));
+    }
+
+    String metadata = String.join(",", auditedAspects);
+    return sqlUpdate.setParameter("metadata", metadata).execute();
```
yangyangv2, line 222: Is the intention to replace :metadata with values (separated with comma) from multiple columns?

```
INSERT INTO table(urn, a_urn, column1, column2) VALUE (:urn, :a_urn, :metadata)
```

and

```
update.setParameter("metadata", "a, b");
```

I suspect if this is allowed, can you double check? and add an unit test for it

kotkar-pallavi: yes `:metadata` will have comma separated values. it is difficult to test and I tried adding unit test here: https://github.com/linkedin/datahub-gma/pull/498/commits/a59095e1b32de56fc8cdf707b3abc2d0145c1c6e

from debugging with breakpoint, i verified the values and they look as follows:

```
 {"lastmodifiedon":"2025-02-23 23:13:29.767","aspect":{"value":"foo"},"lastmodifiedby":"urn:li:testActor:actor","canonicalName":"com.linkedin.testing.AspectFoo"},{"lastmodifiedon":"2025-02-23 23:13:29.767","aspect":{"value":"bar"},"lastmodifiedby":"urn:li:testActor:actor","canonicalName":"com.linkedin.testing.AspectBar"}
```

It has comma separated value for aspect foo and bar.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
kotkar-pallavi: updated javadoc.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java
```diff
@@ -78,6 +78,15 @@ public static String toJsonString(@Nonnull RecordTemplate recordTemplate) {
     }
   }
 
+  /**
+   * Serializes a list of {@link RecordTemplate} to JSON string.
+   * @param recordTemplates the list of record templates to serialize
+   * @return the JSON string serialized using {@link JacksonDataTemplateCodec}.
+   */
+  public static String toJsonString(@Nonnull List<RecordTemplate> recordTemplates) {
```
ZihanLi58, line 86: Where do you use this?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -659,6 +681,51 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> aspectUpdateHelper(URN
...
+   * Adds a new version of an aspect for an entity and triggeers callbacks for all containing aspects if registered.
+   * @param urn the URN for the entity the aspect is attached to
+   * @param aspectValues the list new aspect value to be added to the asset being created
+   * @param aspectCreateLambdas the list of aspect create lambdas to be executed
+   * @param auditStamp the audit stamp for the operation
+   * @param trackingContext the tracking context for the operation
+   * @return the URN of the entity if the operation is successful, otherwise null
+   * @param <ASPECT_UNION> must be a valid aspect union type defined in com.linkedin.metadata.aspect
+   */
+  <ASPECT_UNION extends RecordTemplate> URN createAspectWithCallbacks(@Nonnull URN urn,
```
ZihanLi58, line 694: createAspectsWithCallbacks?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -161,6 +161,28 @@ public static class AspectUpdateLambda<ASPECT extends RecordTemplate> {
...
+  /**
+   * Immutable class to hold the details of a Create to an aspect.
+   *
+   * <p>This class allows the wildcard capture in {@link #create(Urn, List, AuditStamp, IngestionTrackingContext, IngestionParams)}</p>
+   *
+   * @param <ASPECT> the type of the aspect being updated
+   */
+  @AllArgsConstructor
+  @Value
+  public static class AspectCreateLambda<ASPECT extends RecordTemplate> {
```
ZihanLi58, line 173: why do we need this? do we expect different aspects may have different ingestionParams?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -816,6 +883,40 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
...
+
+    // create aspects and process callbacks in a single transaction
+    final List<AddResult<ASPECT>> results = runInTransactionWithRetry(() -> {
+      List<AddResult<ASPECT>> createdAspects = new ArrayList<>();
+      createAspectWithCallbacks(urn, aspectValues, aspectCreateLambdas, auditStamp, trackingContext);
+      return createdAspects;
+      }, maxTransactionRetry
+    );
+
+    results.stream().map(x -> unwrapAddResultToUnion(urn, x, auditStamp, trackingContext));
```
ZihanLi58, line 915: why this is needed?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -816,6 +883,40 @@ public <ASPECT extends RecordTemplate> ASPECT add(@Nonnull URN urn, AspectUpdate
...
+      @Nonnull List<AspectCreateLambda<? extends RecordTemplate>> aspectCreateLambdas,
+      @Nonnull AuditStamp auditStamp, int maxTransactionRetry,
+      @Nullable IngestionTrackingContext trackingContext) {
+
+    // check that all the aspects are valid
+    aspectCreateLambdas.forEach(aspectCreateLambda -> checkValidAspect(aspectCreateLambda.getAspectClass()));
+
+    // create aspects and process callbacks in a single transaction
+    final List<AddResult<ASPECT>> results = runInTransactionWithRetry(() -> {
+      List<AddResult<ASPECT>> createdAspects = new ArrayList<>();
```
ZihanLi58, line 909: seems always empty?

ZihanLi58 (commented):
Logic looks good to me, but seems a lot useless code, can you do a clean up?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kotkar-pallavi: to allow a list of aspects to be passed to create function. without this only one specific aspect can be given as an argument.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kotkar-pallavi: to process any post-update hooks and trigger mae events.

kotkar-pallavi: cleaned up code.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kotkar-pallavi: did refactor.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kotkar-pallavi: refactored method name as suggested.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/RecordUtils.java
kotkar-pallavi: removed code that is no longer used.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -152,6 +153,98 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
...
+          .setAspect(RecordUtils.toJsonString(aspectValues.get(i)))
+          .setCanonicalName(aspectCreateLambdas.get(i).getAspectClass().getCanonicalName())
+          .setLastmodifiedby(actor)
+          .setLastmodifiedon(new Timestamp(timestamp).toString())
+          .setCreatedfor(impersonator, SetMode.IGNORE_NULL);
+      if (ingestionTrackingContext != null) {
+        auditedAspect.setEmitTime(ingestionTrackingContext.getEmitTime(), SetMode.IGNORE_NULL);
+        auditedAspect.setEmitter(ingestionTrackingContext.getEmitter(), SetMode.IGNORE_NULL);
+      }
+      insertSqlValues.append(encloseInSingleQuotes(toJsonString(auditedAspect)));
```
yangyangv2, line 223: instead of using custom quoting, how about using variable substitutions in the insert statement and let the Beans to determine the quoting logic? 

e.g. your insertSqlValues is 

```
insert into table (urn, lastmodifiedtime, ... aspect1, aspect2, ...)
values ('urn:li:', time, .... :aspect1, :aspect2, ... )

and in the sqlUpdate

```
sqlUpdate.setParameter('aspect1', aspect1);
sqlUpdate.setParameter('aspect2', aspect2);

```
....

yangyangv2 (commented):
One comment, it's a much cleaner implementation, thanks for making this improvement!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
kotkar-pallavi: done!

yangyangv2 (approved):
LGTM, thanks for addressing the comments!

PR #504, opened 2025-03-06
[feat] add findRelationshipsV3 API (impl + unit tests)

ybz1013:
## Summary
To support Relationship 2.0 models, which doesnt have source in the relationship itself, we need a new API to return the source in a wrapper class along with the relationship2.0 model.

The internal use is AssetRelationship = source + relationship2.0 (destination only). I didn't want to hardcode the class here, so I left it extendable. User can pass it any RecordTemplate as wrapper class, but the implementation is only for AssetRelationship now.

I added a wrapOptions map in case we want to have other wrapper types in the future. Currently, it acts as a guard to prevent people who doesn't know this API to accidentally use it.

## Testing Done
./gradlew build
unit tests with relationship1.0 and 2.0

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -40,6 +40,8 @@ public class EbeanLocalRelationshipQueryDAO {
   public static final String RELATED_TO = "relatedTo";
   public static final String SOURCE = "source";
   public static final String METADATA = "metadata";
+  public static final String ASSET_RELATIONSHIP_TYPE = "asset.relationship.type";
+  public static final String MG_INTERNAL_ASSET_RELATIONSHIP_TYPE = "AssetRelationship.proto";
   private final EbeanServer _server;
   private final MultiHopsTraversalSqlGenerator _sqlGenerator;
 
@@ -249,7 +251,7 @@ public class EbeanLocalRelationshipQueryDAO {
    * @return A list of relationship records in SqlRow (col: source, destination, metadata, etc).
    */
   @Nonnull
-  public <RELATIONSHIP extends RecordTemplate> List<SqlRow> findRelationshipsV2V3Core(
+  private <RELATIONSHIP extends RecordTemplate> List<SqlRow> findRelationshipsV2V3Core(
       @Nullable String sourceEntityType, @Nullable LocalRelationshipFilter sourceEntityFilter,
       @Nullable String destinationEntityType, @Nullable LocalRelationshipFilter destinationEntityFilter,
       @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter,
@@ -296,8 +298,9 @@ public class EbeanLocalRelationshipQueryDAO {
       @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter,
       @Nonnull Class<ASSET_RELATIONSHIP> assetRelationshipClass, @Nullable Map<String, Object> wrapOptions,
       int offset, int count) {
-    if (wrapOptions == null) {
-      throw new IllegalArgumentException("Please check your use of the findRelationshipsV3 method. wrapOptions cannot be null.");
+    if (wrapOptions == null || !wrapOptions.containsKey(ASSET_RELATIONSHIP_TYPE)
+        || !MG_INTERNAL_ASSET_RELATIONSHIP_TYPE.equals(wrapOptions.get(ASSET_RELATIONSHIP_TYPE))) {
+      throw new IllegalArgumentException("Please check your use of the findRelationshipsV3 method.");
     }
 
     List<SqlRow> sqlRows = findRelationshipsV2V3Core(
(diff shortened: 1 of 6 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -222,6 +226,34 @@ public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationship
...
+   * @param destinationEntityType type of destination entity to query (e.g. "dataset")
+   * @param destinationEntityFilter the filter to apply to the destination entity when querying (not applicable to non-MG entities)
+   * @param relationshipType the type of relationship to query
+   * @param relationshipFilter the filter to apply to relationship when querying
+   * @param offset the offset query should start at. Ignored if set to a negative value.
+   * @param count the maximum number of entities to return. Ignored if set to a non-positive value.
+   * @return A list of relationship records in SqlRow (col: source, destination, metadata, etc).
+   */
+  @Nonnull
+  public <RELATIONSHIP extends RecordTemplate> List<SqlRow> findRelationshipsV2V3Core(
```
jsdonn, line 252: nit: does this need to be public?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -237,11 +269,86 @@ public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationship
...
+   */
+  @Nonnull
+  public <ASSET_RELATIONSHIP extends RecordTemplate, RELATIONSHIP extends RecordTemplate> List<ASSET_RELATIONSHIP> findRelationshipsV3(
+      @Nullable String sourceEntityType, @Nullable LocalRelationshipFilter sourceEntityFilter,
+      @Nullable String destinationEntityType, @Nullable LocalRelationshipFilter destinationEntityFilter,
+      @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter,
+      @Nonnull Class<ASSET_RELATIONSHIP> assetRelationshipClass, @Nullable Map<String, Object> wrapOptions,
+      int offset, int count) {
+    if (wrapOptions == null) {
+      throw new IllegalArgumentException("Please check your use of the findRelationshipsV3 method. wrapOptions cannot be null.");
```
jsdonn, line 300: If it can't be null why is it `@Nullable`?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -237,11 +269,86 @@ public <RELATIONSHIP extends RecordTemplate> List<RELATIONSHIP> findRelationship
...
+   * @param metadata the metadata string which can be parsed into a relationship
+   * @param sourceUrn the source urn
+   * @param wrapOptions options to wrap the relationship. Currently unused. Leaving it open for the future.
+   * @return A wrapped relationship record.
+   */
+  @Nonnull
+  private <ASSET_RELATIONSHIP extends RecordTemplate, RELATIONSHIP extends RecordTemplate> ASSET_RELATIONSHIP createAssetRelationshipWrapperForRelationship(
+      @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull Class<ASSET_RELATIONSHIP> assetRelationshipClass,
+      @Nonnull String metadata, @Nonnull String sourceUrn, @Nullable Map<String, Object> wrapOptions) {
+    // TODO: if other type of ASSET_RELATIONSHIP is needed, we need to distinguish it with wrapOptions and handles differently.
```
jsdonn, line 328: What do you mean by this? What "other type of ASSET_RELATIONSHIP" can there be? Can you share an example of a wrapOption?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: Let me make a change here to better guard this API.

This is more of a way to alert (potential) user if they dont know what this API does. If someone thinks they can pass in any thing as a wrapper and get a result, I would rather throw them an error here instead of letting it fail randomly later.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: my main concern is we cant keep making new versions of the same API, so adding a map for params can make it more flexible if we want to pass in something new later.

jsdonn: Ah that makes sense.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -36,6 +37,11 @@
  */
 @Slf4j
 public class EbeanLocalRelationshipQueryDAO {
+  public static final String RELATED_TO = "relatedTo";
+  public static final String SOURCE = "source";
+  public static final String METADATA = "metadata";
+  public static final String ASSET_RELATIONSHIP_TYPE = "asset.relationship.type";
```
jsdonn, line 43: what if we just call this something like `relationship.return.type`? make it more clear

jsdonn (approved):
LGTM

PR #513, opened 2025-03-27
(feat) Add MAE Emission for Aspect Deletion

jphui:
## Summary

This PR adds MAE emission for deletion operations.

The main implementation methodology here is that a new variant of the `unwrapAddResult()` method is added with a `isDeletion` flag, which is called when a Delete is performed. Slightly different logic is then acted on.

In addition, this PR changes `@Nonnull newValue` --> `@Nullable newValue` in basically all locations that previous claim it as being Non-Null. This is done because:

1. This conceptually is aligned with the concept of a Delete operation, in which the "new value" is simply NULL.
2. We cannot keep it @Nonnull since that would imply attempting to instantiate an "empty" version of the Aspect class...
    1. We cannot instantiate ASPECT with an empty value; it goes against Java's nature of "type erasure" when compiling for generified types -- thus, passing the class is necessary.

Lastly, this PR adds `Class<ASPECT> aspectClass` as an argument for the MAE emission methods. This is necessary because the current way that the Aspect Class is deduced is from Reflection -- `newValue.getClass()` -- which is not a permissible operation if `newValue` is allowed to be `null`.

* Dependency on `getClass()`: https://jarvis.corp.linkedin.com/codesearch/result/?name=KafkaMetadataEventProducerUtil.java&path=metadata-models%2Fmetadata-dao-impl%2Fkafka-producer%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fmetadata%2Fdao%2Fproducer&reponame=linkedin-multiproduct%2Fmetadata-models#39
* It's interesting to note that directly in the DAO code, it is NOT marked as @Nonnull, which is a clear bug risk that magically has not been run into yet lol...

Note that alternatively, we could aggregate `{ oldValue, newValue, aspectClass }` into `BaseLocalDAO`'s `AddResult` class for a more extensible code architecture, but that would require changing its signature to `public`, and as of _right now_, the benefit isn't immediately huge.

## Testing Done

unit testing updated

## Checklist

## Next Steps

1. Make corresponding changes in `metadata-models` in order to satisfy the interface changes here
2. (ideally once testing is complete) Make a follow-up PR for backfill support (this is officially a P1 right now)

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -743,6 +743,11 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
   private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, AddResult<ASPECT> result, @Nonnull AuditStamp auditStamp,
       @Nullable IngestionTrackingContext trackingContext) {
+    return unwrapAddResult(urn, result, auditStamp, trackingContext, false);
+  }
+
+  private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, @Nonnull AddResult<ASPECT> result, @Nonnull AuditStamp auditStamp,
+      @Nullable IngestionTrackingContext trackingContext, boolean isDeletion) {
     if (trackingContext != null) {
       trackingContext.setBackfill(false); // reset backfill since MAE won't be a backfill event
     }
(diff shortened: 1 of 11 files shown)


ZihanLi58 (commented):
This PR only cover the live traffic? and we are going to hav e separate one for backfill? 
Also is there a following up PR to make sure the MAE does reflect whether it's a delete / update operation?

jphui:
@ZihanLi58 

> This PR only cover the live traffic? and we are going to hav e separate one for backfill?

I figured backfill should be in a separate PR, but I can also include it here. In any case, you're right I should call it out explicitly as a `Next Step` if not here. wdyt?

> Also is there a following up PR to make sure the MAE does reflect whether it's a delete / update operation?

IIUC, this needs to be implemented in the `metadata-models` level under `KafkaMetadataEventProducer`, listed as a `Next Step`.

ZihanLi58 (approved):
LGTM

PR #515, opened 2025-04-01
feat(relationship ingestion): add multiple relationships in a single statement

jsdonn:
## Summary
Relationship insertion performance can be potentially improved by combining multiple single-valued INSERT INTO statements into a single INSERT INTO statement with multiple values. Do that in batches of 1000.

Ref to the MySQL documentation on optimizing INSERT statement performance: https://dev.mysql.com/doc/refman/8.4/en/insert-optimization.html#:~:text=If%20you%20are%20inserting%20many,separate%20single%2Drow%20INSERT%20statements.

## Testing Done
make sure existing relationship unit tests pass. add a new unit test ingesting a single relationship to make sure that case is covered as well.
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
@@ -170,7 +170,7 @@ public class EbeanLocalRelationshipWriterDAO extends BaseGraphWriterDAO {
    */
   private <ASPECT extends RecordTemplate, RELATIONSHIP extends RecordTemplate> void addRelationshipGroup(@Nullable Urn urn,
       @Nonnull Class<ASPECT> aspectClass, @Nonnull final List<RELATIONSHIP> relationshipGroup, boolean isTestMode) {
-    if (relationshipGroup.size() == 0) {
+    if (relationshipGroup.isEmpty()) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -220,45 +233,45 @@ private <ASPECT extends RecordTemplate> void removeRelationshipsBySource(@Nonnul
...
-          Thread.currentThread().interrupt(); // Restore interrupted status
-          throw new RuntimeException("Batch deletion interrupted", e);
-        } catch (Exception e) {
-          log.error("Error while executing batch deletion after {} batches", batchCount, e);
-          throw new RuntimeException("Batch deletion failed", e);
+    batchCount = 0;
+    while (batchCount < MAX_BATCHES) {
+      try {
+        // Use the runInTransactionWithRetry method to handle retries in case of transaction failures
+        int rowsAffected = runInTransactionWithRetry(deletionSQL::execute, 3); // Retry up to 3 times in case of transient failures
```
yangyangv2, line 252: is removeRelationshipsBySource already in a transaction at line 71? Generally it's not a good practice to have nested transactions, some DB may not even support it.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipWriterDAOTest.java
```diff
@@ -318,6 +321,44 @@ public void testRemoveRelationshipsDifferentAspect() throws URISyntaxException {
     _server.execute(Ebean.createSqlUpdate("truncate metadata_relationship_pairswith"));
   }
 
+  @Test
+  public void testAddRelationships() throws URISyntaxException, ReflectiveOperationException {
```
yangyangv2, line 331: Thanks for adding these unit tests!! Can we add one more parallel update to cover the current updates?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
```diff
@@ -43,7 +43,8 @@ private static class CommonColumnName {
     private static final String LAST_MODIFIED_ON = "lastmodifiedon";
     private static final String LAST_MODIFIED_BY = "lastmodifiedby";
   }
-  private static final int BATCH_SIZE = 10000; // Process rows in batches of 10,000
+  private static final int INSERT_BATCH_SIZE = 1000;
```
yangyangv2, line 46: May need to double click whether 1000 is a good number for multi-color (EGG) database. Can consult with MySql team? (and perhaps for the batch delete as well)

jsdonn: asked mysql team about this.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
jsdonn: This code is from before (@rakhiagr ). I just fixed the indentation. I think it's in a transaction in the sense that the whole addCommon is in a transaction right?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipWriterDAOTest.java
jsdonn: What do you mean parallel update covering the current updates?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
jsdonn: 1000 is fine: https://linkedin.atlassian.net/browse/MI-18169

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipWriterDAOTest.java
yangyangv2: yes, e.g. two threads are updating multiple relationships on the same assets. We'll expect the ending states should be consistent with either of these updates, but not mixed ones.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
yangyangv2: can you double check if it's in a nested transition? if so, can you leave a comment on here, and explain some expected result, for example if the inner transaction succeed, but failed at somewhere which caused the outer transaction to rollback, will it rollback the changes made by the inner transaction (though its been committed already)

jsdonn: Added this comment:
```
        // Although this whole logic is executed within a runInTransactionWithRetry block already (in addCommon), the nested
        // behavior is supported when using Ebean Transactions despite the fact that true nested transactions are not supported in MySQL.
        // Ebean mimics the nested behavior by using MySQL savepoints under the hood which CAN be nested. Thus, if all the inner commits (below) succeed,
        // but the outer commit (somewhere else outside this logic) does not, the WHOLE TRANSACTION (including the inner commits) will be rolled back.
```

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipWriterDAOTest.java
jsdonn: Added a unit test for concurrent relationship ingestions

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipWriterDAO.java
yangyangv2: Thanks for the confirmation!

PR #519, opened 2025-04-10
(Improvement) Adding Force Index for graph queries filtering on destination field of relationship tables

rakhiagr:
## Summary

We discovered that GQS queries which filters on the destination field of relationship tables were not performing as expected. They were using the wrong index(idx_destination), which was causing performance issues. So, we will force the use of the correct index`idx_destination_deleted_ts` by adding a FORCE INDEX hint to the queries.

 I checked for source filtering, it's using the index efficiently - don't want to over-engineer it right now

## Testing Done
./gradlew build

## Next Steps
- Ensure all the relationship table has the index `idx_destination_deleted_ts`

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -46,8 +46,8 @@ public class EbeanLocalRelationshipQueryDAO {
   public static final String RELATIONSHIP_RETURN_TYPE = "relationship.return.type";
   public static final String MG_INTERNAL_ASSET_RELATIONSHIP_TYPE = "AssetRelationship.proto";
   private static final int FILTER_BATCH_SIZE = 200;
-  private static final String FORCE_IDX_ON_SOURCE = " FORCE INDEX (idx_source_deleted_ts) ";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_comp_dest_del_ts) ";
+  private static final String DESTINATION_FIELD =  "destination";
   private final EbeanServer _server;
   private final MultiHopsTraversalSqlGenerator _sqlGenerator;
 
@@ -611,7 +611,7 @@ public class EbeanLocalRelationshipQueryDAO {
       } else if (!relationshipFilter.getCriteria().isEmpty()) {
         // Check if any relationship-level filter is on "destination"
         for (LocalRelationshipCriterion criterion : relationshipFilter.getCriteria()) {
-          if ("destination".equals(criterion.getField())) {
+          if (DESTINATION_FIELD.equals(criterion.getField())) {
             sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
             break;
           }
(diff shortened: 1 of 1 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -46,6 +46,8 @@ public class EbeanLocalRelationshipQueryDAO {
   public static final String RELATIONSHIP_RETURN_TYPE = "relationship.return.type";
   public static final String MG_INTERNAL_ASSET_RELATIONSHIP_TYPE = "AssetRelationship.proto";
   private static final int FILTER_BATCH_SIZE = 200;
+  private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_comp_dest_del_ts) ";
```
jsdonn, line 49: what happens if this index does not exist (we forgot to set it up for some relationship table)?

Also can we make sure to update the relationship table SMO sql generation script to include this index creation?

rakhiagr: > what happens if this index does not exist (we forgot to set it up for some relationship table)?

It will throw error - `ERROR 1176 (42000): Key 'idx_comp_dest_del_ts' doesn't exist in table 'rt'` - 

 - One way was checking that the index exists in the table before forcing it to use - but that will add latency for each query

 - another way maybe to check for all relationship tables initially and cache the result 

Or we could just add it in SMO sql generation script to ensure every new table has it, and for existing ones, I will add before merging the PR

jsdonn: I like the second option to check once on bootup and then caching the result. Just an additional safeguard that is relatively easy to add that could save us big headache in the future

yangyangv2: Do we need anything like `FORCE_IDX_ON_SOURCE`?

yangyangv2: the relationship creation should be automated going forward and secured by SMO

how about save the effort with try catch and standardize existing relationships --- when we're still in the time there're not too many?

rakhiagr: > Do we need anything like FORCE_IDX_ON_SOURCE?

@yangyangv2 Currently, the database is performing well (automatically picking the efficient index) for source filtering so I am thinking not to force index on source yet. We are not seeing any bottleneck on source filtering yet.

rakhiagr: Thinking to go with Yang right now.

1. Add a try-catch block
2. Update SMO sql generation script
3. Add a TODO for backlog to add extra safeguard by checking the existence of indexes.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -46,6 +48,8 @@ public class EbeanLocalRelationshipQueryDAO {
...
```
jsdonn, line 51: change `idx_comp_dest_del_ts` to `idx_destination_deleted_ts`

rakhiagr: Oops - updated!

jsdonn (approved):
LGTM!

PR #526, opened 2025-04-21
(feat) Adding delete asset implementation

kotkar-pallavi:
## Summary
- Adding new method for Delete Asset: `deleteAll` for hard delete (deleting urn and all aspects from db table)
- Added `ChangeType_DELETE_ALL` to differentiate between asset vs aspect deletion. This will be used in sending MAE events. In case of asset delete the ChangeType will be set to DELETE_ALL. 
- Added new abstract method `produceAspectSpecificMetadataAuditEvent` to `BaseTrackingMetadatProducer` and method `produceAspectSpecificMetadataAuditEvent` to interface `GenericMetadataProducer`. With these methods, both the `oldValue` and `newValue` are `@Nullable`. The methods contain additional `ChangeType` param. So eventually, these methods can be used to emit MAEs on all db operations . for example

|Action| Null Values | ChangeType |
|------|-------|--------------|
|create|oldValue is null|ChangeType.UPSERT|
|update| - | ChangeType.UPSERT|
|delete aspect | newValue is null | ChangeType.DELETE|
|delete asset | newValue is null | ChangeType.DELETE_ALL|

- Adding new helper method `deleteCommon`. Using the existing `addCommon` does not much sense for hard delete, since instead of updating column value to a constant `gma_deleted` we are deleting the entire row from table. 
- Creating new class `DeleteResult` to hold the result of delete, as the new value after delete will always be null. This can probably be re-used in aspect delete. 

References:
[RFC - Delete Asset](https://docs.google.com/document/d/1VdjyNbMrDnoMvgK2A9I_slndcvbPfhh0vq8NkaYLRDc/edit?usp=sharing)

## Testing Done
- Added unit tests
- Local Build

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -38,6 +38,7 @@ import com.linkedin.metadata.dao.tracking.BaseTrackingManager;
 import com.linkedin.metadata.dao.tracking.TrackingUtils;
 import com.linkedin.metadata.dao.urnpath.UrnPathExtractor;
 import com.linkedin.metadata.dao.utils.ModelUtils;
+import com.linkedin.metadata.events.ChangeType;
 import com.linkedin.metadata.events.IngestionMode;
 import com.linkedin.metadata.events.IngestionTrackingContext;
 import com.linkedin.metadata.internal.IngestionParams;
@@ -126,6 +127,12 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   static class AddResult<ASPECT extends RecordTemplate> {
     ASPECT oldValue;
     ASPECT newValue;
+    Class<ASPECT> klass;
+  }
+
+  @Value
+  static class DeleteResult<ASPECT extends RecordTemplate> {
+    ASPECT oldValue;
     Class<ASPECT> klass;
   }
 
@@ -795,6 +802,25 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     return isDeletion ? oldValue : newValue;
   }
 
+  private <ASPECT extends RecordTemplate> ASPECT unwrapDeleteResult(Urn urn, @Nonnull DeleteResult<ASPECT> result,
+      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext trackingContext, ChangeType changeType) {
+
+    final ASPECT oldValue = result.getOldValue();
+
+    if (_emitAspectSpecificAuditEvent) {
+      // For delete operation, the new value is always null
+      if (_trackingProducer != null) {
+        _trackingProducer.produceAspectSpecificMetadataAuditEvent(urn, oldValue, null, result.getKlass(), auditStamp,
+            trackingContext, IngestionMode.LIVE, changeType);
+      } else {
+        _producer.produceAspectSpecificMetadataAuditEvent(urn, oldValue, null, result.getKlass(), auditStamp,
+            IngestionMode.LIVE, changeType);
+      }
+    }
+
+    return oldValue;
+  }
+
   /**
    * Adds a new version of aspect for an entity.
    *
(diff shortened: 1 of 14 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -264,6 +266,17 @@ public static <ASPECT extends RecordTemplate> String createAspectUpsertSql(@Nonn
...
+  /**
+   * Create Delete SQL statement.
+   * @param urn entity urn
+   * @param isTestMode whether the test mode is enabled or not
+   * @return delete sql
+   */
+  public static <ASPECT extends RecordTemplate> String createDeleteSql(@Nonnull Urn urn, boolean isTestMode) {
+    final String tableName = isTestMode ? getTestTableName(urn) : getTableName(urn);
+    return String.format(SQL_DELETE_ASSET_WITH_URN, tableName);
+  }
```
jphui, line 278: If there is an existing SQLStatementUtils file, can you add a unit test for this?

Ofc with a Delete operation just want to be super cautious and make sure the lowest level stuff is working properly, explicitly 😎

dao-api/src/main/java/com/linkedin/metadata/dao/producer/BaseMetadataEventProducer.java
```diff
@@ -67,6 +68,21 @@ public abstract <ASPECT extends RecordTemplate> void produceAspectSpecificMetada
...
+   * Produces an aspect specific Metadata Audit Event (MAE) after a metadata aspect is updated for an entity with Change type.
+   * @param urn {@link Urn} of the entity
+   * @param oldValue the value prior to the update, or null if there's none.
+   * @param newValue the value after the update
+   * @param aspectClass the class of ASPECT
+   * @param auditStamp {@link AuditStamp} containing version auditing information for the metadata change
+   * @param ingestionMode {@link IngestionMode} of the change
+   * @param changeType {@link ChangeType} of the change
+   * @param <ASPECT> must be a supported aspect type in {@code ASPECT_UNION}
+   */
```
jphui, line 81: Can you add a TODO here as well and link this PR: https://github.com/linkedin-multiproduct/metadata-models/pull/1553, just mention that this can be a part of the refactoring work to make Delete Aspect use this pathway.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1185,6 +1327,10 @@ protected abstract <ASPECT_UNION extends RecordTemplate> int createNewAspect(@No
       @Nonnull List<? extends RecordTemplate> aspectValues, @Nonnull AuditStamp newAuditStamp,
       @Nullable IngestionTrackingContext trackingContext, boolean isTestMode);
 
+  protected abstract Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>> cleanUp(@Nonnull URN urn,
+      @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses, @Nullable AuditStamp auditStamp,
+      int maxTransactionRetry, @Nullable IngestionTrackingContext trackingContext, boolean isTestMode);
+
```
jphui, line 1333: I'm a bit lost on what this method does, would you be able to explain / rename it to be more descriptive 😓

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/AspectCallbackRoutingClient.java
```diff
@@ -47,6 +47,19 @@ default AspectCallbackResponse<ASPECT> routeAspectCallback(Urn urn, ASPECT newAs
...
+   * A method that routes the delete request to the appropriate custom API.
+   * @param urn the urn of the asset
+   * @param existingAspectValue the existing aspect value
+   * @param ingestionParams the ingestionParams of current update
+   * @param auditStamp the auditStamp of current request
+   * @return AspectCallbackResponse containing the updated aspect
+   */
+  default AspectCallbackResponse<ASPECT> routeAspectCallback(Urn urn, ASPECT existingAspectValue,
+      IngestionParams ingestionParams, AuditStamp auditStamp) {
+    return null;
```
jphui, line 60: If this is a no-op, maybe at a TODO here?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -925,6 +951,122 @@ public <ASPECT extends RecordTemplate> URN create(@Nonnull URN urn,
...
+  protected Collection<ASPECT_UNION> deleteCommon(@Nonnull URN urn,
+      @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses,
+      @Nonnull AuditStamp auditStamp, int maxTransactionRetry,
+      @Nullable IngestionTrackingContext trackingContext,
+      @Nullable IngestionParams ingestionParams,
+      boolean deleteAll) {
+
+    // TODO: Handle pre-deletion callbacks if any
+
+    // If deleteAll is true, delete entire asset, else mark aspects as deleted iteratively in a transaction
```
jphui, line 978: Curious where is the operation happening where you wrap this in a transaction?

Usually this is wrapped in a `runInTransactionWithRetry()` call?

jphui (commented):
Awesome table in the PR description, it's visually easy to read and understand 😎 

Thanks for also explaining the addition of `deleteCommon` ... the tidbit about "addCommon doesn't make sense since we're removing the row entirely" actually addresses the comment I had yesterday. This is a super helpful statement to have.

jphui:
Relevant comment: https://docs.google.com/document/d/1VdjyNbMrDnoMvgK2A9I_slndcvbPfhh0vq8NkaYLRDc/edit?tab=t.0#bookmark=id.vaxyi2pgsczi

Is the idea to process each `DELETE_ALL` MAE and send a `galeneDocumentDelete` for each? Curious if this was confirmed with HS that this is OK -- it doesn't trigger any errors, etc?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
kotkar-pallavi: added a unit test in `SQLStatementUtilsTest.java`

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
kotkar-pallavi: renamed the method to `permanentDelete` and added javadoc. thanks for pointing out. 
this method will call EBeanLocalAccess.deleteAll to permanently delete the row from the table and execute the sql.

dao-api/src/main/java/com/linkedin/metadata/dao/ingestion/AspectCallbackRoutingClient.java
kotkar-pallavi: the callbacks are not required for this iteration. i removed this definition for now. we can come back to this later.

kotkar-pallavi:
> Relevant comment: https://docs.google.com/document/d/1VdjyNbMrDnoMvgK2A9I_slndcvbPfhh0vq8NkaYLRDc/edit?tab=t.0#bookmark=id.vaxyi2pgsczi
> 
> Is the idea to process each `DELETE_ALL` MAE and send a `galeneDocumentDeleted` for each? Curious if this was confirmed with HS that this is OK -- it doesn't trigger any errors, etc?

great question! yes and yes! 
Since the MAEs are emitted for each aspect, we will continue to use the same model for emitting mae for each aspect. but each MAE will have the `galeneDocumentDeleted=true` 
Confirmed with the HS team that it is ok to emit multiple "_duplicate_" events . ref: https://linkedin-randd.slack.com/archives/CC7784WQ6/p1743454093535589

dao-api/src/main/java/com/linkedin/metadata/dao/producer/BaseMetadataEventProducer.java
kotkar-pallavi: I will make this change next in metadata-models classes that implement this class. it is required for delete asset work too.

jphui (approved):
🚢

PR #528, opened 2025-04-27
Check if aspect is present before processing delete result

kotkar-pallavi:
## Summary
- TL;DR: Check if the aspect value is present before using `.get` and processing the DeleteResult.
- Error founf in API testing

<details>

<summary>Log: Line984 </summary>

```
Caused by: java.util.NoSuchElementException: No value present                                                                                                                     
at java.base/java.util.Optional.get(Optional.java:143)                                                                                                                              
at com.linkedin.metadata.dao.BaseLocalDAO.lambda$deleteCommon$11(BaseLocalDAO.java:984)                                                                                             
at java.base/java.util.HashMap.forEach(HashMap.java:1421)                                                                                                                       
at com.linkedin.metadata.dao.BaseLocalDAO.deleteCommon(
```

</details>

- Explanation:
    - The input from user to delete API for deleteAll is an empty list of aspects. 
    - The deletion is done using the URN, but the MAEs are emitted for every aspect. So..
    - The service retrieves all the aspect class names that are part of this aspect and provides the list to the DAO level delete API - in order to process the MAEs for all aspects that will be deleted. 
    - Since the aspect class names are retrieved from the model, at the db level, the row to be deleted may or may not have all the aspect columns populated. 
    - Do a `.isPresentCheck` before trying to retrieve and process the aspect value. 
    - Only for the aspects that are present, create the DeleteResult object and populate the old value.
    - Use this old value to emit the MAE. 

example: Deletion for DemoAsset.
Scenario: For DemoAsset, 
- only status is added. and then deleteAll is called. 
- The DemoAsset contains other aspects like `DemoInfo`. Thus, the aspectClasses input list will contain the DemoInfo.class too, but the column does not contain any value. 
- Without the `isPresent()` check `.get` will return a null at 
```
DeleteResult deleteResult = new DeleteResult(value.get(), key);
```

## Testing Done
- Unit test added
- Local build

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
+++ b/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
@@ -474,6 +474,39 @@ public class EbeanLocalDAOTest {
   }
 
   @Test
+  public void testPermanentDeleteWithDeletedAspects() {
+    // DELETE ALL is not supported in the old schema.
+    if (_schemaConfig == SchemaConfig.NEW_SCHEMA_ONLY) {
+      // First add a record to db
+      EbeanLocalDAO<EntityAspectUnion, FooUrn> dao = createDao(FooUrn.class);
+      FooUrn urn = makeFooUrn(1);
+      AspectFoo foo = new AspectFoo().setValue("foo");
+      IngestionParams ingestionParams = new IngestionParams().setTestMode(false);
+      dao.setAlwaysEmitAuditEvent(false);
+      dao.setAlwaysEmitAspectSpecificAuditEvent(false);
+      dao.add(urn, foo, _dummyAuditStamp, null, ingestionParams);
+      // Verify the record was added: Foo aspect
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalDAOTest.java
```diff
@@ -440,6 +440,39 @@ public void testPermanentDelete() {
     }
   }
 
+  @Test
+  public void testPermanentDeleteWithNullAspects() {
```
jphui, line 444: Is there an existing test case that tests running `deleteAll()` on a deleted aspect case?

Basically, run delete() on 1 aspect, then run deleteAll.

This is another case where normally we'd return "null" so it kinda falls under the scope of changes made here.

kotkar-pallavi: since we are using optional with isPresent check now, that case should be handled too. 
i added a test case to verify `testPermanentDeleteWithDeletedAspects`

ZihanLi58 (approved):
LGTM once address the comments

PR #533, opened 2025-05-08
fix: skip ingestion when user callback service return updated aspect as NULL

yihzhou:
## Summary
We allow NULL to be produced by user callback service so it indicates the aspect shouldn't be ingested into DAO.
Note: The Type cast on NULL won't through exception.
## Testing Done
N/A
## Checklist


ZihanLi58 (approved):
+1

PR #534, opened 2025-05-13
Add write support for Soft Delete Aspect

jphui:
## Summary

This PR adds WRITE pathway functionality for Aspect soft-deletion, which is a necessary step in order to support Search Backfill -- https://docs.google.com/document/d/1_Hok1RaawgBN1ESGe3D0Qcesa6sBtbHuKjMLhWNgbtM/edit?tab=t.0#bookmark=id.wcf1pnap4gv.

There are a few important Design Choices that have been made here, most should have associated notes in the code:

## Testing Done

unit testing is updated :)

## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -103,6 +103,19 @@ public <ASPECT extends RecordTemplate> int add(@Nonnull URN urn, @Nullable ASPEC
     return addWithOptimisticLocking(urn, newValue, aspectClass, auditStamp, null, ingestionTrackingContext, isTestMode);
   }
 
+  @Override
+  @Transactional
+  public <ASPECT extends RecordTemplate> int addWithOldValue(
```
ZihanLi58, line 108: Why do we need to have this new method?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -112,6 +125,18 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
       @Nullable Timestamp oldTimestamp,
       @Nullable IngestionTrackingContext ingestionTrackingContext,
       boolean isTestMode) {
+    return addWithOptimisticLocking(urn, null, newValue, aspectClass, auditStamp, oldTimestamp, ingestionTrackingContext, isTestMode);
+  }
+
+  private <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
```
ZihanLi58: Also change this name to be updateWithOptimisticLocking. The current name makes me feel even it's an add, you will update the db with new value and old Value.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -905,11 +905,34 @@ protected <ASPECT extends RecordTemplate> void updateWithOptimisticLocking(@Nonn
...
+   * @param urn                 entity urn
+   * @param oldValue            old value of the aspect
+   * @param newValue            new value of the aspect
+   * @param aspectClass         aspect class
+   * @param auditStamp          audit stamp
+   * @param version             version of the record
+   * @param trackingContext     tracking context
+   * @param isTestMode          test mode
+   */
+  protected <ASPECT extends RecordTemplate> void insertWithOldValue(@Nonnull URN urn, @Nullable RecordTemplate oldValue,
```
ZihanLi58: an we just create another insert method with different parameters, feel like different name does not make too much sense here

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -277,15 +263,37 @@ private static <ASPECT extends RecordTemplate> EbeanMetadataAspect readSqlRow(Sq
...
+   * Same as {@link #isSoftDeletedAspect(SqlRow, String)}, but for {@link EbeanMetadataAspect}.
+   */
+  public static boolean isSoftDeletedAspect(@Nonnull EbeanMetadataAspect aspect) {
+    return isSoftDeletedAspect(aspect.getMetadata());
+  }
+
+  private static boolean isSoftDeletedAspect(@Nonnull String metadata) {
     try {
-      SoftDeletedAspect aspect = RecordUtils.toRecordTemplate(SoftDeletedAspect.class, sqlRow.getString(columnName));
-      return aspect.hasGma_deleted() && aspect.isGma_deleted();
```
ZihanLi58, line 288: Why do we need this change? old check does not work?

ZihanLi58: In another word, I'm asking why not use isGma_deleted to check instead of checkin against schema?

ZihanLi58 (commented):
I see it only contains the write api change, does read support this new format?  If not, I believe we need to add those in the same PR, otherwise services depends on this version will be break when reading the deleted aspect

jphui:
@ZihanLi58 

> I see it only contains the write api change, does read support this new format? If not, I believe we need to add those in the same PR, otherwise services depends on this version will be break when reading the deleted aspect

Read "supports" the format in the sense that it will recognize the deleted aspects correctly but will not be able to read any of the new data yet. Existing unit tests are proof / example of this.

Effectively, this means no one can consume this new data yet. I did this on purpose to:

1. ensure a safe change
2. break up the PR

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jphui: As mentioned in the Public DAO Api changes:

> `+ addWithOldValue() -- public helper is needed since the DAO is only aware of the public interface and NEEDS to be able to call functionality that can take advantage of an OldValue for writing`

Otherwise, there will be an additional get() that's performed "below" this level in order to access this data.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
jphui: This is explained the **isSoftDeletedAspect() changes** section:

```
 / *
   * Currently, there are no cases where the aspect is soft-deleted but this flag is NOT set to
   * true: the very existence of this Aspect type indicates soft-deletion. This is to say that
   * any aspect that is not soft-deleted will simply not exist -- cannot be cast -- as this type.
   * /
```

and also in the nearby documentation:

```
   * <p>This validation approach is necessary because many usages of checking soft-deletion are followed by
   * a deserialization call to {@link RecordUtils#toRecordTemplate(Class, String)}, which will fail if
   * we try to deserialize a SoftDeletedAspect -- to another Aspect Type -- with the flag set to "false".
```

kotkar-pallavi: i see your point from the explanation, but some confusion may arise later on why we are relying on the schema validation to check the state of aspect. 
adding a comment, something like "If this is does not fir the model of a deleted aspect, it is not not deleted" might help. just a suggestion..

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -136,9 +161,18 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
...
     }
 
-    // newValue is null if aspect is to be soft-deleted.
+    // newValue is null if aspect is to be (soft-)deleted.
     if (newValue == null) {
-      return sqlUpdate.setParameter("metadata", DELETED_VALUE).execute();
+      if (oldValue == null) {
+        // Shouldn't happen: shouldn't run delete() on an already-null/deleted aspect, or if calling to soft-delete, should
+        // pass in the old value.
+        // Since Old Schema logic that calls this method does NOT have access to the old value (without significant refactoring)
```
kotkar-pallavi, line 169: would it be more helpful to throw an exception if it is NEW_SCHEMA or DUAL_SCHEMA and oldValue=null?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -136,9 +161,18 @@ public <ASPECT extends RecordTemplate> int addWithOptimisticLocking(
...
+        // we make the decision to log a warning and proceed with the soft-deletion with an empty "deletedContent" value.
+        log.warn(String.format(
+            "OldValue should not be null when NewValue is null (soft-deletion). Urn: <%s> and aspect: <%s>", urn, aspectClass));
+      }
+      return sqlUpdate.setParameter("metadata", RecordUtils.toJsonString(
```
kotkar-pallavi, line 174: curious if we have verified with deleting more than 1 aspect in unit tests.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -423,6 +431,31 @@ public static LocalRelationshipCriterion buildRelationshipFieldCriterion(LocalRe
...
+   * @param aspectClass the class of the aspect
+   * @param oldValue the old value of the aspect
+   * @param deletedTimestamp the timestamp of the old value
+   * @param deletedBy the user who deleted the aspect
+   * @return a SoftDeletedAspect with the given old value and timestamp
+   */
+  @Nonnull
+  public static <ASPECT extends RecordTemplate> SoftDeletedAspect createSoftDeletedAspect(
+      @Nonnull Class<ASPECT> aspectClass, @Nullable ASPECT oldValue, @Nonnull Timestamp deletedTimestamp, @Nonnull String deletedBy) {
+    SoftDeletedAspect sda = new SoftDeletedAspect().setGma_deleted(true)
```
kotkar-pallavi: [nit][nit] can we rename this variable to something like `softDeletedAspect`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
jphui: Thanks both for the input, I will include my validation steps as well but still ultimately keep the original check. Note a few things:

1. the fact of the matter is still that in the case where this boolean is set to "false", there will be some DAO errors that will result from this due to the inability to parse the returned object...
2. ... however, this is OK because right now there are no cases that set it to "false"

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jphui: This would basically result in DAO failures for cases on the Old or Dual schemas, I think a warning log here should be sufficient for now.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
jphui: Yes, the `testDeleteManyWithRelationshipRemoval()` test has this:

```java
...
    // soft delete the AspectFooBar and AspectFooBaz aspects
    Collection<EntityAspectUnion> deletedAspects =
        fooDao.deleteMany(fooUrn, new HashSet<>(Arrays.asList(AspectFooBar.class, AspectFooBaz.class)), _dummyAuditStamp);
...
```

kotkar-pallavi (approved):

PR #537, opened 2025-05-19
(feat) Add SchemaValidatorUtil to Enable Safe Decoupling from DB Schema Evolution

rakhiagr:
## Summary

This PR introduces SchemaValidatorUtil, an improved and reusable utility for validating column and index existence at runtime. While column existence checks already existed in our codebase, they were not refreshed once initialized — which could lead to stale reads if a Pretzel patch was applied mid-runtime.

The new utility addresses this by adding a refresh-on-miss mechanism and supporting index checks, enabling the service to operate safely even when schema changes are delayed or applied asynchronously.

### Scenario 1: New Aspect Introduced, but Column Not Yet Added

Let’s consider a case where we introduce a new aspect DemoAspect for the Demo asset. We deploy the model changes to Corp, but the corresponding schema changes (e.g., adding the a_demoaspect column) are not yet deployed.

#### Read Behavior (Before Schema Change)

1. Fetching DemoAspect: This will return an empty result, as the corresponding column does not exist in the table.

3. Batch get all aspects for a Demo asset: This will work partially — the missing aspect (DemoAspect) is silently skipped, and other available aspects are returned.

4. Other read requests for unrelated aspects/assets: These continue to function as expected.

#### Write Behavior (Before Schema Change)

1. Single aspect write (DemoAspect): This will fail — the DAO will attempt to insert into a non-existent column, resulting in a SQL error.

2. Multi-aspect write: We don’t currently support multi-aspect writes. If introduced in the future, we’ll need logic to filter out missing columns to avoid partial failure.

3. Writes to other aspects: These will work normally since their corresponding schema is already present.

### After Schema Change Is Applied (But Without Service Restart)

Once the schema migration is applied (i.e., the column a_demoaspect is added), the following happens:
- No service restart is needed.
- The SchemaValidatorUtil uses a Caffeine cache with:
            10-minute TTL for column existence.
            Automatic refresh of cache entries upon expiration or first miss.

✅ Read, write operations involving DemoAspect will begin working once the cache refreshes.

### Scenario 2: New Filter API Deployed, but Index column Not Yet Added

Let’s assume we deploy a new version of our service that supports filtering on a new path (e.g. /status.removed in Status Aspect). But the corresponding index column (e.g., i_statust$removed) has not yet been added to the table.

#### Read Behavior (filter APIs) - (Before Schema Change)

1. Filter with new field (missing index column): The filtering logic checks if the corresponding virtual column exists. Since the column doesn’t exist, the filter condition is skipped. The query will return unfiltered results (possibly more data than intended).

2. Filter with old/existing indexed fields: These will work as expected.

3. Group-by with missing group column: If the group-by column is missing, the service skips the entire aggregation and returns an empty result.

#### Write Behavior
Filtering/indexes are not used in write paths — so:
- Writes work as expected.
- No errors even if index column is missing.

### Schema is Applied, but Service Not Restarted
- When the schema change (index column added) is applied:
          -- We do not need to restart the service.
          -- The SchemaValidatorUtil uses Caffeine cache with TTL of 10 minutes. Once the cache expires, the next query will 
               see the new column and include the filter/group-by in the SQL.

### Scenario 3: New Asset Type Deployed, but Table Not Yet Created

Assume:
- We add a new asset type called PizzaAsset and define aspects like PizzaBase, Toppings, etc.
- We deploy the service with model changes for this asset type and aspect schemas.
- However, the database table metadata_entity_pizzahas not yet been created (e.g., DDL not applied yet).

#### Read Behavior (Before Schema Change)

1. Get aspect for urn:li:pizza:...
-- Fails with SQL error — table doesn't exist.
-- Ebean/SQL will throw an exception (e.g., Table 'metadata_entity_pizza' doesn't exist).

2. Batch get
-- Skips any request for pizza aspects only if the column doesn't exist.
-- But if the whole table is missing, it fails.

3. Filter/group-by
 -- All filter APIs assume table exists.
 -- Fails with SQL error on access — e.g., trying to run SELECT ... FROM metadata_entity_pizza will fail immediately.

🟥 All read operations fail for that asset type if table is missing.

#### Write Operations

1. Add/Update aspect
 -- Fails with SQL error due to missing table.
-- Ebean cannot insert into a non-existent table.

🟥 All write operations fail for that asset type.

### Table is Created, but Service is Not Restarted

 - No need to restart the service.
 - The table creation will be automatically picked up by the next operation.

### Scenario 4: New Relationship Type, but Table Not Yet Created

When using GQS (single hop or multi-hop), we will get 500 errors if the table is missing or if the column which we are filtering on s missing.

### Table is Created, but Service is Not Restarted

 - No need to restart the service.
 - The table creation will be automatically picked up by the next operation.

## Testing Done
Added unit tests
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -12,6 +12,7 @@ import com.linkedin.metadata.dao.utils.RecordUtils;
 import com.linkedin.metadata.dao.utils.RelationshipLookUpContext;
 import com.linkedin.metadata.dao.utils.SQLSchemaUtils;
 import com.linkedin.metadata.dao.utils.SQLStatementUtils;
+import com.linkedin.metadata.dao.utils.SchemaValidatorUtil;
 import com.linkedin.metadata.query.Condition;
 import com.linkedin.metadata.query.LocalRelationshipCriterion;
 import com.linkedin.metadata.query.LocalRelationshipFilter;
@@ -46,6 +47,7 @@ public class EbeanLocalRelationshipQueryDAO {
   public static final String RELATIONSHIP_RETURN_TYPE = "relationship.return.type";
   public static final String MG_INTERNAL_ASSET_RELATIONSHIP_TYPE = "AssetRelationship.proto";
   private static final int FILTER_BATCH_SIZE = 200;
+  private static final String IDX_DESTINATION_DELETED_TS = "idx_destination_deleted_ts";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_destination_deleted_ts) ";
   private static final String DESTINATION_FIELD =  "destination";
   private final EbeanServer _server;
@@ -55,17 +57,20 @@ public class EbeanLocalRelationshipQueryDAO {
 
   private Set<String> _mgEntityTypeNameSet;
   private EbeanLocalDAO.SchemaConfig _schemaConfig = EbeanLocalDAO.SchemaConfig.NEW_SCHEMA_ONLY;
+  private SchemaValidatorUtil _schemaValidatorUtil;
 
   public EbeanLocalRelationshipQueryDAO(EbeanServer server, EBeanDAOConfig eBeanDAOConfig) {
     _server = server;
     _eBeanDAOConfig = eBeanDAOConfig;
     _sqlGenerator = new MultiHopsTraversalSqlGenerator(SUPPORTED_CONDITIONS);
+    _schemaValidatorUtil = new SchemaValidatorUtil(server);
   }
 
   public EbeanLocalRelationshipQueryDAO(EbeanServer server) {
     _server = server;
     _eBeanDAOConfig = new EBeanDAOConfig();
     _sqlGenerator = new MultiHopsTraversalSqlGenerator(SUPPORTED_CONDITIONS);
+    _schemaValidatorUtil = new SchemaValidatorUtil(server);
   }
 
   static final Map<Condition, String> SUPPORTED_CONDITIONS =
@@ -552,12 +557,14 @@ public class EbeanLocalRelationshipQueryDAO {
         // non-mg entity case, applying dest filter on relationship table
         filters.add(new Pair<>(destinationEntityFilter, "rt"));
       } else if (!relationshipFilter.getCriteria().isEmpty()) {
-        //TODO: Add a safeguard to check if the FORCE_IDX_ON_DESTINATION is present in the table, we can check once on bootup and then caching the result
-        // Check if any relationship-level filter is on "destination"
+        // Apply FORCE INDEX if destination field is being filtered, and the index exists
         for (LocalRelationshipCriterion criterion : relationshipFilter.getCriteria()) {
           LocalRelationshipCriterion.Field field = criterion.getField();
           if (field.getUrnField() != null && DESTINATION_FIELD.equals(field.getUrnField().getName())) {
-            sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
+            // Check if index exists on 'destination' before applying FORCE INDEX
+            if (_schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
+              sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
+            }
             break;
           }
         }
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
@@ -300,9 +300,10 @@ public class SQLStatementUtils {
    * @param nonDollarVirtualColumnsEnabled  true if virtual column does not contain $, false otherwise
    * @return translated SQL where statement
    */
-  public static String createFilterSql(String entityType, @Nullable IndexFilter indexFilter, boolean hasTotalCount, boolean nonDollarVirtualColumnsEnabled) {
+  public static String createFilterSql(String entityType, @Nullable IndexFilter indexFilter, boolean hasTotalCount, boolean nonDollarVirtualColumnsEnabled,
+      @Nonnull SchemaValidatorUtil schemaValidator) {
     final String tableName = getTableName(entityType);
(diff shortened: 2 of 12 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -552,12 +557,14 @@ private String buildFindRelationshipSQL(
...
       } else if (!relationshipFilter.getCriteria().isEmpty()) {
-        //TODO: Add a safeguard to check if the FORCE_IDX_ON_DESTINATION is present in the table, we can check once on bootup and then caching the result
-        // Check if any relationship-level filter is on "destination"
+        // Apply FORCE INDEX if destination field is being filtered, and the index exists
         for (LocalRelationshipCriterion criterion : relationshipFilter.getCriteria()) {
           LocalRelationshipCriterion.Field field = criterion.getField();
           if (field.getUrnField() != null && DESTINATION_FIELD.equals(field.getUrnField().getName())) {
-            sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
+            // Check if index exists on 'destination' before applying FORCE INDEX
+            if (_schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
```
ZihanLi58, line 565: Is the force index being used for asset filtering query? if so, do we need to add the same check there?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -112,8 +112,6 @@ public class SQLStatementUtils {
 
   private static final String INDEX_GROUP_BY_CRITERION = "SELECT count(*) as COUNT, %s FROM %s";
 
-  private static final String SQL_GET_ALL_COLUMNS =
```
ZihanLi58, line 115: We don't need this anymore?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -342,9 +340,6 @@ public static String createGroupBySql(String entityType, @Nullable IndexFilter i
     return sb.toString();
   }
 
-  public static String getAllColumnForTable(String tableName) {
```
ZihanLi58, line 345: Make sure no one else using this method? or it will be a backward incompatible change

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param indexName the index to check
+   * @return true if the index exists, false otherwise
+   */
+  public boolean indexExists(@Nonnull String tableName, @Nonnull String indexName) {
+    String lowerTable = tableName.toLowerCase();
+    String lowerIndex = indexName.toLowerCase();
+
+    tableIndexes.computeIfAbsent(lowerTable, this::loadIndexes);
+    if (!tableIndexes.get(lowerTable).contains(lowerIndex)) {
+      Set<String> refreshed = loadIndexes(lowerTable);
```
ZihanLi58, line 67: Do we need some kind of cache and periodically refresh on column? the current implementation seems to check the columns info for every request when the schema update is not applied yet, will that be expensive and potentially affect our system performance?

ZihanLi58 (commented):
can you list out what happen in the below scenario:
1. We deploy the service with new asset type, but table not created yet, what happens for read/write/filter? later on when we apply the schema change, but not restart the service, what happens for read/write/filter?
2. We deploy the service with new aspect, but column not added yet, what happens for read/write/filter? later on when we apply the schema change, but not restart the service, what happens for read/write/filter?
3. We deploy the service with new filter api, but index not added yet, what happens for read/write/filter? later on when we apply the schema change, but not restart the service, what happens for read/write/filter?
4.  We deploy the service with new relationship type,  but no table created yet, what happens for read/write/filter? later on when we apply the schema change, but not restart the service, what happens for read/write/filter? Will trait get affected?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
rakhiagr: Confirmed. It's not used anywhere outside `datahub-gma`. 
Screenshot 2025-05-19 at 5 08 26 PM

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
rakhiagr: It was used by `getAllColumnForTable` method only. Moved it to SchemaValidatorUtil class.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
rakhiagr: We are using FORCE_INDEX only at this place.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
rakhiagr: Good point. I have used Caffeine's cache to implement it.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
```diff
@@ -0,0 +1,177 @@
...
+      .build();
+
+  // Cache: tableName → Set of column names
+  private final Cache<String, Set<String>> columnCache = Caffeine.newBuilder()
+      .expireAfterWrite(10, TimeUnit.MINUTES)
+      .maximumSize(1000)
+      .build();
+
+  // Negative cache: table:column → true (missing columns)
+  private final Cache<String, Boolean> missingColumnCache = Caffeine.newBuilder()
```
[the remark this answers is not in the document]
rakhiagr: Discussed offline. Will use just 2 cache. no need for negative cache. 10 mins wait is fine.

ybz1013:
merged on behalf of @rakhiagr

PR #539, opened 2025-05-28
Updating exists() check and filter query to filter by deleted_ts column

kotkar-pallavi:
## Summary

This is PR#1 in a series of PRs which will contain code changes to support usage of `deleted_ts` column in DAO methods for read and write.

What?
- Adding a check on `deleted_ts` column in dao level queries for exists() and filter() methods.
- Updating unit tests to check the query strings now include `deleted_ts` check.
- Updating test tables to have `deleted_ts` column with NULL as the default value.

Why?
- `deleted_ts` column is required for supporting delete operations on EGG databases. Instead of doing a hard delete using `DELETE FROM <table> where <some condition>`, the records to be deleted with have a non-null value for `deleted_ts` column indicating the record has been deleted. It will actually be removed from the table periodically by a scheduled purge job. Till then, records marked as deleted should be filtered out.
- All DAO level operations need to be modified to handle the new column  
    - For read(exists, get, batchGet and filter) by doing a simple `IS NULL` check 
    - For write(create, update, delete) the logic needs to be modified and will be done in the following PRs. 

How?
- Update the filter method to ALWAYS have a check on deleted_ts column so that only records that are NOT marked for deletion are returned in the filtered results. This is consistent with the current behavior. 
- Update the exists method to return true IFF record is NOT marked as deleted. 

Ref: [RFC: Delete Asset - Soft Delete in EGG](https://docs.google.com/document/d/1hamWjkvKTdzARVWYmxVBSmIsv2CrVjkUr9Zyxm22mec/edit?usp=sharing)

## Testing Done

1. Local Build
2. Updated some existing unit tests and added 1 unit test

## Checklist

- [X] Docs related to the changes have been added/updated (if applicable)

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
@@ -111,27 +111,19 @@ public class SQLIndexFilterUtils {
           sqlFilters.add(String.format(SOFT_DELETED_CHECK, aspectColumn));
         }
 
-    for (IndexCriterion indexCriterion : indexFilter.getCriteria()) {
-      final String aspect = indexCriterion.getAspect();
-      if (!isUrn(aspect)) {
-        // if aspect is not urn, then check aspect is not soft deleted and is not null
-        final String aspectColumn = getAspectColumnName(entityType, indexCriterion.getAspect());
-        sqlFilters.add(aspectColumn + " IS NOT NULL");
-        sqlFilters.add(String.format(SOFT_DELETED_CHECK, aspectColumn));
-      }
-
-      final IndexPathParams pathParams = indexCriterion.getPathParams(GetMode.NULL);
-      if (pathParams != null) {
-        validateConditionAndValue(indexCriterion);
-        final Condition condition = pathParams.getCondition();
-        final String indexColumn = getGeneratedColumnName(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled);
-        final String tableName = SQLSchemaUtils.getTableName(entityType);
-        // New: Skip filter if column doesn't exist
-        if (!schemaValidator.columnExists(tableName, indexColumn)) {
-          log.warn("Skipping filter: virtual column '{}' not found in table '{}'", indexColumn, tableName);
-          continue;
+        final IndexPathParams pathParams = indexCriterion.getPathParams(GetMode.NULL);
+        if (pathParams != null) {
+          validateConditionAndValue(indexCriterion);
+          final Condition condition = pathParams.getCondition();
+          final String indexColumn = getGeneratedColumnName(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled);
+          final String tableName = SQLSchemaUtils.getTableName(entityType);
+          // New: Skip filter if column doesn't exist
+          if (!schemaValidator.columnExists(tableName, indexColumn)) {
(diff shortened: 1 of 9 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
```diff
@@ -120,9 +120,9 @@ public static String parseIndexFilter(@Nonnull String entityType, @Nullable Inde
     }
 
     if (sqlFilters.isEmpty()) {
-      return "";
+      return "WHERE deleted_ts IS NULL ";
     } else {
-      return "WHERE " + String.join("\nAND ", sqlFilters);
+      return "WHERE deleted_ts IS NULL \nAND " + String.join("\nAND ", sqlFilters);
```
ZihanLi58, line 125: Just a suggestion: can you add deleted_ts IS NULL  at the end of the query? or even can you just add  "deleted_ts IS NULL" in to sqlFilters?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
```diff
@@ -98,7 +98,7 @@ public static String parseIndexFilter(@Nonnull String entityType, @Nullable Inde
     List<String> sqlFilters = new ArrayList<>();
 
     if (indexFilter == null || !indexFilter.hasCriteria()) {
-      return "";
+      return "WHERE deleted_ts IS NULL ";
```
ZihanLi58, line 101: an you help make sure this is only used for asset filtering? Asking because in relationship table, we actually sometime want to see the record where deleted_ts is not null. Also update the java doc to reflect this change

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -114,7 +114,7 @@ public class SQLStatementUtils {
 
   private static final String SQL_GET_ALL_COLUMNS =
       "SELECT COLUMN_NAME FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = database() AND TABLE_NAME = '%s'";
-  private static final String SQL_URN_EXIST_TEMPLATE = "SELECT urn FROM %s WHERE urn = '%s'";
+  private static final String SQL_URN_EXIST_TEMPLATE = "SELECT urn FROM %s WHERE urn = '%s' AND deleted_ts IS NULL";
```
ZihanLi58, line 117: Double confirm here that you went through all place where we call the exist and verify none of them will be break by this change right?

kotkar-pallavi: yes, verified it and added details of usage and how to verify updated API logic in [this comment ](https://linkedin.atlassian.net/browse/META-22475?focusedCommentId=37025006)of the tracking ticket

i will do snapshot testing for the GMSs with hot-fix branch on metadata-models to minimize he potential impact radius. 🤞

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
kotkar-pallavi: as discussed offline, i didnt find any param/flag indicating getting soft deleted records.
also confirmed, i did not find ay usage of exists() in relationship. 

having said that, to minimize the risk I will do snapshot testing with the hot-fix version before updating version of datahub-gma in master branch.

PR #543, opened 2025-06-05
[feat] add with optimistic locking in BaseLocalDAO

ybz1013:
## Summary
IngestionParams now can contain a list of IngestionETags (aspect name -> timestamp). If this is provided by user, we will use the provided timestamp for optimistic locking for given aspect.

The optimistic locking mechanism was already in place. Before it was locking only during the ingestion, now since we have given the eTags, the locking period can be extended for user to have read-modify-write consistency.

## Testing Done
./graldew build

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -76,6 +76,8 @@ import lombok.Data;
 import lombok.NonNull;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
+import pegasus.com.linkedin.metadata.events.IngestionAspectETag;
+import pegasus.com.linkedin.metadata.events.IngestionAspectETagArray;
 
 import static com.linkedin.metadata.dao.utils.IngestionUtils.*;
 import static com.linkedin.metadata.dao.utils.ModelUtils.*;
@@ -484,7 +486,6 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     _emitAuditEvent = emitAuditEvent;
   }
 
-
   /**
    * Logic common to both {@link #add(Urn, Class, Function, AuditStamp)} and {@link #delete(Urn, Class, AuditStamp, int)} methods.
    *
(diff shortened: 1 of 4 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -549,17 +550,42 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
+      return null;
+    }
+
+    AuditStamp optimisticLockAuditStamp = null;
+
+    final IngestionAspectETagArray ingestionAspectETags = ingestionParams.getIngestionETags();
+
+    if (ingestionAspectETags != null) {
+      for (IngestionAspectETag ingestionAspectETag: ingestionAspectETags) {
+        if (aspectClass.getSimpleName().equals(ingestionAspectETag.getAspect_name()) && ingestionAspectETag.getETag() != null) {
```
yangyangv2: I had a question earlier in another PR about whether to use aspect field name or aspect class name for this comparison. the aspect field name is future oriented, can we can start using aspect field name from internal logic going forward.

The consideration of aspect field name is - the asset type + aspect field name will be come a AspectURI, which is globally unique and consistent from modeling to storage schemas. More consideration is from here:

https://docs.google.com/document/d/1eqSYTf9jgUx5w0x_IBGfochF1H9A65TCiqI6fEMu1OY/edit?tab=t.0#heading=h.932rva5dqbdh

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -549,17 +550,42 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
+          optimisticLockAuditStamp = new AuditStamp();
+          optimisticLockAuditStamp.setTime(ingestionAspectETag.getETag());
```
yangyangv2, line 583: What is the reason only taking one AuditStamp from the ingestionAspectETags? The value of e-Tag should come from the storage time of each aspect, which are different. For example

T1: user1 --(write)--> Asset1 { Aspect1 {T1}}
T2: user2 --(write)--> Asset1 {Aspect2 {T2}}

T3: user3 <--(read)-- Asset {Aspect1 {T1}, Aspect2 {T2}}
T4: user4 --{modify and write} -->  Asset {Aspect1" {T1}, Aspect2" {T2}}

T5: MG should compare the modified Time T1 in Aspect1" and T2 in Aspect2" with what's been stored on the database and decide whether this update can be made. It should be different Ts.

Let me know if I miss anything -- and it's more clear if you include the comments for this method.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1313,15 +1339,15 @@ private <ASPECT extends RecordTemplate> void applyRetention(@Nonnull URN urn, @N
    * @param urn the URN for the entity the aspect is attached to
    * @param aspectClass the aspectClass of the aspect being saved
    * @param oldEntry {@link RecordTemplate} of the previous latest value of aspect, null if new value is the first version
-   * @param oldAuditStamp the audit stamp of the previous latest aspect, null if new value is the first version
+   * @param optimisticLockAuditStamp the audit stamp of the previous latest aspect, null if new value is the first version. Used for optimistic locking.
```
yangyangv2, line 1344: Thanks for change this name, it's more for its purpose.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
ybz1013: the workflow you described is correct. The reason we only take one etag is this addCommon is only handling one aspect at a time.

I believe in the asset service API, it takes the request asset and do a for each aspect loop and call this add here multiple time (each with one aspect only).

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
ybz1013: will use field name

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
```diff
@@ -802,4 +804,18 @@ public void testAspectCallbackHelperWithUnregisteredAspect() throws URISyntaxExc
     // Verify that the result is the same as the input aspect since it's not registered
     assertEquals(result.getUpdatedAspect(), foo);
   }
+
+  @Test
```
kotkar-pallavi, line 808: can we also add a small test for the `if (ingestionParams == null)` code branch?

kotkar-pallavi (commented):
Couple of high level questions:
1. Where do users retrieve the value of last modified Ts? i am not sure if the response for Get has the AuditStamp info being returned in response. 
2. Can we confirm the time in audit stamp is updated when aspect update happens with update API?

dao-api/src/test/java/com/linkedin/metadata/dao/BaseLocalDAOTest.java
ybz1013: for your questions:

1. I can check on that later. I think it should have it in the get response, if not, we can add another endpoint for that.
2. yes, it will either use producer provided timestamp or current system time

ybz1013: 1. we do not currently have the timestamp in the aspect response. I have updated the doc to include a new API for this. Please merge this PR for now.

kotkar-pallavi: thank you!

PR #544, opened 2025-06-10
[feat] Updating methods for get() and batchGet() APIs to filter by deleted_ts

kotkar-pallavi:
## Summary

This is PR#1 in a series of PRs which will contain code changes to support usage of deleted_ts column in DAO methods for read and write.

What?

* Updating the DAO queries for get and batchGet APIs to filter return rows only where `deleted_ts IS NULL`
* Updating unit tests to have an additional check on `deleted_ts`

Why?
* The DB table may contain "deleted" rows till the purge job runs. 
* Get and BatchGet APIs should not return results based on rows which are marked for deletion, i.e. records with `deleted_ts=<some_value>` should be excluded.

How?
* Updated the queries for get and batchGet to ALWAYS have the check `WHERE deleted_ts IS NULL` so that records marked for deletion are not shown in response of these api's.

Ref: [RFC: Delete Asset - Soft Delete in EGG](https://docs.google.com/document/d/1hamWjkvKTdzARVWYmxVBSmIsv2CrVjkUr9Zyxm22mec/edit?usp=sharing)

## Testing Done
- Local testing
- Fixed unit tests

## Checklist


jphui (approved):
Chatted offline, lgtm

PR #551, opened 2025-06-20
[fix] minor fixes related to optimistic locking

ybz1013:
## Summary
Found two bugs during user-facing optimistic locking update bug bash.

1. For some reason, the IngestionParams.TestMode "false" value from the asset service API call will be ignored. The code in add will create a default IngestionParams in that case, which will overwrite our eTags.
2. the shouldUpdateAspect should also catch the case when user provided timestamp is older than the latest timestamp from DB. 

UPDATE:
seems like 1. is fixed just now in 
https://github.com/linkedin/datahub-gma/commit/d136c6fd272e427e2dcd1bc17ebf7de797e25e3b

## Testing Done
./gradlew build
## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -2035,11 +2036,22 @@ protected boolean aspectVersionSkipWrite(@Nullable RecordTemplate newValue, @Nul
     return aspectVersionComparator(newValue, oldValue) == -1;
   }
 
+  /**
+   * Returns true if we should skip writing newValue(when etag timestamp is older than latest timestamp).
+   */
+  @VisibleForTesting
+  protected boolean aspectTimestampSkipWrite(@Nullable AuditStamp eTagValue, @Nullable AuditStamp oldValue) {
+    return oldValue != null && eTagValue != null
```
kotkar-pallavi, line 2044: will this fail on oldValue=null i.e. when new asset is created 1st time using upsert or create?

and also when eTag is not provided?

ybz1013: nope, if oldValue is null then it will not skip. When eTag is not provided, it will also not skip because the default update does not have eTag. These are covered in unit tests.

kotkar-pallavi: oh i see, its an AND of all. 
yes i see the last ut. thanks!

kotkar-pallavi (approved):

PR #555, opened 2025-06-25
[EGG Migration] Bug fix for dual read

rakhiagr:
## Summary
This pull request refines the logic for determining which aspect value to use in the `getInternalAspectsWithShadowComparison` method of `BaseAspectRoutingResource.java`. It introduces clearer handling for different cases of local and shadow data presence, improves logging messages, and adjusts the behavior for shadow-only data.

Key changes:

### Logic Refinement:
* Refactored the conditional logic to explicitly handle four cases: both local and shadow present and equal, both present but not equal, only shadow present, and only local present. This improves code readability and maintainability.

### Logging Improvements:
* Enhanced log messages to provide more context about each case, including warnings for mismatches or shadow-only data and better explanations of the actions taken.

### Behavior Adjustment:
* Modified the behavior for the "only shadow present" case to skip the shadow-only data instead of using it, aligning with the intended logic.

## Testing Done
./gradlew build
## Checklist

### DIFF
--- a/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectRoutingResource.java
+++ b/restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectRoutingResource.java
@@ -601,15 +601,24 @@ public abstract class BaseAspectRoutingResource<
           java.util.Optional<? extends RecordTemplate> shadow = shadowResults.getOrDefault(key, java.util.Optional.empty());
 
           RecordTemplate valueToUse = null;
-          if (shadow.isPresent() && local.isPresent() && !Objects.equals(local.get(), shadow.get())) {
-            log.warn("Aspect mismatch for URN {} and aspect {}: local = {}, shadow = {}",
-                key.getUrn(), key.getAspectClass().getSimpleName(), local.get(), shadow.get());
-            valueToUse = local.get();
+          if (local.isPresent() && shadow.isPresent()) {
+            if (Objects.equals(local.get(), shadow.get())) {
+              // Case 1: both present and equal → use shadow
+              valueToUse = shadow.get();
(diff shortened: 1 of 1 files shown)


restli-resources/src/main/java/com/linkedin/metadata/restli/BaseAspectRoutingResource.java
```diff
@@ -601,15 +601,24 @@ private List<INTERNAL_ASPECT_UNION> getInternalAspectsWithShadowComparison(Set<A
...
+            }
           } else if (shadow.isPresent()) {
-            log.warn("Only shadow value present for URN {} and aspect {}", key.getUrn(), key.getAspectClass().getSimpleName());
-            valueToUse = shadow.get();
+            // Case 3: only shadow present → log warning, skip
+            log.warn("Only shadow value present for URN {} and aspect {}. Skipping shadow-only data.",
+                key.getUrn(), key.getAspectClass().getSimpleName());
           } else if (local.isPresent()) {
-            log.info("Only local value present for URN {} and aspect {}. Using local.", key.getUrn(), key.getAspectClass().getSimpleName());
+            // Case 4: only local present → use local, log warning
```
yihzhou, line 619: What will happen is both are not present? we just skip value?

rakhiagr: Yes.

rakhiagr: `RecordTemplate valueToUse = null;` - this will remain null which is the expected behavior.

yihzhou (approved):

PR #562, opened 2025-07-29
[feat] support filter criterion with logical expression (AND/OR)

ybz1013:
## Summary
The old filter doesn't support AND/OR between criterion. It will always try to figure out the logic by looking at the field name, and apply OR to the same fields and AND between field groups.

This PR will allow a more flexible representation of the logical expression in filters. It will now allow user to use a tree structured filter. And added a new API v4 for only this filter use.

Changes:
1. new logical expression filter model. It is on the criterion level, but in the future, we may want to make a filter level one to allow AND/OR between different type of assets.
2. new V4 API, which ONLY supports the new logical expression (and new validation logic)
3. normalize logic to convert old filter criteria into the new model (to unify SQL building logic)
4. new SQL building logic that uses new filter models
5. related util

Note:
1. v2/v3 API can only use old filter model, and v4 API can only use new filter model (both are checked)
2. old schema will still have the same only one urn criterion restriction, but it can be in the new filter model, so v4 still supports old schema

## Testing Done
./gradlew build
unit tests added

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
@@ -1,5 +1,6 @@
 package com.linkedin.metadata.dao.utils;
 
+import com.linkedin.data.template.SetMode;
 import com.linkedin.metadata.query.LocalRelationshipCriterion;
 import com.linkedin.metadata.query.LocalRelationshipCriterionArray;
 import com.linkedin.metadata.query.LocalRelationshipFilter;
@@ -70,7 +71,9 @@ public final class LogicalExpressionLocalRelationshipCriterionUtils {
     }
 
     // return a hard copy of filter with new LogicalExpressionCriteria
-    return new LocalRelationshipFilter().setLogicalExpressionCriteria(root);
+    return new LocalRelationshipFilter()
+        .setLogicalExpressionCriteria(root)
+        .setDirection(filter.getDirection(), SetMode.IGNORE_NULL);
   }
 
   /**


dao-api/src/main/pegasus/com/linkedin/metadata/query/LocalRelationshipFilter.pdl
```diff
@@ -13,4 +15,9 @@ record LocalRelationshipFilter {
    * The relationship direction between two nodes.
    */
   direction: optional RelationshipDirection
+
+  /**
+   * Logical Expression Criteria. Tree structure
+   */
+  logicalExpressionCriteria: optional LogicalExpressionLocalRelationshipCriterion
```
jsdonn, line 22: is this supposed to be an array?

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/query/LogicalOperation/Operator.pdl
```diff
@@ -0,0 +1,8 @@
+namespace pegasus.com.linkedin.metadata.query.LogicalOperation
+package pegasus.com.linkedin.metadata.query.innerLogicalOperation
```
jsdonn, line 2: why do we need namespace and package?

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/query/LogicalOperation/Operator.pdl
```diff
@@ -0,0 +1,8 @@
...
+
+enum Operator {
+  UNKNOWN
```
jsdonn, line 5: what is unknown?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -322,9 +330,48 @@ public <ASSET_RELATIONSHIP extends RecordTemplate, RELATIONSHIP extends RecordTe
...
+  @Nonnull
+  public <ASSET_RELATIONSHIP extends RecordTemplate, RELATIONSHIP extends RecordTemplate> List<ASSET_RELATIONSHIP> findRelationshipsV4(
+      @Nullable String sourceEntityType, @Nullable LocalRelationshipFilter sourceEntityFilter,
+      @Nullable String destinationEntityType, @Nullable LocalRelationshipFilter destinationEntityFilter,
+      @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter,
+      @Nonnull Class<ASSET_RELATIONSHIP> assetRelationshipClass, @Nullable Map<String, Object> wrapOptions,
+      int offset, int count, RelationshipLookUpContext relationshipLookUpContext) {
+    if (wrapOptions == null || !wrapOptions.containsKey(RELATIONSHIP_RETURN_TYPE)
+        || !MG_INTERNAL_ASSET_RELATIONSHIP_TYPE.equals(wrapOptions.get(RELATIONSHIP_RETURN_TYPE))) {
+      throw new IllegalArgumentException("Please check your use of the findRelationshipsV3 method.");
```
jsdonn, line 372: v3?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -410,23 +457,48 @@ private <ENTITY extends RecordTemplate> void validateEntityFilter(@Nonnull Local
...
+      throw new IllegalArgumentException(
+          "Please do not use both the 'criteria' field and the 'logicalExpressionCriteria' field.");
+    }
+
+    if (logicalExpressionFilterEnabled && filter.hasCriteria()) {
+        throw new IllegalArgumentException(
+            "Please do not use the 'criteria' field and use the 'logicalExpressionCriteria' field instead for findRelationshipsV4 API.");
+    } else if (!logicalExpressionFilterEnabled && filter.hasLogicalExpressionCriteria()) {
+        throw new IllegalArgumentException(
+            "Please do not use the 'logicalExpressionCriteria' field and use the 'criteria' field instead for findRelationshipsV2/V3 API.");
```
jsdonn, line 504: should we still be recommending v2 when there is v3?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -438,50 +510,108 @@ private void validateEntityTypeAndFilter(@Nullable LocalRelationshipFilter filte
...
+            String.format("Condition %s is not supported by local relationship DAO.", condition));
       }
-      return null; // unused
-    });
+  }
+
+  /**
+   * Recursively validate logical expression criteria. Checks Operator and uses validateLocalRelationshipCriterion to check each criterion.
+   */
+  private static void validateLogicalExpression(@Nullable LogicalExpressionLocalRelationshipCriterion logicalExpressionCriteria) {
```
jsdonn, line 588: do we need to add a limit on how many layers of logical expressions a user can add or how many clauses/branches a logical expression can have so that this method does not go through too many recursions?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -438,50 +510,108 @@ private void validateEntityTypeAndFilter(@Nullable LocalRelationshipFilter filte
...
+      }
+
+      if (expr.isCriterion()) {
+        validateLocalRelationshipCriterion(expr.getCriterion());
+      }
+
+      if (expr.isLogical()) {
+        LogicalOperation logical = expr.getLogical();
+
+        if (!logical.hasOp() || logical.getOp() == Operator.UNKNOWN || logical.getOp() == Operator.$UNKNOWN) {
```
jsdonn, line 606: why we need two different conditions to check for UNKNOWN?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -322,9 +330,48 @@ public <ASSET_RELATIONSHIP extends RecordTemplate, RELATIONSHIP extends RecordTe
...
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Finds a list of relationships of a specific type (Urn) based on the given filters if applicable.
+   * Similar to findRelationshipsV3, but this method takes filters with logical expressions (AND/OR).
+   * The intended use case is for MG internally with AssetRelationship, but since it is an open API, we are leaving room for extendability.
+   *
+   * @param sourceEntityType type of source entity to query (e.g. "dataset")
+   * @param sourceEntityFilter the filter to apply to the source entity when querying (not applicable to non-MG entities)
```
jsdonn: specify that criteria must be null and logicalExpressionCriteria must be nonnull?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
```diff
@@ -0,0 +1,162 @@
...
+      if (!current.hasExpr()) {
+        continue;
+      }
+
+      final LogicalExpressionLocalRelationshipCriterion.Expr expr = current.getExpr();
+
+      if (expr.isCriterion()) {
+        flattened.add(expr.getCriterion());
+      } else if (expr.isLogical()) {
+        final LogicalOperation operation = expr.getLogical();
```
jsdonn, line 105: what if operation is UNKNOWN here?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
```diff
@@ -0,0 +1,162 @@
...
+        final List<LogicalExpressionLocalRelationshipCriterion> expressions = operation.getExpressions();
+        // reverse to maintain original left-to-right order
```
jsdonn, line 107: does order matter much?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
```diff
@@ -0,0 +1,162 @@
...
+   * with the given operator.
+   * @param op one of the Operator, e.g. AND
+   * @param children LogicalExpressionLocalRelationshipCriterionArray
+   * @return LogicalExpressionLocalRelationshipCriterion
+   */
+  @Nonnull
+  public static LogicalExpressionLocalRelationshipCriterion buildLogicalGroup(@Nonnull Operator op,
+      @Nonnull LogicalExpressionLocalRelationshipCriterionArray children) {
+    if (children.size() == 1) {
+      // avoid wrapping unnecessarily
```
jsdonn, line 146: if children.size is 0, do we still want to wrap?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -473,48 +478,56 @@ public static String whereClause(@Nonnull Map<Condition, String> supportedCondit
...
-      if (orClauses.size() == 1) {
-        andClauses.add(orClauses.get(0));
-      } else {
-        andClauses.add("(" + String.join(" OR ", orClauses) + ")");
-      }
+  private static String buildSQLQueryFromLogicalExpression(@Nonnull LogicalExpressionLocalRelationshipCriterion criterion,
+      @Nonnull Map<Condition, String> supportedConditions, @Nullable String tablePrefix,
+      boolean nonDollarVirtualColumnsEnabled) {
+    if (!criterion.hasExpr() || criterion.getExpr() == null) {
+      throw new IllegalArgumentException("no expr");
```
jsdonn: better error message?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -473,48 +478,56 @@ public static String whereClause(@Nonnull Map<Condition, String> supportedCondit
...
       }
-      return andClause;
     }
 
-    return String.join(" AND ", andClauses);
+    // expr is logical
+    final LogicalOperation logicalOperation = expr.getLogical();
+
+    final Operator op = logicalOperation.getOp();
+    final String opString = op == Operator.AND ? " AND " : " OR ";
```
jsdonn, line 522: double checking your intention to handle UNKNOWN as OR?

dao-api/src/main/pegasus/com/linkedin/metadata/query/LocalRelationshipFilter.pdl
ybz1013: no, because then we dont know how to combine between the elements. i.e. are they AND'd or OR'd together?

this is essentially a wrapper of the array + AND/OR logic anyway.

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/query/LogicalOperation/Operator.pdl
ybz1013: not sure, this is generated from metadata-models... grpc team gave me the steps to do it.

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/query/LogicalOperation/Operator.pdl
ybz1013: I think it is because in protobuf we have to have an unknown as default value, and this pdl is generated from that protobuf.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: good call, v2 should really be deprecated since it doesn't support relationship v2 well (it doesn't return AssetRelationship which has the source urn).

But it is still not deprecated, that way I added it here.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: maybe 3 or 5?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: I dont know which will be actually converted by the proto bridge. This catches both, so its safer IMO

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
ybz1013: discussed in person, no need for now

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
ybz1013: I hope the validation logic will catch all UNKOWN, so we only need to check for that once.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
ybz1013: not really, but could matter in the future?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/LogicalExpressionLocalRelationshipCriterionUtils.java
ybz1013: good call. Returning null and fixed upstream code.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
ybz1013: ^ validation should catch UNKNOWN

ybz1013:
tested with GQS (with ab testing with old code). Old workflow non-broken.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jsdonn: do we have plans to deprecate it?

ybz1013: no plans for now. Im not sure who's still using it

PR #564, opened 2025-08-05
feat(filter): add ARRAY_CONTAINS condition for filtering

jsdonn:
## Summary
Add ARRAY_CONTAINS condition to support the filtering case where you want to find if an aspect's array field contains the value you provide. This will be useful for labeling where the labels will be stored as an array of strings and the user is looking for all assets which have a particular label.

## Testing Done
Add unit test and check the generated sql statement.

Unfortunately this cannot be properly tested since MariaDB (used for local testing) does not support the MEMBER OF function. Similar behavior can be achieved with JSON_CONTAINS but it is also slower than MEMBER OF. Further testing will happen at the service integration layer instead of here in the DAO library.
## Checklist

### DIFF
--- a/dao-api/src/main/pegasus/com/linkedin/metadata/query/Condition.pdl
+++ b/dao-api/src/main/pegasus/com/linkedin/metadata/query/Condition.pdl
@@ -49,4 +49,9 @@ enum Condition {
    * Represent the relation: String field starts with value, e.g. name starts with PageView
    */
   START_WITH
+
+  /**
+   * Represent the relation: array field contains string value, e.g. labels{["abc", "def"]} contains "abc"
+   */
+  ARRAY_CONTAINS
 }
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
@@ -179,19 +179,6 @@ public class EbeanLocalDAO<ASPECT_UNION extends UnionTemplate, URN extends Urn>
     public Object value;
   }
 
-  private static final Map<Condition, String> CONDITION_STRING_MAP =
-      Collections.unmodifiableMap(new HashMap<Condition, String>() {
-        {
-          put(Condition.EQUAL, "=");
-          put(Condition.GREATER_THAN, ">");
-          put(Condition.GREATER_THAN_OR_EQUAL_TO, ">=");
-          put(Condition.IN, "IN");
-          put(Condition.LESS_THAN, "<");
-          put(Condition.LESS_THAN_OR_EQUAL_TO, "<=");
-          put(Condition.START_WITH, "LIKE");
-        }
-      });
-
   /**
    * Constructor for EbeanLocalDAO.
    *
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
@@ -143,6 +143,8 @@ public class SQLIndexFilterUtils {
    */
   private static String parseSqlFilter(String indexColumn, Condition condition, IndexValue indexValue) {
     switch (condition) {
+      case ARRAY_CONTAINS:
+        return String.format("'%s' MEMBER OF(%s)", parseIndexValue(indexValue), indexColumn);
       case CONTAIN:
         return String.format("JSON_SEARCH(%s, 'one', '%s') IS NOT NULL", indexColumn, parseIndexValue(indexValue));
       case IN:
@@ -204,7 +206,7 @@ public class SQLIndexFilterUtils {
     final Condition condition = criterion.getPathParams().getCondition();
     final IndexValue indexValue = criterion.getPathParams().getValue();
 
-    if (condition == Condition.IN && (!indexValue.isArray() || indexValue.getArray().size() == 0)) {
+    if (condition == Condition.IN && (!indexValue.isArray() || indexValue.getArray().isEmpty())) {
       throw new IllegalArgumentException("Invalid condition " + condition + " for index value " + indexValue);
     }
   }
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
@@ -11,6 +11,7 @@ import com.linkedin.metadata.query.IndexSortCriterion;
 import com.linkedin.metadata.query.IndexValue;
 import com.linkedin.metadata.query.SortOrder;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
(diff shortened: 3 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -179,19 +179,6 @@ static class GMAIndexPair {
     public Object value;
   }
 
-  private static final Map<Condition, String> CONDITION_STRING_MAP =
```
jsdonn, line 182: this was unused

dao-api/src/main/pegasus/com/linkedin/metadata/query/Condition.pdl
```diff
@@ -49,4 +49,9 @@ enum Condition {
    * Represent the relation: String field starts with value, e.g. name starts with PageView
    */
   START_WITH
+
+  /**
+   * Represent the relation: array field contains string value, e.g. labels{["abc", "def"]} contains "abc"
+   */
+  ARRAY_CONTAINS
```
ZihanLi58, line 56: I think we define this condition in metadata-models? do we still use this old one?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
```diff
@@ -165,6 +166,28 @@ public void testCreateFilterSql() {
...
+
+    IndexCriterion indexCriterion1 =
+        SQLIndexFilterUtils.createIndexCriterion(AspectFooBar.class, "bars", Condition.ARRAY_CONTAINS,
+            IndexValue.create("bar1"));
+
+    indexCriterionArray.add(indexCriterion1);
+    indexFilter.setCriteria(indexCriterionArray);
+
+    String sql1 = SQLStatementUtils.createFilterSql("foo", indexFilter, true, false, mockValidator);
+    String expectedSql1 = "SELECT *, (SELECT COUNT(urn) FROM metadata_entity_foo WHERE a_aspectfoobar IS NOT NULL\n"
```
ZihanLi58, line 182: This sql cmd does not make sense to me, can you help clarify that?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
```diff
@@ -143,6 +143,8 @@ public static String parseIndexFilter(@Nonnull String entityType, @Nullable Inde
    */
   private static String parseSqlFilter(String indexColumn, Condition condition, IndexValue indexValue) {
     switch (condition) {
+      case ARRAY_CONTAINS:
+        return String.format("'%s' MEMBER OF(%s)", parseIndexValue(indexValue), indexColumn);
```
ZihanLi58, line 149: I know this is POC, but do we need to add some validation later here like the type of the index column should be array?

dao-api/src/main/pegasus/com/linkedin/metadata/query/Condition.pdl
ZihanLi58: Also any limitation on this index value? can it be array of string? or it has be to a single string?

jsdonn: The condition needs to be defined both here and in metadata-models. It is needed here in pdl format since we don't have access to the internal translated pdls. Here are the steps to modify a pdl in datahub-gma which I followed (thanks Yanbo):
```
Add proto model in metadata-dao/src/main/dataProto.
Enable translateProtoToPdl task for metadata-dao [here](https://github.com/linkedin-multiproduct/metadata-models/blob/29e19e3190f1e81c350eebb2a0a328ebb23a92f5/metadata-proto/externalProtoformingTasks.gradle#L5) (externalProtoformingTasks.gradle) (translateProtoToPdlBridged.enabled)
execute ./gradlew :metadata-dao:translateProtoToPdlBridged
copy the PDLs generated in metadata-dao/src/main/pegasus in datahub-gma dao-api/src/main/pegasus
release datahub-gma version and use it in metadata-models
disable `translateProtoToPdlBridged` enabled in step2. now run metadata-models build
submit proto schemas changes in metadata-dao
```

jsdonn: > Also any limitation on this index value? can it be array of string? or it has be to a single string?

I am thinking it has to be a single string, I can add validation for that. You can always combine multiple filter criteria to do `"abc" in labels OR "def" in labels`

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
jsdonn: We want to select all columns (*) AND total_count of rows that match the filter criteria. This is a really dumb way to do it where you can think of it like:
```
QUERY = SELECT * FROM table WHERE conditions;

JOIN 

SELECT COUNT(*) FROM QUERY;
```

This was before MySQL 8 so it could be updated to use WITH (not available in previous versions) to look cleaner...but don't want to touch that in the scope of this PR.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
jsdonn: It's a little hard because it will require an additional query to the db to check that column's value first. Then depending on the result, it may throw a validation error if the value is not a json array i.e. begins with `[` and ends with `]`. If the index column type is not an array, it will not throw any errors during queries, it will just return an empty result:
```
MySQL [mg_db_1_ei]> select * from metadata_entity_demo where "abc" member of(a_status);
Empty set (0.04 sec)
```

But I agree, it would be best to add validation here from the user's perspective so they know their query does not make sense instead of just silently failing. I will probably do so during actual implementation, let me add a TODO comment.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
ZihanLi58: hmm, if I read it correctly, it seems to be select * , (count(*) from QUERT ) FROM table WHERE conditions? which does not seem correct to me

jsdonn:
```
SELECT *, (
    SELECT COUNT(urn)
    FROM metadata_entity_foo
    WHERE a_aspectfoobar IS NOT NULL
    AND JSON_EXTRACT(a_aspectfoobar, '$.gma_deleted') IS NULL
    AND 'bar1' MEMBER OF(i_aspectfoobar$bars)
    AND deleted_ts IS NULL
  ) as _total_count 
FROM metadata_entity_foo
WHERE a_aspectfoobar IS NOT NULL
AND JSON_EXTRACT(a_aspectfoobar, '$.gma_deleted') IS NULL
AND 'bar1' MEMBER OF(i_aspectfoobar$bars)
AND deleted_ts IS NULL
```

should be correct - since _total_count is gonna be the same value in the whole column. so the result is something like:
```
urn | aspect1 | _total_count
a   | ...     | 2
b   | ...     | 2
```

jsdonn: it's like doing 
```
A = SELECT * FROM table WHERE conditions;
B = SELECT COUNT(*) FROM A;

result = | A | + | B |
```

jsdonn: we can't do just `SELECT *, COUNT(urn) FROM table WHERE conditions` because this will just return 1 row

PR #571, opened 2025-08-13
feat(collections): add annotations for handling collections

jsdonn:
## Summary
Add `@gma.collection.` annotation for supporting collections. There are 3 items to specify:

1) isCollection - should be true for collections. default false.

2) primaryKeys - struct (essentially a map) representing which fields to use as primary keys for collection-typed fields. for example, given:
```
upstream: {
    dataset_urn: string
    audit_stamp: {
        time: long
        actor: string
    }
}
```
I can specify that I want the primary key for this field to be the dataset_urn and actor fields. i.e. `"upstream": ["dataset_urn", "audit_stamp.actor"]`

3) defaultUpdateBehavior - enum indicating how to handle updates to existing entries in the collection. default REPLACE_BY_KEY

Next step is to add logic to metadata models plugin to convert proto annotations to this pdl annotation

## Testing Done
Add a unit test. 
## Checklist

### DIFF
--- a/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/CollectionAnnotation.pdl
+++ b/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/CollectionAnnotation.pdl
@@ -1,0 +1,30 @@
+namespace com.linkedin.metadata.annotations
+
+/**
+ * Annotation for collections-related attributes
+ */
+record CollectionAnnotation {
+  /**
+   * Whether a model is a collection
+   */
+  isCollection: optional boolean
+
+  /**
+   * A list of key-values representing the primary keys to use for updating collection-typed fields. The keys are
+   * strings representing collections field names. The values are lists of field masks indicating the key(s) to be used for updates on that field.
+   *
+   * e.g. [{"my_struct.foo": {"paths": ["x.y", "x.z"]}}] -> the foo field of my_struct is a list, which has primary keys of x.y (where x is a struct
+   * and y is a field of x) and x.z (where z is also a field of x). A real world example could be lineage where x is an Upstream struct,
+   * x.y is the dataset urn, and x.z is the lastmodifiedby value.
+   */
+  primaryKeys: optional array[map[string, map[string, array[string]]]]
+
+  /**
+   * How to handle updates. The default/implicit behavior is REPLACE_BY_KEY
+   */
+  defaultUpdateBehavior: optional enum UpdateBehavior {
+    REPLACE_BY_KEY,
+    REPLACE_BY_ACTOR,
+    REPLACE_ALL
+  }
+}
--- a/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/GmaAnnotation.pdl
+++ b/gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/GmaAnnotation.pdl
@@ -27,4 +27,10 @@ record GmaAnnotation {
    * Information about GMA Search functionality.
    */
   search: optional SearchAnnotation
+
+  /**
+   * Information about collections
+   */
+  collection: optional CollectionAnnotation
+
 }
--- a/gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java
+++ b/gradle-plugins/metadata-annotations-lib/src/test/java/com/linkedin/metadata/annotations/GmaAnnotationParserTest.java
@@ -5,10 +5,14 @@ import com.google.common.collect.ImmutableSetMultimap;
 import com.linkedin.data.schema.RecordDataSchema;
 import com.linkedin.data.template.DataTemplateUtil;
 import com.linkedin.data.template.StringArray;
+import com.linkedin.data.template.StringArrayMap;
+import com.linkedin.data.template.StringArrayMapMap;
+import com.linkedin.data.template.StringArrayMapMapArray;
 import com.linkedin.testing.AnnotatedAspectBar;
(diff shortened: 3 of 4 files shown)


gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/CollectionAnnotation.pdl
```diff
@@ -0,0 +1,30 @@
...
+   * and y is a field of x) and x.z (where z is also a field of x). A real world example could be lineage where x is an Upstream struct,
+   * x.y is the dataset urn, and x.z is the lastmodifiedby value.
+   */
+  primaryKeys: optional array[map[string, map[string, array[string]]]]
+
+  /**
+   * How to handle updates. The default/implicit behavior is REPLACE_BY_KEY
+   */
+  defaultUpdateBehavior: optional enum UpdateBehavior {
+    REPLACE_BY_KEY,
```
ZihanLi58, line 26: how do you plan to support delete?

gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/GmaAnnotation.pdl
```diff
@@ -27,4 +27,10 @@ record GmaAnnotation {
    * Information about GMA Search functionality.
    */
   search: optional SearchAnnotation
+
+  /**
+   * Information about collections
+   */
+  collection: optional CollectionAnnotation
```
ZihanLi58, line 34: I think we have another annotation defined in the plugin, how will that work? can we consolidate the two here? I will prefer us to deprecate the ppl annotation here at all.

jsdonn: The annotation on the internal model will get translated to this pdl annotation (there will be logic in metadata-models-plugin to do this). We cannot fully deprecate the pdl annotation without either moving datahub-gma into a LI MP or starting to rely on proto in datahub-gma. 

Essentially, the proto annotation (i.e. `(.proto.mg.collection={...})` will get translated into this pdl annotation `@gma.collection = {...}`

gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/CollectionAnnotation.pdl
jsdonn: I don't think we can do it here at the model level, since all AssetLabel instances will have to have the same UpdateBehavior. Instead, I'm thinking to use the IngestionMode field of [IngestionParams](https://github.com/linkedin/datahub-gma/blob/34e0ee0f8172e2e7ebf7154e399d11917054f0e3/dao-api/src/main/pegasus/com/linkedin/metadata/internal/IngestionParams.pdl#L14) to specify what kind of update is to be performed. So something like:
```
LabelService
    -> update(new label)
             -> call assetService.update(mode=LIVE) w/ default UpdateBehavior
    -> delete(existing label)
             -> call assetService.update(mode=DELETE_FROM_COLLECTION) which will override the UpdateBehavior
```

jsdonn: or even more generic than the LabelService, maybe we can have a CollectionsService just for handling Collections-type aspects

ZihanLi58: if we do the collection, let's not have a separate service for it then. Probably you can explore the existing delete api we already have in MMG. And saying that we have default update behavior and default delete behavior etc.

gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/GmaAnnotation.pdl
ZihanLi58: If we need to process the annotation and do some customization on the annotation, can we just do it in MGA (i.e. a layer up of DAO). I think you will anyway need to check the update method and the collection type before we enter DAO layer?

jsdonn: I guess the main tradeoff for where we put the annotation processing logic is how many db operations need to be done?

if we put in MGA, we need to do 
1) get() to get the latest value of the collection so we may manipulate it before calling dao.add
2a) getLatest() as part of the dao.add
2b) saveLatest()

if we put in DAO, we just don't need the first get() so it will just be 1 read, 1 write.

but agree with you that in the second case, we will have to process the annotation in both MGA and in DAO.

I admit that the performance gained by going with this approach is insignificant, let's go just put the annotation and processing in MGA instead. thanks for the comment.

gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/CollectionAnnotation.pdl
jsdonn: Sounds good, I can also update the delete API to include IngestionParams

gradle-plugins/metadata-annotations-schema/src/main/pegasus/com/linkedin/metadata/annotations/GmaAnnotation.pdl
ZihanLi58: Oh I just saw this message, thanks for taking the suggestion. But for the part where we process the annotation on MGA, I would suggest you to call the add with preprocessFunction method to avoid the duplicate get call and make sure the update is atomic.

PR #578, opened 2025-09-05
add three new IngestionModes for collections

jsdonn:
## Summary
Add 3 ingestion modes for overriding collection aspect update behaviors:
- REPLACE_BY_KEY
- REPLACE_ALL
- DELETE_BY_KEY

The PDL schema changes were generated in metadata-models by converting protobuf files to pdl, hence the small changes to make the schemas match exactly. 

To use this, a user will define the new aspect value to contain only values for fields that they want to {replace|delete} and then specify the IngestionMode in the IngestionParams of the MCE.

Ex.
ORIGINAL:
```
OriginalMyAspect {
    ids = [1, 2, 3]
    strings = ["foo", "bar"]
}
```

REPLACE_BY_KEY:
```
NewMyAspect {
    strings = ["foobar"]
}

ResultMyAspect {
    ids = [1, 2, 3]
    strings = ["foo", "bar", "foobar"]
}
```

REPLACE_ALL:
```
NewMyAspect {
    strings = ["foobar"]
}

ResultMyAspect {
    // notice ids is null
    strings = ["foobar"]
}
```

DELETE_BY_KEY:
```
NewMyAspect {
    strings = ["foo"]
}

ResultMyAspect {
    ids = [1, 2, 3]
    strings = ["bar"]
}
```

## Testing Done
N/A
## Checklist

### DIFF
--- a/dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/events/CollectionUpdateBehavior.pdl
+++ b/dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/events/CollectionUpdateBehavior.pdl
@@ -18,7 +18,7 @@ enum CollectionUpdateBehavior {
   REPLACE_ALL
 
   /**
-   * Can only be specified when passed through IngestionParams and cannot defined in the CollectionOptions as a default
+   * Can only be specified when passed through IngestionParams and cannot be defined in the CollectionOptions as a default
    * update behavior. Delete by key is different than replace by key as it will only remove elements which share the
    * same key as the new provided aspect, rather than replace.
    */


dao-api/src/main/pegasus/com/linkedin/metadata/events/IngestionMode.pdl
```diff
@@ -27,4 +27,13 @@ enum IngestionMode {
    */
   LIVE_OVERRIDE
 
+  /**
+   * Append one or more items to one or more collections. Only applies to fields present in the new aspect.
+   */
+  APPEND_TO_COLLECTION
```
ZihanLi58: think loud here: do you think we should have separate field for it? If we put it here, how do you handle the scenario where it's backfill and happen to collection?

jsdonn: I'm not sure if we should allow backfills to perform partial updates even if it is a collection. But at the same time, agreed that these two enums should be separated into their own field in IngestionParams so that they may be combined with the other ingestion modes.

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/events/CollectionIngestionMode.pdl
```diff
@@ -0,0 +1,18 @@
+namespace pegasus.com.linkedin.metadata.events
+
+/**
+ * this ingestion parameter represents how to process the ingestion of a collection aspect
+ */
+enum CollectionIngestionMode {
+  UNKNOWN
```
jsdonn: there seems to be a bug when translated proto enums to pdl where the comment of the first enum value is not properly translated over to the pdl.

dao-api/src/main/pegasus/com/linkedin/metadata/internal/IngestionParams.pdl
```diff
@@ -2,30 +2,37 @@ namespace com.linkedin.metadata.internal
...
    */
   ingestionTrackingContext: optional IngestionTrackingContext
 
-  /*
+  /**
    * When testMode is true, the data will be persisted into {entity}_test and {relationship}_test table.
    * For validation purposes. And no MAE will be emitted during the persistance
    */
-  testMode: boolean = false
+  testMode: optional boolean = false
```
ZihanLi58, line 27: let's not add optional and with default value

dao-api/src/main/pegasus/com/linkedin/metadata/internal/IngestionParams.pdl
```diff
@@ -2,30 +2,37 @@ namespace com.linkedin.metadata.internal
...
 
   /**
    * eTag for aspect ingestion optimistic locking
    */
   ingestionETags: optional array[IngestionAspectETag] = [ ]
+
+  /**
+   * Optional parameter which if present, allows users to override the default update behavior when ingesting a collection aspect
+   */
+  collectionUpdateBehavior: optional CollectionUpdateBehavior = "UNKNOWN"
```
ZihanLi58, line 37: why do we need default value here? can we just say it's optional?

dao-api/src/main/pegasus/pegasus/com/linkedin/metadata/events/CollectionUpdateBehavior.pdl
```diff
@@ -0,0 +1,26 @@
+namespace pegasus.com.linkedin.metadata.events
+
+/**
+ * Enum that represents how to process a collection aspect update
+ */
+enum CollectionUpdateBehavior {
+  UNKNOWN
```
ZihanLi58, line 7: I don't even think we need UNKNOWN here? if you refer to photo unknown, that's automatically generated

PR #579, opened 2025-09-19
Splitting filter queries

rakhiagr:
### **Problem**

The existing listUrns query used a single SQL statement with `SELECT * , (SELECT COUNT(..)) AS _total_count`. This design caused MySQL to:
- Perform large in-memory sorts (filesort), especially when downstream passed `Integer.MAX_VALUE` as the page size.

- Fetch unnecessary columns (`SELECT *`) when only `urn` was needed.

Result: frequent Out of sort memory errors and inefficient queries.

### **Solution**

- Split queries:
-- Run a dedicated `COUNT(*)` query to compute totalCount (no sort, no limit).
-- Run a separate page query (`SELECT urn … ORDER BY urn LIMIT :pageSize OFFSET :start`) for values.

- Removed `SELECT *`:
-- Only fetch urn in page queries.

- Updated toListResult:
-- Added a new overload that accepts (values, totalCount, start, pageSize).
-- No longer depends on _total_count embedded in each SqlRow.

## Testing Done
./gradlew build
Fixed the unit tests
## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -335,7 +335,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   @Override
   public List<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion,
       @Nullable URN lastUrn, int pageSize) {
-    SqlQuery sqlQuery = createFilterSqlQuery(indexFilter, indexSortCriterion, lastUrn, pageSize);
+    SqlQuery sqlQuery = createSelectFilterSqlQuery(indexFilter, indexSortCriterion, lastUrn, pageSize);
     final List<SqlRow> sqlRows = sqlQuery.setFirstRow(0).findList();
     return sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString("urn"), _urnClass)).collect(Collectors.toList());
   }
@@ -369,7 +369,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
     }
 
     // 2. Run the paginated URN query for the current page
-    final SqlQuery pageQuery = createFilterSqlQuery(indexFilter, indexSortCriterion, start, pageSize);
+    final SqlQuery pageQuery = createSelectFilterSqlQuery(indexFilter, indexSortCriterion, start, pageSize);
     final List<SqlRow> sqlRows = pageQuery.findList();
 
     // Map the SQL rows to URN objects
@@ -509,10 +509,10 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
    * @param pageSize The maximum number of results to return.
    * @return A parametrized SqlQuery for fetching the paginated URNs.
    */
-  private SqlQuery createFilterSqlQuery(@Nullable IndexFilter indexFilter,
+  private SqlQuery createSelectFilterSqlQuery(@Nullable IndexFilter indexFilter,
       @Nullable IndexSortCriterion indexSortCriterion, int offset, int pageSize) {
     String filterSql =
-        SQLStatementUtils.createFilterSql(_entityType, indexFilter, false, _nonDollarVirtualColumnsEnabled, validator)
+        SQLStatementUtils.createSelectFilterSql(_entityType, indexFilter, _nonDollarVirtualColumnsEnabled, validator)
             + "\n" + parseSortCriteria(_entityType, indexSortCriterion, _nonDollarVirtualColumnsEnabled)
             + String.format(" LIMIT %d", Math.max(pageSize, 0)) + String.format(" OFFSET %d", Math.max(offset, 0));
     return _server.createSqlQuery(filterSql);
@@ -528,18 +528,18 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
    * @return A parametrized SqlQuery for counting matching rows.
    */
   private SqlQuery createCountSqlQuery(@Nullable IndexFilter indexFilter) {
-    String countSql = SQLStatementUtils.createFilterSql(
-        _entityType, indexFilter, true, _nonDollarVirtualColumnsEnabled, validator);
+    String countSql = SQLStatementUtils.createCountFilterSql(
+        _entityType, indexFilter, _nonDollarVirtualColumnsEnabled, validator);
     return _server.createSqlQuery(countSql);
   }
 
   /**
    * Produce {@link SqlQuery} for list urns by last urn.
    */
-  private SqlQuery createFilterSqlQuery(@Nullable IndexFilter indexFilter,
+  private SqlQuery createSelectFilterSqlQuery(@Nullable IndexFilter indexFilter,
       @Nullable IndexSortCriterion indexSortCriterion, @Nullable URN lastUrn, int pageSize) {
     StringBuilder filterSql = new StringBuilder();
-    filterSql.append(SQLStatementUtils.createFilterSql(_entityType, indexFilter, false, _nonDollarVirtualColumnsEnabled, validator));
+    filterSql.append(SQLStatementUtils.createSelectFilterSql(_entityType, indexFilter, _nonDollarVirtualColumnsEnabled, validator));
 
     if (lastUrn != null) {
       // because createFilterSql will always include a WHERE clause to filter by deleted_ts is NULL
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -391,6 +391,11 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
    * @return Resolved totalCount that accounts for potential concurrent insertions/deletions
    */
   protected int resolveTotalCount(int valuesSize, int totalCount, int start, int pageSize) {
+    // If requesting beyond available data, preserve original totalCount
+    if (start >= totalCount) {
+      return totalCount;
+    }
+    
     if (valuesSize < pageSize && start + valuesSize < totalCount) {
       // Deletion race condition detected: hit end early due to records being deleted between queries
       // Adjust totalCount to reflect the actual end position
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -377,8 +377,30 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
         sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString("urn"), _urnClass)).collect(Collectors.toList());
 
     // 3. Build and return the ListResult with values, total count, and pagination metadata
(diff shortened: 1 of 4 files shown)


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -180,11 +180,38 @@ public void testListUrnsWithOffset() {
 
     ListResult<FooUrn> listUrns = _ebeanLocalAccessFoo.listUrns(indexFilter, indexSortCriterion, 5, 5);
 
-    assertEquals(5, listUrns.getValues().size());
-    assertEquals(5, listUrns.getPageSize());
-    assertEquals(10, listUrns.getNextStart());
-    assertEquals(25, listUrns.getTotalCount());
-    assertEquals(5, listUrns.getTotalPageCount());
+    assertEquals(listUrns.getValues().size(), 5);
```
rakhiagr, line 183: Just swapped the arguments to make it `(actual, expected)`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -161,6 +161,13 @@ public class SQLStatementUtils {
    *  WHERE i_aspectfoo$value >= 25 AND i_aspectfoo$value < 50 AND JSON_EXTRACT(a_aspectfoo, '$.gma_deleted') IS NULL;
    */
   private static final String SQL_FILTER_TEMPLATE = "SELECT *, (%s) as _total_count FROM %s";
```
mridul111998: remove

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -341,18 +340,44 @@ public List<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSort
...
+    // 2. Run the paginated URN query for the current page
+    final SqlQuery pageQuery = createFilterSqlQuery(indexFilter, indexSortCriterion, start, pageSize);
+    final List<SqlRow> sqlRows = pageQuery.findList();
+
+    // Map the SQL rows to URN objects
+    final List<URN> values =
+        sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString("urn"), _urnClass)).collect(Collectors.toList());
+
+    // 3. Build and return the ListResult with values, total count, and pagination metadata
+    return toListResult(values, totalCount, start, pageSize);
```
ZihanLi58: what happens if after calculating the totalCount, we insert one more row in db before second query execute? will that be some mismatch on the real rows and total count we return?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
```diff
@@ -143,29 +145,37 @@ public void testCreateFilterSql() {
     indexCriterionArray.add(indexCriterion2);
     indexFilter.setCriteria(indexCriterionArray);
 
+    // count query with dollar sign
     String sql1 = SQLStatementUtils.createFilterSql("foo", indexFilter, true, false, mockValidator);
-    String expectedSql1 = "SELECT *, (SELECT COUNT(urn) FROM metadata_entity_foo WHERE a_aspectfoo IS NOT NULL\n"
+    String expectedSql1 = "SELECT COUNT(urn) AS total_count FROM metadata_entity_foo WHERE a_aspectfoo IS NOT NULL\n"
```
ZihanLi58, line 150: since we change * to urn, please make sure this won't break any other function rely on this

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -327,29 +334,31 @@ public static <ASPECT extends RecordTemplate> String createAspectUpdateWithOptim
...
     final String tableName = getTableName(entityType);
     String whereClause = parseIndexFilter(entityType, indexFilter, nonDollarVirtualColumnsEnabled, schemaValidator);
-    String totalCountSql = String.format("SELECT COUNT(urn) FROM %s %s", tableName, whereClause);
-    StringBuilder sb = new StringBuilder();
 
-    if (hasTotalCount) {
-      sb.append(String.format(SQL_FILTER_TEMPLATE, totalCountSql, tableName));
+    if (isCountQuery) {
+      // Build count query
+      return String.format(SQL_COUNT_TEMPLATE, tableName, whereClause);
```
ZihanLi58: Since we are basically create two template, can we just use another method for this? instead of mixing it inside the filter method?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -473,20 +498,39 @@ public Map<String, Long> countAggregate(@Nullable IndexFilter indexFilter,
...
+  /**
+   * Builds a {@link SqlQuery} for counting the number of rows matching the provided filter.
+   * Generates a query like:
+   *   SELECT COUNT(urn) AS total_count FROM table WHERE filters ...
+   * Uses the provided IndexFilter and schema validator to construct the WHERE clause.
+   *
+   * @param indexFilter The filter criteria to apply (nullable).
+   * @return A parametrized SqlQuery for counting matching rows.
+   */
+  private SqlQuery createCountSqlQuery(@Nullable IndexFilter indexFilter) {
```
ZihanLi58, line 558: why don't we need indexSortCriterion in this case?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -341,18 +340,44 @@ public List<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSort
...
-      final List<SqlRow> totalCountResults = createFilterSqlQuery(indexFilter, indexSortCriterion, 0, DEFAULT_PAGE_SIZE).findList();
-      final int actualTotalCount = totalCountResults.isEmpty() ? 0 : totalCountResults.get(0).getInteger("_total_count");
-      return toListResult(actualTotalCount, start, pageSize);
+    // 1. Run the count query to get total count of matching rows
+    final SqlQuery countQuery = createCountSqlQuery(indexFilter);
+    final SqlRow countRow = countQuery.findOne();
+    final int totalCount = (countRow == null) ? 0 : countRow.getInteger("total_count");
+
+    // If there are no matching rows, return an empty result immediately
+    if (totalCount == 0) {
```
mridul111998, line 367: Do we know how frequently we are getting count as 0. Separating this function in two queries is only helpful when most of the times we are getting empty results(0 rows). In all other cases we are adding one extra db call to first fetch the count and fetching urn's

mridul111998 (commented):
Additionally we should have a check on pagingContext, we should not allow clients to pass `Integer.MAX_VALUE` as pageSize. We should enforce a max limit on pageSize

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
rakhiagr: Yes, there is potential for a mismatch. This means the client could see more data than the totalCount suggests. Few ideas around it:

1. Our implementation already supports a `hasMore` flag (via the `havingMore` field in ListResult). This allows clients to know if additional results are available, even if the totalCount is only approximate. I am not sure if users are dependent on totalCount and it needs to be exact.

2. Run both the count and data fetch queries in a single transaction with `TRANSACTION_REPEATABLE_READ` mode, where both queries see the exact same snapshot of the database. This ensures that the totalCount and the actual rows returned are always consistent, even if new rows are inserted, updated, or deleted by other transactions during the operation. But this may hold locks longer or require more resources from the database.

Given our current use case and scale, I recommend continuing to rely on the hasMore flag and documenting that totalCount is approximate.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
rakhiagr: It's simply a count query, we don't need the sort criterion.
```sql
-- With filter only (correct for count)
SELECT COUNT(*) FROM my_table WHERE <filter conditions>;

-- Adding ORDER BY (not needed, and can slow down the query)
SELECT COUNT(*) FROM my_table WHERE <filter conditions> ORDER BY some_column;
```

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/utils/SQLStatementUtilsTest.java
rakhiagr: We should be good, as these sql queries were used only for filters and we return list of urns only, we don't need any other columns information.

rakhiagr:
> Additionally we should have a check on pagingContext, we should not allow clients to pass Integer.MAX_VALUE as pageSize. We should enforce a max limit on pageSize

Yes, that is the next plan. We are seeing errors not because of `Integer.MAX_VALUE`. The real issue is that `metadata_entity_mlmodelinstance` rows are huge (large JSON blobs). With SELECT *, MySQL had to pull all columns into memory and sort them, which caused the Out of sort memory errors. By switching to SELECT urn only, we avoid fetching those heavy blobs and keep memory usage under control.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
rakhiagr: The main issue isn’t whether the count is 0 or not. The problem is that our old query was doing `SELECT *` on `metadata_entity_mlmodelinstance`, which has very wide rows (large JSON blobs). Even though we only needed `urn`, MySQL was pulling in all columns, hold them in memory, and sort them. That’s what caused the `Out of sort memory` errors.

By splitting into two queries:

- The count query is cheap (just `COUNT(*)`, no JSON blobs, no sort).

- The page query fetches only urn, so MySQL doesn’t have to load or sort the heavy columns.

So the real win isn’t about optimizing for “count = 0”, it’s about avoiding huge payloads and memory usage from SELECT *.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
ZihanLi58: can you just set the total count to be Max(count sql result, size of list)?

rakhiagr: Discussed offline.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -336,23 +335,77 @@ public int softDeleteAsset(@Nonnull URN urn, boolean isTestMode) {
...
+   * 
+   * @param valuesSize Number of results returned from the data query
+   * @param totalCount Total count from the count query (potentially stale)
+   * @param start Starting offset for pagination
+   * @param pageSize Requested page size
+   * @return Resolved totalCount that accounts for potential concurrent insertions/deletions
+   */
+  protected int resolveTotalCount(int valuesSize, int totalCount, int start, int pageSize) {
+    // If requesting beyond available data, preserve original totalCount
+    if (start >= totalCount) {
```
ZihanLi58: start>=totalCount && value size == 0?

PR #598, opened 2026-01-23
(feat) Transactional Multi-Aspect Update

jphui:
## Impact AOE Callout

This PR only affects:

1. new pathway `batchUpsert()` / `addManyBatch()` exposed to the DAO layer
2. `AspectUpdateLambda` promoted to **public contract** between service layer (MGA) and DAO — bundles aspect class, merge lambda, and `IngestionParams`
3. `create()` pathway that uses some shared logic
4. `shouldUpdateAspect()` logic shared by all Aspect addition patterns (refactor to be more comprehensive

For the vast majority of "normal MG traffic", "core logic" is not touched here.

In addition to our incremental ramping approach, this should be considered a safer change.

## Overview

This PR introduces a new transactional batch update API that performs true batch SQL operations for multiple aspects in a single transaction, while also refactoring the existing codebase to eliminate code duplication and improve type safety through the introduction of wrapper classes.

**Two public entry points:**
- `batchUpsert(urn, List<AspectUpdateLambda>, auditStamp, trackingContext)` — **recommended**: accepts pre-built `AspectUpdateLambda` objects with custom `IngestionParams` and collection merge lambdas
- `addManyBatch(urn, List<RecordTemplate>, auditStamp, trackingContext)` — convenience wrapper that wraps raw aspects in default `AspectUpdateLambda` (LIVE mode, trivial lambda)

### 1. **New `batchUpsert()` / `addManyBatch()` API - True Batch SQL Operations**

Introduced a new batch update pathway that performs **single-query batch upserts** instead of N individual queries:

**Before (addMany):**

```java
// N separate INSERT/UPDATE queries
for (aspect : aspects) {
  INSERT INTO metadata_aspect ... ON DUPLICATE KEY UPDATE ...
}
```

**After (addManyBatch):**

```java
// Single batch INSERT with multiple rows
INSERT INTO metadata_aspect (urn, aspect, ...) VALUES
  ('urn1', 'AspectFoo', ...),
  ('urn2', 'AspectBar', ...)
ON DUPLICATE KEY UPDATE ...
```

**Performance Benefits:**

- Reduces database round-trips from N to 1
- Single transaction for atomicity
- Batch read of old values (1 query instead of N)
- Significantly faster for multi-aspect updates

### 2. **`AspectUpdateLambda` as Public Contract**

`AspectUpdateLambda` is now the **public contract** between service layer (MGA) and the DAO. Constructors are public, enabling callers to build typed tuples that bundle:

```java
public static class AspectUpdateLambda<ASPECT> {
  Class<ASPECT> aspectClass;                        // type token
  Function<Optional<ASPECT>, ASPECT> updateLambda;  // merge function (identity for simple updates, merge-by-key for collections)
  IngestionParams ingestionParams;                  // routing params (ingestion mode, test mode, collection update behavior)
}
```

**Why this was done:**
- Currently, Aspect Collection Lambdas are passed individually into each `dao.add()` operation

### 3. **Introduced AspectUpdateContext Wrapper Class**

Replaced error-prone parallel positional lists with a type-safe wrapper:

**Before:**

```java
List<RecordTemplate> aspectValues;
List<AspectUpdateLambda> aspectLambdas;
List<RecordTemplate> oldValues;
// Risk: Lists must stay aligned by position
```

**After:**

```java
class AspectUpdateContext<ASPECT> {
  @Nullable ASPECT oldValue;
  @Nonnull ASPECT newValue;  // Final value after all transformations
  @Nonnull AspectUpdateLambda<ASPECT> lambda;
}
List<AspectUpdateContext<RecordTemplate>> contexts;
// Guaranteed alignment by construction
```

**Benefits:**

- Eliminates positional alignment bugs
- Type-safe pairing of related data
- Clearer API contracts
- Self-documenting code
- `newValue` contains the final computed value after all transformations (lambda application, callback processing,
  lambda function registry)

_**NOTE: My**_ decision here (don't worry I actually changed what AI was trying to get me to do) was motivated by trying to shrink /
prevent method signatures from getting too large:

- if we added 'oldValue' as a new parameter entirely, then we have to expand the callee's (and its callees') signatures
- if we add 'oldValue' as a parameter for AspectUpdateLambda, this semantically doesn't work because AspectUpdateLambdas
  are created super high up in the call chain (in fact there are public methods that can take them as input) --> it's
  odd to create them knowingly that a field will be 'null' until populated (at least my view is that)
- if you look at the preceding logic, you'll notice we are doing list processing, and a previous design iteration had
  List, List or something like that where basically you had to ensure that positional alignment was maintained --> this
  takes on the risk that some future refactor doesn't know this ==> I introduce AspectUpdateContext which contains the
  lambda, oldValue, and newValue (the final transformed value)

### 4. **Unified Logic Between `create()` and `batchUpsert()` Pathways**

Refactored to share common logic:

- Both use AspectUpdateContext for consistency
- Shared `batchUpsertAspects()` implementation
- Shared SQL building logic (`prepareMultiColumnInsert()`)
- Reduced code duplication by ~200 lines

**Key Shared Components:**

- Batch old value retrieval
- Equality checking and filtering
- Pre/post update hooks
- Callback execution
- Relationship ingestion
- MAE emission logic

## E2E Testing

- `/test-gma-changes` used to publish SNAPSHOT and propagate through metadata-models dependency chain
- MGA PR: https://github.com/linkedin-multiproduct/metadata-graph-assets/pull/830
- Deploy: `mint build && mint build-cfg -f qei-ltx1 && mint deploy -f qei-ltx1 --debug-app`
- test1: batch upsert with multiple regular aspects (Deprecation + InferredStatus) https://paste.corp.linkedin.com/show/72018148/
- test2: batch upsert verifying gridTable dual-write Status hook excluded from gridTable variant https://paste.corp.linkedin.com/show/72018150/
- test3: collection aspect (AssetLabels) merge-by-key via batch path — 3-call sequence validating REPLACE_BY_KEY semantics https://paste.corp.linkedin.com/show/72018174/
- test4: log output confirming `[BatchUpsert]` pathway taken https://paste.corp.linkedin.com/show/72018152/
- test5: non-Dataset entity update confirming old per-aspect pathway, no regressions https://paste.corp.linkedin.com/show/72018198/

### **BaseLocalDAOTest - 8 New Orchestration Tests**

Added tests for `batchUpsertAspects()`.

#### **MAE Emission & Equality Tests**

- ✅ `testAddManyBatchMAEEmission` - Verifies MAE emitted for new aspects
- ✅ `testAddManyBatchMAEEmissionWithEqualitySkip` - **Critical:** Verifies MAE NOT emitted when `oldValue == newValue` (prevents event spam)
- ✅ `testAddManyBatchReturnsAllAspects` - All aspects returned even if some skipped

#### **Selective Insertion Tests**

- ✅ Verifies batch API only persists aspects that pass equality checks
- ✅ Confirms no-op batches when all aspects equal (true no-op, no MAE spam)

#### **Hook Tests**

- ✅ `testAddManyBatchPreUpdateHook` - Pre-update hooks fire for batch
- ✅ `testAddManyBatchPostUpdateHook` - Post-update hooks fire for batch

#### **AspectCallbackRegistry Tests**

Tests verifying that `AspectCallbackRegistry` callbacks are correctly invoked in the batch update pathway:

- ✅ `testAddManyBatchWithSingleAspectCallback` - Single aspect with callback that transforms value (`foo` → `bar`),
  verifies persistence via `getLatest()`
- ✅ `testAddManyBatchWithSingleAspectCallbackSkipped` - Single aspect with callback that returns
  `isSkipProcessing=true` - verifies no MAE emitted and empty results
- ✅ `testAddManyBatchWithTwoAspectsWithCallbacksNeitherSkipped` - Two aspects, both have callbacks, both transform
  values correctly, verifies persistence via `getLatest()`
- ✅ `testAddManyBatchWithTwoAspectsWithCallbacksOneSkipped` - Two aspects with callbacks, one skipped - only
  non-skipped aspect emits MAE, verifies persistence via `getLatest()`
- ✅ `testAddManyBatchWithMixedCallbackRegistration` - Two aspects, only one has callback registered - callback
  transforms its aspect, other passes through unchanged, verifies persistence via `getLatest()`

#### **LambdaFunctionRegistry Tests**

Tests verifying that `LambdaFunctionRegistry` lambda functions are correctly invoked in the batch update pathway:

- ✅ `testAddManyBatchWithSingleAspectLambda` - Single aspect with lambda that transforms value, verifies persistence
  via `getLatest()`
- ✅ `testAddManyBatchWithMixedLambdaRegistration` - Two aspects, only one has lambda registered - lambda transforms its
  aspect, other passes through unchanged, verifies persistence via `getLatest()`
- ✅ `testAddManyBatchWithLambdaMergingOldValue` - Lambda that merges old and new values, verifies the lambda receives
  the old value correctly
- ✅ `testAddManyBatchWithLambdaAndEqualitySkip` - Aspect with lambda where transformed value equals existing - verifies
  equality skip still works after lambda transformation

#### **Comprehensive Integration Test**

- ✅ `testAddManyBatchComprehensiveWithAllCodePaths` - Tests 4 aspects simultaneously exercising different code paths:
  - **AspectFoo**: Equality skip (same value, verified via unchanged timestamp)
  - **AspectBar**: FORCE_UPDATE annotation override (same value but write forced due to `@gma.aspect.ingestion`
    annotation, verified via changed timestamp)
  - **AspectFooBar**: Callback transformation before ingestion
  - **AspectAttributes**: Lambda transformation before ingestion

#### **Transaction & Validation Tests**

- ✅ `testAddManyBatchTransactionBehavior` - Single transaction for entire batch
- ✅ `testAddManyBatchRejectsDuplicateAspects` - Duplicate aspect classes rejected
- ✅ `testAddManyBatchWithTrackingContext` - Tracking context properly propagated

### **EbeanLocalAccessTest - 11 New SQL-Level Tests**

Added comprehensive testing for batch SQL operations:

- ✅ `testBatchUpsert` - Basic batch upsert functionality
- ✅ `testBatchUpsertMultipleAspects` - Multiple aspects in single batch
- ✅ `testBatchUpsertWithExistingAspects` - Updates existing records
- ✅ `testBatchGetLatestAspectsWithExtraInfo` - Batch read operations
- ✅ `testBatchUpsertEmptyList` - Edge case handling
- ✅ Plus 6 more tests for various scenarios

### **EbeanLocalDAOTest - Relationship Ingestion Tests**

- ✅ `testAddManyBatchMultipleAspectsWithRelationshipsInSingleBatch` -
  **Fixed test bug:** Relationships ARE created correctly; test was using `.toString()` on union type instead of
  `.getString()`
- ✅ Verifies relationships created when aspects added
- ✅ Verifies relationships NOT created when aspects skipped due to equality
- ✅ Tests for both `BelongsTo` and `BelongsToV2` relationship types

## **Backward Compatibility**

- All existing APIs maintained
- No breaking changes to public interfaces
- Existing tests continue to pass

## Next Steps

1. ~~bumps to propagate to _metadata-graph-assets_~~ ✅ Done via `/test-gma-changes`
2. ~~Add new config-gated logic pathway to call the new API~~ ✅ Done in MGA PR (Dataset entity type gate)
3. ~~E2E testing on QEI~~ ✅ Verified batch upsert, collection aspects, dual-write, IngestionParams propagation
4. Fix 6 failing unit tests in MGA `DatasetAssetServiceImplTest` (tests mock `dao.add()` but batch path calls `batchUpsert()`)
5. Ramp incrementally to additional entity types

## Checklist


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -649,6 +622,184 @@ public List<ASPECT_UNION> addMany(@Nonnull URN urn, @Nonnull List<? extends Reco
...
+      Optional oldValue = Optional.ofNullable(oldAspect);
+      ExtraInfo oldExtraInfo = oldInfo != null ? oldInfo.getExtraInfo() : null;
+      AuditStamp oldAuditStamp = oldExtraInfo != null ? oldExtraInfo.getAudit() : null;
+      Long oldEmitTime = oldExtraInfo != null ? oldExtraInfo.getEmitTime() : null;
+
+      // Execute lambda WITH old value (not Optional.empty())
+      RecordTemplate newValue = (RecordTemplate) updateLambda.getUpdateLambda().apply(oldValue);
+
+      // Equality testing & backfill logic using unified shouldUpdateAspect
+      EqualityTester<RecordTemplate> equalityTester = getEqualityTester(aspectClass);
```
arvindpruthi, line 762: Why is Equality Testing important? My understanding of the logic is that the aspectUpdateLambdas carries the stream of updates for a given row, not the entire row. So if there were indeed unchanged aspects that the user requested to write; what are we really saving with this logic? Why not just carry the user intent and keep things simpler.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -649,6 +622,184 @@ public List<ASPECT_UNION> addMany(@Nonnull URN urn, @Nonnull List<? extends Reco
...
+   * Batch upsert with AspectUpdateLambda support (enables test mode and ingestion params).
+   *
+   * <p>Note: Transaction retry is handled by the subclass implementation (e.g., EbeanLocalDAO),
+   * not at this layer. The maxTransactionRetry parameter is kept for API compatibility but is
+   * currently not used. Transaction handling follows the same pattern as createNewAssetWithAspects().
+   *
+   * <p>NOTE: This is structurally very similar to createAspectsWithCallbacks(), some future refactoring can be done to
+   * share the logic between the two.
+   */
+  List<ASPECT_UNION> addManyBatchWithCallbacks(@Nonnull URN urn,
```
arvindpruthi: You may need to add a check against duplicate aspects. What if the caller includes 2 copies (updates) of the same aspect in the batch. The current logic may lead to last-rite-wins without any explicit guarantee. Suggestion: Reject duplicates or apply lambdas sequentially per aspect class where the output of the first becomes the “old” for the next, and only final value is written once.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -649,6 +622,184 @@ public List<ASPECT_UNION> addMany(@Nonnull URN urn, @Nonnull List<? extends Reco
...
+      results.add(result);
+    }
+
+    return results;
+  }
+
+  /**
+   * Helper class to hold processed aspect data.
+   */
+  private static class ProcessedAspect {
```
arvindpruthi: ProcessedAspect class is dead code. It’s not used anywhere.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -287,6 +287,139 @@ public <ASPECT_UNION extends RecordTemplate> int create(
...
+   *
+   * @param urn entity URN
+   * @param aspectValues list of aspect values to upsert
+   * @param aspectUpdateLambdas list of aspect update lambdas to upsert
+   * @param auditStamp audit stamp for tracking
+   * @param ingestionTrackingContext tracking context for ingestion
+   * @param isTestMode whether this is a test mode operation
+   * @return number of rows affected
+   */
+  public <ASPECT_UNION extends RecordTemplate> int batchUpsert(
```
arvindpruthi, line 218: There is a lot of commonality between the code in create() and this function. Pls see if some of the common code can be extracted out into a helper function.

arvindpruthi (commented):
I think the philosophy followed in this PR is the right approach to cut multiple SQL updates. Left a few review comments.

Also the overall change is complex enough that we may need to roll it out cautiously.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: MG policy has always been that if the content of the Aspect is the same, then a write will be abstained to save a DB write operation.

Without further analysis into seeing how much DB load we're saving and perhaps if there is existing functionality that relies on this feature, I'm hesitant to change it atm as a part of the scope here.

If anything I can add a TODO / note in the code that points out the potential for improvement here?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: This is a great comment, I checked the current implementation and while most strongly-typed customers (java clients) will not have an issue with this, JSON-based users (grpcurli) would.

I think it's best to reject duplicates, will call this out in the next commit 👍

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -545,9 +539,9 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
 
     final AuditStamp optimisticLockAuditStamp = extractOptimisticLockForAspectFromIngestionParamsIfPossible(ingestionParams, aspectClass, urn);
 
-    // Logic determines whether an update to aspect should be persisted.
+    // Unified logic determines whether an update to aspect should be persisted (includes backfill, equality, version, mode checks)
     if (!shouldUpdateAspect(ingestionParams.getIngestionMode(), urn, oldValue, newValue, aspectClass, auditStamp, equalityTester,
-        oldAuditStamp, optimisticLockAuditStamp)) {
+        oldAuditStamp, optimisticLockAuditStamp, trackingContext, oldEmitTime)) {
```
ZihanLi58, line 549: why we need to add tow more parameters here?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -649,6 +643,180 @@ public List<ASPECT_UNION> addMany(@Nonnull URN urn, @Nonnull List<? extends Reco
...
+      contextsToWrite.add(new AspectUpdateContext<>(oldAspect, (AspectUpdateLambda<RecordTemplate>) updateLambda));
+    }
+
+    // STEP 3: Execute batch SQL (1 query) - only for changed aspects
+    if (!contextsToWrite.isEmpty()) {
+      boolean isTestMode = contextsToWrite.stream()
+          .anyMatch(ctx -> ctx.getLambda().getIngestionParams().isTestMode());
+      
+      // NOTE: basically if an aspect appears here then its relationships are meant to be written (as well)
+      batchUpsertAspects(urn, contextsToWrite, auditStamp, trackingContext, isTestMode);
```
ZihanLi58, line 808: this does not seem correct to me, if you only pass the old value into the ebeanDao to update, then where do we use the newValue?

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -2007,10 +2191,33 @@ protected boolean aspectTimestampSkipWrite(@Nullable AuditStamp eTagValue, @Null
...
+   * Includes backfill logic, equality testing, version checking, and ingestion mode handling.
    */
   private <ASPECT extends RecordTemplate> boolean shouldUpdateAspect(IngestionMode ingestionMode, URN urn, ASPECT oldValue,
       ASPECT newValue, Class<ASPECT> aspectClass, AuditStamp auditStamp, EqualityTester<ASPECT> equalityTester,
-      AuditStamp oldValueAuditStamp, AuditStamp eTagAuditStamp) {
+      AuditStamp oldValueAuditStamp, AuditStamp eTagAuditStamp, @Nullable IngestionTrackingContext trackingContext,
+      @Nullable Long oldEmitTime) {
+
+    // Backfill check - if this is a backfill event, verify it's newer than existing data
+    final boolean isBackfillEvent = trackingContext != null
```
ZihanLi58, line 2251: 1. can you make sure whether we have this check before? I remember that some test I did before see that we do skip the backfill event correctly. 
2. Can you check and make sure this is the correct way for us to mark the event as backfill?

ZihanLi58 (commented):
Can you add more test for the call back situation? and since this is the core function of our ingestion logic, I will require a thorough test and roll out plan to make sure we don't break anything here.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: If you look at what was moved "in" from _**addCommon()**_ which is the pathway that all single aspect updates go through:

```java
    final boolean isBackfillEvent = trackingContext != null
        && trackingContext.hasBackfill() && trackingContext.isBackfill();
    if (isBackfillEvent) {
      boolean shouldBackfill =
          // new value is being inserted. We should backfill
          oldValue == null
              || (
              // tracking context should ideally always have emitTime. If it's not present, we will skip backfilling
              trackingContext.hasEmitTime()
                  && (
                  // old emit time is available so we'll use it for comparison
                  // if new event emit time > old event emit time, we'll backfill
                  (oldEmitTime != null && trackingContext.getEmitTime() > oldEmitTime)
                      // old emit time is not available, so we'll fall back to comparing new emit time against old audit time
                      // old audit time represents the last modified time of the aspect
                      || (oldEmitTime == null && oldAuditStamp != null && oldAuditStamp.hasTime() && trackingContext.getEmitTime() > oldAuditStamp.getTime())));

      log.info("Encounter backfill event. Old value = null: {}. Tracking context: {}. Urn: {}. Aspect class: {}. Old audit stamp: {}. "
              + "Old emit time: {}. "
              + "Based on this information, shouldBackfill = {}.",
          oldValue == null, trackingContext, urn, aspectClass, oldAuditStamp, oldEmitTime, shouldBackfill);

      if (!shouldBackfill) {
        return new AddResult<>(oldValue, oldValue, aspectClass);
      }
    }
```

This is just moved into the _**shouldUpdateAspect()**_ method to consolidate this logic and make it easier to read and avoid duplication.

This is logically sound because if for a backfill event, if we "shouldn't backfill" we shouldn't be performing any DB operations (which is exactly what the _shouldUpdateAspect()_ method does to begin with) -- even the return value is the same in that logic pathway:

```java
return new AddResult<>(oldValue, oldValue, aspectClass);
```

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: Does my previous reply 2 comments up answer this?

This is NOT a new check, just moved over to my "unified / consolidated" method to make it easier to read and centralize the "should something happen" call.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
jphui: I think you're referring to how I construct `contextsToWrite`:

```java
contextsToWrite.add(new AspectUpdateContext<>(oldAspect, (AspectUpdateLambda<RecordTemplate>) updateLambda));
```

and then pass it into:

```java
batchUpsertAspects(urn, contextsToWrite, auditStamp, trackingContext, isTestMode);
```

......

It might look like I'm not passing in the Old Value, but I actually am:

```java
    AspectUpdateLambda(ASPECT value) {
      this.aspectClass = (Class<ASPECT>) value.getClass();
      this.updateLambda = (ignored) -> value;
      this.ingestionParams = new IngestionParams().setIngestionMode(IngestionMode.LIVE);
    }
```

Note that when an AspectUpdateLambda is created with JUST a 'value' the lambda itself "becomes" that value.

This is also noted a few lines above:

```java
      // NOTE: if the lambda is "empty" then it will return the value to be written (newValue) since that is how
      // AspectUpdateLambda is defined and constructed
      RecordTemplate newValue = (RecordTemplate) updateLambda.getUpdateLambda().apply(oldValue);
```

In short, newValue is a derivative of "apply the lambda to the old value" ==> any time we have both, we have the correct newValue on hand.


.....


My decision here (don't worry I actually changed what AI was trying to get me to do) was motivated by trying to shrink / prevent method signatures from getting too large:

1. if we added 'oldValue' as a new parameter entirely, then we have to expand the callee's (and its callees') signatures
2. if we add 'oldValue' as a parameter for AspectUpdateLambda, this semantically doesn't work because AspectUpdateLambdas are created super high up in the call chain (in fact there are public methods that can take them as input) --> it's odd to create them knowingly that a field will be 'null' until populated (at least my view is that)
3. if you look at the preceding logic, you'll notice we are doing list processing, and a previous design iteration had List<OldValue>, List<NewValue> or something like that where basically you had to ensure that positional alignment was maintained --> this takes on the risk that some future refactor doesn't know this

==> I introduce AspectUpdateContext which contains BOTH the lambda + oldValue


.....


I will update the PR description to make this^^ more obvious and see what opportunities there are to also update the code comments as well.

arvindpruthi (approved):
Lgtm

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
ZihanLi58: the new value here I mean the output of the call back, but I don'e see you use it anywhere other than the return value. Am I missing anything here?

jphui:
AI Review: https://docs.google.com/document/d/1PL4b6aIcrpBMKvu1TwlmW223rfovyy5KaXDjlDz3mHY/edit?tab=t.0#heading=h.cajy1010xsbt

jphui:
>   Potential concern with batch upsert + callbacks
> 
>   The agent found something worth noting: your batch path calls dao.addManyBatch() which does not pass pegasusParams
>   (IngestionParams), while the per-aspect path passes it to dao.add(). Some routing clients like
>   StatusCallbackRoutingClient have a 5-parameter callback signature that requires IngestionParams. If callbacks fire
>   during the batch path, they might not get IngestionParams.

jphui:
Closing since we are splitting it up.

PR #602, opened 2026-03-05
Fix backfill gaps for soft-deleted aspects

rakhiagr:
## Summary
Fix two backfill logic gaps that allow stale events to resurrect soft-deleted aspects.

**RFC:** https://docs.google.com/document/d/1zOKrxbm_zGxTpMwJ7Wer7owY-YcI0nXDvY6DWRGXj0M/edit?tab=t.0

**Scope:** Aspect-level soft deletion on new schema only. Old schema and asset-level deletion (`deleted_ts`) are out of scope for this PR.

### Problem
All these gaps are reproduced in EI and captured in this doc: [Backfill Gaps in EI ](https://docs.google.com/document/d/17XNuwhBBanPuT5w5Bo-40ozN5bhTSSqyHxSTYaiXybs/edit?tab=t.0)
### Gap 1: `shouldBackfill()` treats soft-deleted aspects as "never existed"

**Problem:** When a backfill event arrives for a soft-deleted aspect, `oldValue == null` caused the code to short-circuit to `shouldBackfill=true` without checking `isSoftDeleted`. Stale DLQ replays could silently resurrect deleted aspects.

**Fix:** Replace the single boolean expression with a three-way branch in `shouldBackfill()`:

  | Condition | Behavior |
  |-----------|----------|
  | `oldValue == null && !isSoftDeleted` | Aspect never existed — allow backfill unconditionally |
  | `oldValue == null && isSoftDeleted` | Compare `emitTime > deletionTimestamp`; only allow if newer |
  | `oldValue != null` | Existing emitTime comparison(unchanged) |

Additionally: reject backfill if `emitTime` is missing (cannot assess staleness), and log a warning if deletion timestamp is missing/invalid.

### Gap 2: Entity-level `lastmodifiedon` used as deletion timestamp (timestamp corruption)

  **Problem:** The entity table has a single `lastmodifiedon` column shared across ALL aspects. When aspect A is deleted at T1, then aspect B is updated at T2, the `lastmodifiedon` bumps to T2. The Gap 1 fix compares `emitTime > oldAuditStamp.getTime()`, but `oldAuditStamp` was reading from this shared column — returning T2 instead of T1.
The soft-delete marker `{"gma_deleted":true}` had no embedded timestamp, so there was no per-aspect deletion time to compare against.

**Fix:** Enrich the soft-delete marker JSON with per-aspect deletion metadata:

  ```json
  // Before (legacy)
  {"gma_deleted": true}

  // After (enriched)
  {"gma_deleted": true, "deleted_timestamp": "2026-03-06 18:41:59.000", "deleted_by": "urn:li:corpuser:actor"}

```

  Write side (EbeanLocalAccess.addWithOptimisticLocking): When soft-deleting, build enriched JSON with deleted_timestamp and deleted_by instead of the bare marker.

  Read side (EBeanDAOUtils.readSqlRow): For soft-deleted aspects, extract deleted_timestamp from the enriched JSON. Falls back to entity-level lastmodifiedon for legacy rows that only have {"gma_deleted":true}.

  Detection (EBeanDAOUtils.isSoftDeletedAspect): Changed from exact string equals to parsing-based check (gma_deleted == true), so enriched JSON with extra fields is still correctly identified as soft-deleted.
 
Backward Compatibility

  - Legacy rows with {"gma_deleted":true} continue to work — detection falls back to entity-level lastmodifiedon (same behavior as before)
  - New fields in SoftDeletedAspect.pdl are optional — no schema migration required
  - Old schema paths (metadata_aspect table, .ne(METADATA_COLUMN, DELETED_VALUE) filters) are unchanged
  
### Testing Done

Screenshot 2026-03-06 at 4 11 28 PM

  Unit tests added:
  - testIsSoftDeletedAspectWithEnrichedJson — enriched JSON detected by SqlRow-based check
  - testIsSoftDeletedAspectOnEbeanMetadataAspectWithEnrichedJson — both legacy and enriched detected
  - testBuildDeletedValue — enriched JSON contains timestamp and actor
  - testIsSoftDeletedMetadata — handles legacy, enriched, non-deleted, null, empty
  - testSoftDeleteWritesEnrichedJson — end-to-end: soft-delete writes enriched JSON to DB

  Existing tests (from prior commits):
  - testBackfillNoopWhenSoftDeletedAndEmitTimeIsStale
  - testBackfillSucceedsWhenSoftDeletedAndEmitTimeIsNewer
  - testBackfillNoopWhenSoftDeletedAndNoEmitTime
  - testBackfillNoopWhenSoftDeletedAndEmitTimeEqualsDeletionTime
  - testBackfillNoopWhenSoftDeletedAndMissingDeletionTimestamp

### Tested snapshot in MGA

All the gaps are fixed, tested by locally deploying MGA with this snapshot version. All results captured in this doc:
[Backfill gaps fixed](https://docs.google.com/document/d/17XNuwhBBanPuT5w5Bo-40ozN5bhTSSqyHxSTYaiXybs/edit?tab=t.0)

 ### Known Remaining Gap: Asset-Level Delete (follow-up PR)

  Entity-level deletes (deleteAll → softDeleteAsset) set deleted_ts=NOW() on the entity row. batchGetUnion() filters deleted_ts
  IS NULL, hiding these rows entirely — making them indistinguishable from "never existed" and bypassing the isSoftDeleted
  check. Will be addressed in a separate PR.

## Checklist

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -510,24 +510,32 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     final boolean isBackfillEvent = trackingContext != null
         && trackingContext.hasBackfill() && trackingContext.isBackfill();
     if (isBackfillEvent) {
-      boolean shouldBackfill =
-          // new value is being inserted. We should backfill
-          oldValue == null
-              || (
-              // tracking context should ideally always have emitTime. If it's not present, we will skip backfilling
-              trackingContext.hasEmitTime()
-                  && (
-                  // old emit time is available so we'll use it for comparison
-                  // if new event emit time > old event emit time, we'll backfill
-                  (oldEmitTime != null && trackingContext.getEmitTime() > oldEmitTime)
-                      // old emit time is not available, so we'll fall back to comparing new emit time against old audit time
-                      // old audit time represents the last modified time of the aspect
-                      || (oldEmitTime == null && oldAuditStamp != null && oldAuditStamp.hasTime() && trackingContext.getEmitTime() > oldAuditStamp.getTime())));
+      boolean shouldBackfill;
+      if (oldValue == null && !latest.isSoftDeleted) {
+        // Aspect has never existed — safe to backfill unconditionally
+        shouldBackfill = true;
+      } else if (oldValue == null && latest.isSoftDeleted) {
+        // Aspect/entity was soft-deleted (oldValue is null but deletion occurred).
+        // Compare emitTime against the deletion timestamp (oldAuditStamp.time) to prevent
+        // stale backfill events from resurrecting soft-deleted entities.
+        shouldBackfill = trackingContext.hasEmitTime()
+            && oldAuditStamp != null && oldAuditStamp.hasTime()
+            && trackingContext.getEmitTime() >= oldAuditStamp.getTime();
+      } else {
+        // oldValue exists (not soft-deleted) — normal emitTime comparison
+        shouldBackfill =
+            trackingContext.hasEmitTime()
+                && (
+                (oldEmitTime != null && trackingContext.getEmitTime() > oldEmitTime)
+                    || (oldEmitTime == null && oldAuditStamp != null && oldAuditStamp.hasTime()
+                    && trackingContext.getEmitTime() > oldAuditStamp.getTime()));
+      }
 
-      log.info("Encounter backfill event. Old value = null: {}. Tracking context: {}. Urn: {}. Aspect class: {}. Old audit stamp: {}. "
-              + "Old emit time: {}. "
+      log.info("Encounter backfill event. Old value = null: {}. isSoftDeleted: {}. Tracking context: {}. Urn: {}. "
+              + "Aspect class: {}. Old audit stamp: {}. Old emit time: {}. "
               + "Based on this information, shouldBackfill = {}.",
-          oldValue == null, trackingContext, urn, aspectClass, oldAuditStamp, oldEmitTime, shouldBackfill);
+          oldValue == null, latest.isSoftDeleted, trackingContext, urn, aspectClass, oldAuditStamp, oldEmitTime,
+          shouldBackfill);
 
       if (!shouldBackfill) {
         return new AddResult<>(oldValue, oldValue, aspectClass);
@@ -629,7 +637,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     } else {
       // no atomic multiple updates: run each in its own transaction. This is the same as repeated calls to add
       results = aspectUpdateLambdas.stream().map(x -> runInTransactionWithRetry(() ->
-              aspectUpdateHelper(urn, x, auditStamp, trackingContext), maxTransactionRetry)).collect(Collectors.toList());
+          aspectUpdateHelper(urn, x, auditStamp, trackingContext), maxTransactionRetry)).collect(Collectors.toList());
     }
 
     // send the audit events etc
@@ -960,8 +968,8 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
     // create aspects and process callbacks in a single transaction
     return runInTransactionWithRetry(() -> {
-      return createAspectsWithCallbacks(urn, aspectValues, aspectCreateLambdas, auditStamp, trackingContext);
-      }, maxTransactionRetry
+          return createAspectsWithCallbacks(urn, aspectValues, aspectCreateLambdas, auditStamp, trackingContext);
+        }, maxTransactionRetry
     );
   }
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java [hunk 1]
```diff
@@ -503,37 +503,22 @@ private <ASPECT extends RecordTemplate> AddResult<ASPECT> addCommon(@Nonnull URN
...
-
-      log.info("Encounter backfill event. Old value = null: {}. Tracking context: {}. Urn: {}. Aspect class: {}. Old audit stamp: {}. "
-              + "Old emit time: {}. "
+      boolean shouldBackfill = shouldBackfill(urn, latest, aspectClass, trackingContext);
+      log.info("Encounter backfill event. Old value = null: {}. isSoftDeleted: {}. Tracking context: {}. Urn: {}. "
+              + "Aspect class: {}. Old audit stamp: {}. Old emit time: {}. "
               + "Based on this information, shouldBackfill = {}.",
-          oldValue == null, trackingContext, urn, aspectClass, oldAuditStamp, oldEmitTime, shouldBackfill);
-
+          oldValue == null, latest.isSoftDeleted, trackingContext, urn, aspectClass, oldAuditStamp, oldEmitTime, shouldBackfill);
```
jphui: 1. (Medium) latest.isSoftDeleted accessed as field, not getter — BaseLocalDAO.java:539

  In the log statement on the new line 539 (diff line 39):
  oldValue == null, latest.isSoftDeleted, trackingContext, ...
  AspectEntry is annotated with @Data (Lombok), which generates isSoftDeleted(). Direct field access works since it's package-private, but it's inconsistent with how
  latest.getAspect() and latest.getExtraInfo() are accessed elsewhere. Same issue in shouldBackfill() at line 567:
  if (oldValue == null && !latest.isSoftDeleted) {
  Recommend using latest.isSoftDeleted() for consistency.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -65,9 +65,38 @@ public class EBeanDAOUtils {
...
+    softDeletedAspect.setDeleted_timestamp(deletedTimestamp);
+    softDeletedAspect.setDeleted_by(deletedBy);
+    return RecordUtils.toJsonString(softDeletedAspect);
+  }
+
+  /**
+   * Check if a metadata JSON string represents a soft-deleted aspect.
+   * Handles both legacy format {"gma_deleted":true} and enriched format with extra fields.
+   */
+  public static boolean isSoftDeletedMetadata(@Nullable String metadata) {
```
jphui, line 100: 2. (Medium) isSoftDeletedMetadata swallows all exceptions silently — EBeanDAOUtils.java

  public static boolean isSoftDeletedMetadata(@Nullable String metadata) {
      ...
      try {
        SoftDeletedAspect aspect = RecordUtils.toRecordTemplate(SoftDeletedAspect.class, metadata);
        return aspect.hasGma_deleted() && aspect.isGma_deleted();
      } catch (Exception e) {
        return false;
      }
  }
  Catching all exceptions and returning false means corrupted/malformed JSON is silently treated as "not deleted." This could mask real data issues. Consider logging at
  debug/trace level so corruption is at least observable.

validators/src/main/pegasus/com/linkedin/metadata/aspect/SoftDeletedAspect.pdl
```diff
@@ -9,4 +9,14 @@ record SoftDeletedAspect {
    * Flag to indicate if the aspect is soft deleted in GMA
    */
   gma_deleted: boolean
+
+  /**
+   * Timestamp of when the aspect was soft deleted (string format matching DB lastmodifiedon column, e.g. "2026-03-06 18:41:59.000")
+   */
+  deleted_timestamp: optional string
```
jphui: Is there a reason we want to do String instead of a long (epoch?)

jphui (commented):
(AI review first)

Can you add a .md that tracks this feature change?

Just ask Claude to summarize the change in like 10-20 sentences and add it for easy readability :)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -65,9 +65,38 @@ public class EBeanDAOUtils {
       .toFormatter();
   public static final String DIFFERENT_RESULTS_TEMPLATE = "The results of %s from the new schema table and old schema table are not equal. Reason: %s. "
       + "Defaulting to using the value(s) from the old schema table.";
-  // String stored in metadata_aspect table for soft deleted aspect
+  // Bare soft-delete marker (legacy format, used only for detection/filtering of old rows)
   private static final RecordTemplate DELETED_METADATA = new SoftDeletedAspect().setGma_deleted(true);
   public static final String DELETED_VALUE = RecordUtils.toJsonString(DELETED_METADATA);
```
jphui, line 81: > Would it be smart to add a note on DELETED_VALUE + DELETED_METADATA to say that it is NOT VALID to use this anymore in new code?

● Yes, absolutely. DELETED_VALUE is now only valid for old-schema SQL filters (where the old schema still writes bare JSON). Any new code should use isSoftDeletedMetadata() for
  detection and buildDeletedValue() for writing. A deprecation note on the constant would be the right guardrail.

validators/src/main/pegasus/com/linkedin/metadata/aspect/SoftDeletedAspect.pdl
rakhiagr: Good call — no strong reason for String, it was just matching the DB lastmodifiedon column format. Changed to long (epoch millis) to stay consistent with emitTime and AuditStamp.time, and to avoid the unnecessary long → String → Timestamp → long conversion chain.

jphui (approved):

PR #606, opened 2026-04-02
[BatchUpdate 1/3] refactor(dao-api): promote wrapper types and unify shouldUpdateAspect

jphui:
## Overview

Extracted from #598 for easier review. This PR contains only the **type refactoring and shouldUpdateAspect unification** — no new batch upsert logic.

### 1. `AspectUpdateLambda` — public contract between service and DAO

- Replaced `@AllArgsConstructor @Value` with `@Getter` + explicit constructors
- Made all constructors **public** so MGA can build lambdas with custom `IngestionParams` and collection merge functions
- Added 3-arg constructor accepting `(Class, Function, IngestionParams)`
- Fields changed to `protected final` for subclass access

### 2. `AspectCreateLambda` — now extends `AspectUpdateLambda`

- Create is a special case of update where the lambda ignores old value
- Eliminates duplicated fields (`aspectClass`, `ingestionParams`)
- Constructor delegates to `super(value)`

### 3. `AspectUpdateContext` — new inner class

- Bundles `oldValue`, `newValue`, and `AspectUpdateLambda` in a single object
- Eliminates error-prone parallel positional lists (e.g., `List<oldValues>` aligned with `List<lambdas>`)
- Used by batch operations in #598

### 4. `shouldUpdateAspect()` — unified skip-or-write logic

- Absorbed backfill logic that was previously inline in `addCommon()`
- Added `trackingContext` and `oldEmitTime` parameters
- Single method now handles: backfill checks, equality, version skip, timestamp skip, `FORCE_UPDATE` annotation, `LIVE_OVERRIDE` mode
- Reusable by both single-aspect `add()` and batch `batchUpsert()` paths

## Testing

- All existing `dao-api` tests pass
- No behavioral change to the `add()` / `addCommon()` pathway — backfill logic is moved, not modified
- The new types (`AspectUpdateContext`) are unused until #598 lands

## Checklist

- [x] Compiles with `./gradlew :dao-api:compileJava`
- [x] Tests pass with `./gradlew :dao-api:test`
- [x] No breaking changes to public interfaces

PR #607, opened 2026-04-06
[BatchUpdate 2/3] feat: add batch SQL layer for multi-aspect upsert and batch deletion

jphui:
## Overview

Extracted from #598 for easier review (part 2 of 4, depends on #606). This PR contains only the **SQL-level batch operations** — no orchestration logic in `BaseLocalDAO` or `EbeanLocalDAO`.

Note that upon inspection, the comparison between this PR and the relevant sections of 598 are identical 😎 

### 1. `EbeanLocalAccess.batchUpsert()` — multi-aspect upsert in a single SQL call

- Generates one `INSERT ... ON DUPLICATE KEY UPDATE` that writes all aspects for a URN at once
- Uses UPSERT semantics: always updates if exists, clears `deleted_ts`
- Contrast with `create()` which throws `DuplicateKeyException` on non-soft-deleted duplicates

### 2. `prepareMultiColumnInsert()` — shared SQL builder

- Extracted common logic from `create()` into a reusable helper
- Handles: INSERT INTO + VALUES clause building, audit stamp extraction, aspect parameter binding, URN extraction
- Caller provides only the ON DUPLICATE KEY clause, eliminating ~80 lines of duplication
- **Note:** `create()` now calls this helper instead of inline SQL construction. Output SQL is identical — pure extract-method refactor.

### 3. `buildOnDuplicateKeyForCreate()` / `buildOnDuplicateKeyForUpsert()` — split clause generation

- `create`: sets aspects + conditionally throws via `CAST('DuplicateKeyException' AS UNSIGNED)`
- `upsert`: sets aspects + `lastmodifiedon` + `deleted_ts = NULL`
- `DELETED_TS_DUPLICATE_KEY_CHECK` renamed to `ON_DUPLICATE_KEY_UPDATE` for clarity

### 4. `IEbeanLocalAccess` + `InstrumentedEbeanLocalAccess` — interface and instrumentation

- Added `batchUpsert()` to the interface
- `InstrumentedEbeanLocalAccess` delegates with latency/error tracking

### 5. `EbeanLocalAccessTest` — integration tests against embedded MariaDB

- `testBatchUpsertMultipleAspects`: inserts two aspects in one call, verifies both persisted
- `testBatchUpsertOverwritesExistingValues`: upserts over existing data, verifies update
- `testBatchUpsertClearsDeletedTs`: soft-delete then upsert, verifies `deleted_ts` cleared
- `testBatchUpsertSingleAspect`: single-aspect degenerate case
- `testBatchUpsertPreservesIngestionTrackingContext`: verifies emitter/emitTime round-trip

## Testing

- All `EbeanLocalAccessTest` and `InstrumentedEbeanLocalAccessTest` tests pass
- Existing tests unaffected — `create()` produces identical SQL output via the shared helper
- *** The FULL E2E Testing suite from the original mega PR still applies here too

## Checklist

- [x] Compiles with `./gradlew :dao-api:compileJava :dao-impl:ebean-dao:compileJava`
- [x] Tests pass with `./gradlew :dao-impl:ebean-dao:test --tests EbeanLocalAccessTest --tests InstrumentedEbeanLocalAccessTest`
- [x] No breaking changes to public interfaces
- [x] Files are byte-identical to corresponding code in #598

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -32,6 +32,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -50,6 +51,7 @@
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 
+import static com.linkedin.metadata.dao.BaseReadDAO.LATEST_VERSION;
```
chmeng0: nit: it's not used in this pr, but i see it's used in the large pr containing all the changes (https://github.com/linkedin/datahub-gma/pull/598) so it should be good

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -784,4 +736,186 @@ static <ASPECT extends RecordTemplate, URN> EbeanMetadataAspect findLatestMetada
...
+        onDuplicateKey.append(", ");
+      }
+    }
+    onDuplicateKey.append(DELETED_TS_SET_VALUE_CONDITIONALLY);
+    return onDuplicateKey.toString();
+  }
+
+  /**
+   * Helper method to build the ON DUPLICATE KEY UPDATE clause for batchUpsert() method.
+   * This clause always updates the row and clears deleted_ts (UPSERT semantics).
```
chmeng0, line 894: @jphui could you help confirm the behavior of batchUpsert() on duplicate key (row exists)?
in that case, this buildOnDuplicateKeyForUpsert() appends `ON DUPLICATE KEY UPDATE <aspect columns>, lastmodifiedon = :lastmodifiedon, deleted_ts = NULL;` and lastmodified**by** is not updated? should we update `lastmodifiedby` as well?

jphui: >   ON DUPLICATE KEY UPDATE %s = :metadata, lastmodifiedon = :lastmodifiedon, deleted_ts = NULL;
> 
>   Same pattern — lastmodifiedby is not updated on duplicate key in the existing single-aspect path either. So the batch upsert is consistent with the existing behavior. It's not a
>    new bug we introduced — it's pre-existing.
> 
>   The lastmodifiedby is baked into each aspect's JSON blob (via AuditedAspect.setLastmodifiedby), so the actual actor info is preserved inside the aspect column. The top-level
>   lastmodifiedby column just doesn't get refreshed on update, which has been the case all along.

This is basically just keeping behavior consistent with existing behavior.

chmeng0 (approved):
LGTM

PR #611, opened 2026-04-10
split filter query to avoid long transactions

mridul111998:
### Problem

The `listUrns()` filter query in `EbeanLocalAccess` was embedding a `SELECT COUNT(urn)` subquery inside the main `SELECT urn` query, executing both within a **single implicit Ebean transaction**:

```sql
SELECT urn, (SELECT COUNT(urn) FROM metadata_entity_foo WHERE <filters>) as _total_count
FROM metadata_entity_foo
WHERE <filters>
LIMIT :limit OFFSET :offset
```

While MySQL optimizes the uncorrelated scalar subquery and evaluates the COUNT only once, the single SQL statement still performs two full operations within one transaction:
1. A full COUNT scan over all matching rows
2. A paginated SELECT scan with LIMIT/OFFSET

When combined, these operations push the total transaction duration past the **5-second kill threshold** enforced by LinkedIn's [MySQL Long Transaction Audit and Kill](https://linkedin.atlassian.net/wiki/spaces/SOP/pages/794689680/MySQL+Long+Transaction+Audit+and+Kill) policy. The kill is **per-transaction** — any transaction exceeding 5 seconds is terminated, regardless of whether it contains one statement or many.

**Impact**: ~40 transactions were killed in the past week due to this pattern ([dashboard](https://observe.prod.linkedin.com/g/d/cfadqnbp2e9z4c/wip-corp-long-txn-dashboard-for-customers?orgId=1&from=now-7d&to=now&var-threshold=5&var-db=metagalaxy)), making it the **highest single contributor** to long-running transaction kills across metadata services.

### Why Splitting Fixes This

The call chain `MGA filter() → filterInternal() → dao.listUrns() → _localAccess.listUrns()` does **not** wrap `listUrns()` in an explicit transaction. In Ebean, each `SqlQuery.findOne()` / `SqlQuery.findList()` without an explicit `beginTransaction()` runs in its own **implicit autocommit transaction**.

**Before (single transaction):**
- 1 implicit transaction containing 1 SQL that does COUNT scan + paginated SELECT
- If total execution (e.g. 3s COUNT + 3s SELECT = 6s) > 5s → **killed**

**After (two independent transactions):**
- Transaction 1: `countQuery.findOne()` — COUNT scan only (~3s) → **under 5s ✓**
- Transaction 2: `pageQuery.findList()` — paginated SELECT only (~3s) → **under 5s ✓**

Each transaction is independent and each individually stays under the kill threshold.

A `resolveTotalCount()` helper handles **race conditions** between the two queries (rows inserted or deleted between count and select), ensuring the `ListResult` metadata stays consistent.

### Limitations

If the COUNT query alone takes >5s on a very large table, this split won't help — index optimization would be needed instead. For the current workloads (e.g., metagalaxy dataset table with ~1822 openhouse-platform rows), each individual query is well under 5s.

### `SQLStatementUtils.java`
- Removed `SQL_FILTER_TEMPLATE` (the embedded subquery template)
- Removed `createFilterSql()` (which took a `hasTotalCount` boolean to toggle the subquery)
- Added `SQL_COUNT_TEMPLATE` and `SQL_SELECT_URN_WHERE_TEMPLATE`
- Added `createSelectFilterSql()` — generates `SELECT urn FROM table WHERE ...`
- Added `createCountFilterSql()` — generates `SELECT COUNT(urn) AS total_count FROM table WHERE ...`

### `EbeanLocalAccess.java`
- `listUrns(filter, sort, start, pageSize)` — split into count query + paginated select query
- `listUrns(filter, sort, lastUrn, pageSize)` — updated to use `createSelectFilterSqlQuery()`
- Added `createSelectFilterSqlQuery()` and `createCountSqlQuery()` private helpers
- Added `resolveTotalCount()` to handle count/data race conditions
- Added `toListResult(List<T>, int totalCount, int start, int pageSize)` overload

### Test Changes
- `SQLStatementUtilsTest`: Updated `testCreateFilterSql()` and `testCreateFilterSqlWithArrayContainsCondition()` to validate both `createCountFilterSql()` and `createSelectFilterSql()` with dollar and non-dollar virtual column variants
- `EbeanLocalAccessTest`: Added `testResolveTotalCount()` covering 7 scenarios (deletion race, insertion race, normal pagination, empty results, boundary conditions) and `testListUrnsReturnsEmptyWhenNoResults()`

### Unit Tests
All existing + new tests pass:
- `SQLStatementUtilsTest` — validates generated SQL for count and select queries
- `EbeanLocalAccessTest` — validates `resolveTotalCount()` logic and empty-result handling

### Integration Testing (local MGA deployment)

**Baseline (current master via EI D2)**:
```
grpcurli --dv-auth SELF -f ei-ltx1 d2://datasetAssetService \
  proto.com.linkedin.mg.assets.service.DatasetAssetService.filter -d '{
    "indexFilter": {
        "criteria": {
            "items": [{
                "aspect": "com.linkedin.common.urn.DatasetUrn",
                "pathParams": {
                    "condition": "Condition_EQUAL",
                    "path": "/platform/platformName",
                    "value": { "stringValue": "openhouse" }
                }
            }]
        }
    },
    "paging": { "start": 3, "count": 2 }
}'
```
Response:
```json
{
  "paging": { "start": 3, "count": 2, "total": 1822 },
  "values": [
    { "urn": { "platform": { "platformName": "openhouse" }, "datasetName": "actual_dataset", "origin": "FabricType_EI" } },
    { "urn": { "platform": { "platformName": "openhouse" }, "datasetName": "c1.d1.t111", "origin": "FabricType_PROD" } }
  ]
}
```

**Split-query branch (local MGA with datahub-gma SNAPSHOT)**:
```
grpcurli --dv-auth SELF -f ei-ltx1 localhost:25403 \
  proto.com.linkedin.mg.assets.service.DatasetAssetService.filter -d '{ <same payload> }'
```
Response: **Identical** — same `total: 1822`, same URN values, same ordering.

Pipeline: datahub-gma `publishToMavenLocal` → metadata-models `mint build && mint snapshot && mint release` (277.0.2-SNAPSHOT) → MGA `mint debug` (local deploy on localhost:25403)

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -360,15 +360,24 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   @Override
   public ListResult<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion,
       int start, int pageSize) {
-    final SqlQuery sqlQuery = createFilterSqlQuery(indexFilter, indexSortCriterion, start, pageSize);
-    final List<SqlRow> sqlRows = sqlQuery.findList();
-    if (sqlRows.isEmpty()) {
-      final List<SqlRow> totalCountResults = createFilterSqlQuery(indexFilter, indexSortCriterion, 0, DEFAULT_PAGE_SIZE).findList();
-      final int actualTotalCount = totalCountResults.isEmpty() ? 0 : totalCountResults.get(0).getInteger("_total_count");
-      return toListResult(actualTotalCount, start, pageSize);
+    // Run COUNT in a separate query/transaction so neither query exceeds the 5s kill threshold.
+    final String baseSql = SQLStatementUtils.createFilterSql(_entityType, indexFilter, _nonDollarVirtualColumnsEnabled, validator);
+    final String countSql = baseSql.replaceFirst("SELECT urn", "SELECT COUNT(urn) AS _total_count");
+    final SqlRow countRow = _server.createSqlQuery(countSql).findOne();
+    final int totalCount = (countRow == null) ? 0 : countRow.getInteger("_total_count");
+    if (totalCount == 0) {
+      return toListResult(0, start, pageSize);
     }
+
+    // Paginated URN query without embedded COUNT subquery.
+    StringBuilder selectSql = new StringBuilder(baseSql);
+    selectSql.append("\n");
+    selectSql.append(parseSortCriteria(_entityType, indexSortCriterion, _nonDollarVirtualColumnsEnabled, validator));
+    selectSql.append(String.format(" LIMIT %d", Math.max(pageSize, 0)));
+    selectSql.append(String.format(" OFFSET %d", Math.max(start, 0)));
+    final List<SqlRow> sqlRows = _server.createSqlQuery(selectSql.toString()).findList();
     final List<URN> values = sqlRows.stream().map(sqlRow -> getUrn(sqlRow.getString("urn"), _urnClass)).collect(Collectors.toList());
-    return toListResult(values, sqlRows, null, start, pageSize);
+    return toListResult(totalCount, start, pageSize, values);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


jphui (approved):
Nice thanks for the E2E testing as well done.

rakhiagr:
I agree with the idea to split the COUNT and SELECT queries to avoid long-running transactions. But I think this introduces a behavior change and potential risk (thinking out loud — please correct me if I'm misunderstanding anything):

Before:  COUNT and SELECT were part of the same SQL statement and  calculated from the same snapshot. So, pagination metadata (totalCount, havingMore) remained internally consistent.

After: We now run two independent queries: COUNT and SELECT , so inserts or updates between them can lead to inconsistent pagination metadata.

Example:

Assume:

start = 90
pageSize = 10
COUNT returns 95 rows


Before SELECT runs, 5 new rows are inserted.
Now:

COUNT = 95
SELECT OFFSET 90 LIMIT 10 returns 10 rows


So:
values.size() = 10
totalCount - start = 5

This hits exception path values.size() > totalCount - start . In this case, we would fail the request due to concurrent inserts/updates, whereas previously we would return results.

rakhiagr:
Also, since this change affects pagination behavior under concurrent writes, can we add a couple of unit tests around this?

For example:

Case where `values.size() < totalCount - start`
Case where `values.size() == totalCount - start`
Case where v`alues.size() > totalCount - start `(new exception path)

PR #613, opened 2026-04-22
feat(dao-api): add server-side URN validation on all DB write paths

jsdonn:
## Summary
- Adds `UrnValidator` utility and `InvalidUrnException` to reject URNs with blank fields, embedded whitespace, or control/format characters on all write paths
- Guards 7 chokepoint methods in `BaseLocalDAO` (`add`, `addMany`, `batchUpsert`, `create`, `delete`, `deleteMany`, `deleteAll`) and 2 methods in `EbeanGenericLocalDAO` (`save`, `delete`)
- `InvalidUrnException` extends `IllegalArgumentException` so existing gRPC/Rest.li error-mapping handlers automatically return `INVALID_ARGUMENT` / 400 with no new plumbing

## Motivation
Incident where `urn:li:dataset:(urn:li:dataPlatform:openhouse, ,PROD)` (single-space datasetName) was ingested — proto-level validation considered it valid, but it caused downstream failures. Original fix was in [metadata-graph-assets PR #883](https://github.com/linkedin-multiproduct/metadata-graph-assets/pull/883), but reviewer feedback directed moving validation into datahub-gma since it is the common layer for all ingestion services (MGA, TMS, etc.), avoiding duplication.

## Design decisions
- **Reads are NOT validated** — historical bad rows remain accessible; cleanup requires DB-direct admin action
- **Backfill methods are NOT validated** — they re-emit existing data, not create new data
- Validation fires **before** any DB interaction or transaction
- Checks: blank/whitespace-only (`BLANK_FIELD`), any whitespace including NBSP (`CONTAINS_WHITESPACE`), Unicode control/format chars like ZWSP and BOM (`CONTROL_CHARACTER`)
- `quote()` escapes all non-printable-ASCII as `\uXXXX` in log messages to prevent ANSI injection
- Nested URNs within tuple key parts are validated recursively

## Test plan
- [x] 20 unit tests in `UrnValidatorTest` (happy path, incident URN, blank fields, NBSP, nested URN recursion, leading/trailing/embedded whitespace, tab, ZWSP, BOM, quote hardening, codepoint rendering, IllegalArgumentException subtyping)
- [x] 9 integration tests in `BaseLocalDAOTest` (all 7 write paths reject invalid URNs, valid URN passes through, exception is catchable as IllegalArgumentException)
- [x] `./gradlew :dao-api:test` passes
- [x] `./gradlew :dao-impl:ebean-dao:compileJava` passes
- [x] `./gradlew checkstyleMain` passes


dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Validates a Pegasus URN for write operations. Throws {@link InvalidUrnException} on the first
+   * violation found.
+   *
+   * @param urn the URN to validate
+   * @throws InvalidUrnException if any key part contains invalid content
+   */
+  public static void validateUrn(@Nonnull Urn urn) {
```
mridul111998:
```suggestion
  private static void validateUrn(@Nonnull Urn urn) {
```
Why do we need this method to be public?

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
```diff
@@ -0,0 +1,198 @@
...
+          String.format("URN %s contains control/format character at %s: %s codepoints=%s",
+              entityType, path, quote(part), codepointList(part)));
+    }
+
+    if (part.startsWith(URN_PREFIX)) {
+      validateNestedUrn(part, entityType, path);
+    }
+  }
+
+  private static void validateNestedUrn(@Nonnull String urnString, @Nonnull String parentEntityType,
```
mridul111998:
```suggestion
  private static void validateNestedUrn(@Nonnull String urnString,
```
seems unused variable

dao-api/src/main/java/com/linkedin/metadata/dao/exception/InvalidUrnException.java
```diff
@@ -0,0 +1,36 @@
+package com.linkedin.metadata.dao.exception;
```
mridul111998, line 1: pulling [this](https://github.com/linkedin-multiproduct/metadata-graph-assets/pull/883#discussion_r3126021357) comment from previous PR, we can cleanup those 31 rows directly by running MGA curli right ? We don't need mysql team's help on it

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
```diff
@@ -0,0 +1,198 @@
...
+      @Nonnull String parentPath) {
+    Urn nested;
+    try {
+      nested = Urn.createFromString(urnString);
+    } catch (Exception e) {
+      // If it can't be parsed as a URN, the string-level checks above already cover it.
+      return;
```
mridul111998: I don't think it is correct to silently consume exception in this case. Also it is not always the case that above checks would have covered it. Take this case for example `urn:li:dataset:(urn:li:DataPlatform:hdfs,/data/tracking/PageViewEvent,PROD)` . This is a wrong urn and this exception will silently return and this invalid urn will get ingested in our DB. 
I would recommend to log error here and throw the exception

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
```diff
@@ -0,0 +1,198 @@
...
+      throw new InvalidUrnException(entityType, Reason.BLANK_FIELD, path, quote(part),
+          String.format("URN %s is blank or whitespace-only at %s: %s codepoints=%s",
+              entityType, path, quote(part), codepointList(part)));
+    }
+
+    if (containsAnyWhitespace(part)) {
+      throw new InvalidUrnException(entityType, Reason.CONTAINS_WHITESPACE, path, quote(part),
+          String.format("URN %s contains whitespace at %s: %s codepoints=%s",
+              entityType, path, quote(part), codepointList(part)));
+    }
```
mridul111998: there is overlap in these 2 cases, `containsAnyWhitespace` contains everything inside `isBlank` except `part.isEmpty` condition. I would recommend that you only do `part.isEmpty` check in `isBlank` function and let everything get captured by `containsAnyWhitespace`. Or another option is to merge these 2 cases as `isEmptyOrContainsWhitespaces`

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -81,6 +81,7 @@
 
 import static com.linkedin.metadata.dao.utils.IngestionUtils.*;
 import static com.linkedin.metadata.dao.utils.ModelUtils.*;
+import static com.linkedin.metadata.dao.utils.UrnValidator.validateUrnForWrite;
```
mridul111998: I can see multiple methods in this file through which invalid urn's can get into our db. Most of the backfill methods are currently not guarded. Example 
```
public <ASPECT extends RecordTemplate> Optional<ASPECT> backfill(@Nonnull Class<ASPECT> aspectClass,
      @Nonnull URN urn) {
    return backfill(BackfillMode.BACKFILL_ALL, aspectClass, urn);
  }
```

Also abstract methods like `updateEntityTables` have unchecked urn's passing. 
Please check all methods and ensure that there are no loop holes through which invalid urn's can pass through

jsdonn: there's no real gap for backfill methods bc they can only ever backfill urns that are already in the db. if a user tries to backfill some new, malformed urn, it will just be a no-op

jsdonn: same for updateEntityTables - it tries to read the urn first and if the urn doesn't exist, it doesn't do any writes.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
jsdonn: it doesn't thanks for pointing out.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
jsdonn: thanks for catching this, fixed.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
jsdonn: thanks for catching this, fixed.

dao-api/src/main/java/com/linkedin/metadata/dao/utils/UrnValidator.java
jsdonn: right, I was thinking separating them out could make the error message more clear but that's fine, we can just treat blank / whitespace only the same as the "contains whitespace" case

dao-api/src/main/java/com/linkedin/metadata/dao/exception/InvalidUrnException.java
jsdonn: right, we don't need MySQL team's help. I will clean these up.

mridul111998 (commented):
Can you add end to end testing results as well. something where we are passing invalid urn's from MGA or TMS and that is causing failures rather than ingesting in DB. and the same curli's which run perfectly fine on master and ingests data in DB

mridul111998 (approved):

PR #614, opened 2026-04-24
fix(ebean-dao): skip redundant aspect null/deleted checks in filter when path filter present

rakhiagr:
## Summary
- Skip `<aspect_column> IS NOT NULL` and `JSON_EXTRACT(<aspect_column>, '$.gma_deleted') IS NULL` checks in `parseIndexFilter()` when the criterion has an effective path filter (generated column exists)
-  These checks are added for every non-URN aspect criterion (e.g. `a_status`, `a_ownership`, `a_accountableownership`, etc.) and force MySQL to do row lookups on the JSON column, preventing index-only scans 
- When a path filter like `i_status$removed = 'false'` is present, the generated column already excludes null/deleted aspects — NULL fails all SQL comparisons
- Checks are preserved for bare criteria (no pathParams) and when the generated column doesn't exist (schema mismatch fallback)

## Why
These conditions force MySQL to perform row lookups on the JSON column for every matching row, preventing index-only scans. Combined with a covering index (`platformName, removed, deleted_ts`), removing them reduces filter COUNT queries from ~1.2s to ~0.07s on prod.

### How aspect soft-deletion works
When an aspect is deleted, `EbeanLocalAccess.add()` overwrites the entire aspect JSON with `{"gma_deleted":{...}}`. The generated column (e.g. `i_status$removed`) is derived via `JSON_EXTRACT(<aspect_column>, '$.aspect.<field>')` — after deletion this returns NULL. Since NULL fails all SQL comparisons (`=`, `<`, `>`, `IN`, `LIKE`, etc.), the path filter itself excludes deleted aspects.

 | Row state | Aspect column | Generated column | Matches any comparison? |
  |---|---|---|---|
  | Aspect exists | `{"aspect":{"removed":false},...}` | `'false'` | YES |
  | Aspect soft-deleted | `{"gma_deleted":{...}}` | NULL | NO |
  | Aspect never written | NULL | NULL | NO |

 ## Proof (metagalaxy_stg with covering index)

  | Query | Time | EXPLAIN |
  |---|---|---|
  | With aspect null + `JSON_EXTRACT` checks | 0.47s | 45% filtered, Using where |
  | Without checks (path filter only) | 0.07s | 100% filtered, Using index |
  | Row counts identical | 77,521 | Both queries return same count |

## Checklist
- [x] All `SQLIndexFilterUtilsTest` tests updated and pass
- [x] All `SQLStatementUtilsTest` filter/groupBy tests updated and pass
### E2E: datahub-gma → metadata-models → metadata-graph-assets
- [x] Published datahub-gma v0.6.183 locally
- [x] Built metadata-models v277.0.11-SNAPSHOT with new GMA
- [x] Built metadata-graph-assets with new metadata-models — all 236 MGA unit tests pass
### QEI deploy (metadata-graph-assets with new GMA)
- [x] Deployed MGA to QEI with the change
- [x] Verified via IntelliJ debugger — SQL confirmed no `a_status IS NOT NULL` and no `JSON_EXTRACT(a_status, '$.gma_deleted') IS NULL` in COUNT/DATA queries
- [x] DatasetAssetService/filter: 590–770ms (QEI, pageSize=1000, records=1000)
- [x] MlModelInstanceAssetService/filter: working with IN + EQUAL across two aspects
- [x] Pagination tested at OFFSET 0/2000/4000 — consistent latency
- [x] PR conforms to DataHub contributing guidelines (commit format, etc.)
- [x] Docs updated (if applicable)

#### DatasetAssetService/filter (kafkasyncer pattern — broad scan) on EI with index `idx_platform_status_deleted` ([PR](https://github.com/linkedin-multiproduct/metadata-graph-assets/pull/890/changes))

Old COUNT query:

```sql
  SELECT COUNT(urn) AS _total_count FROM metadata_entity_dataset
  WHERE a_status IS NOT NULL
  AND JSON_EXTRACT(a_status, '$.gma_deleted') IS NULL
  AND i_status$removed = 'false'
  AND i_urn$platform$platformName = 'kafka'
  AND deleted_ts IS NULL

+--------------+
| _total_count |
+--------------+
|        77589 |
+--------------+
1 row in set (1.09 sec)
```

New COUNT query:
```sql
  SELECT COUNT(urn) AS _total_count FROM metadata_entity_dataset
  WHERE i_status$removed = 'false'
  AND i_urn$platform$platformName = 'kafka'
  AND deleted_ts IS NULL

+--------------+
| _total_count |
+--------------+
|        77589 |
+--------------+
1 row in set (0.10 sec)
```

Old DATA query:
```sql
  SELECT urn FROM metadata_entity_dataset
  WHERE a_status IS NOT NULL
  AND JSON_EXTRACT(a_status, '$.gma_deleted') IS NULL
  AND i_status$removed = 'false'
  AND i_urn$platform$platformName = 'kafka'
  AND deleted_ts IS NULL
  ORDER BY URN LIMIT 1000 OFFSET 0

+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1000 rows in set (0.06 sec)
```

New DATA query:
```sql
  SELECT urn FROM metadata_entity_dataset
  WHERE i_status$removed = 'false'
  AND i_urn$platform$platformName = 'kafka'
  AND deleted_ts IS NULL
  ORDER BY URN LIMIT 1000 OFFSET 0

+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1000 rows in set (0.06 sec)
```

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
@@ -185,27 +185,46 @@ public class SQLIndexFilterUtils {
     if (indexFilter != null && indexFilter.hasCriteria()) {
       for (IndexCriterion indexCriterion : indexFilter.getCriteria()) {
         final String aspect = indexCriterion.getAspect();
-        if (!isUrn(aspect)) {
-          // if aspect is not urn, then check aspect is not soft deleted and is not null
-          final String aspectColumn = getAspectColumnName(entityType, indexCriterion.getAspect());
-          sqlFilters.add(aspectColumn + " IS NOT NULL");
-          sqlFilters.add(String.format(SOFT_DELETED_CHECK, aspectColumn));
-        }
-
         final IndexPathParams pathParams = indexCriterion.getPathParams(GetMode.NULL);
-        if (pathParams != null) {
-          validateConditionAndValue(indexCriterion);
-          final Condition condition = pathParams.getCondition();
 
-          final String indexedExpressionOrColumn =
-              getIndexedExpressionOrColumn(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled, schemaValidator);
-          if (indexedExpressionOrColumn == null) {
-            log.warn("Skipping filter: Neither expression index nor virtual column found for Aspect '{}' and Path '{}' for Asset '{}'",
-                aspect, pathParams.getPath(), entityType);
-            continue;
+        if (!isUrn(aspect)) {
+          boolean hasEffectivePathFilter = false;
+
+          if (pathParams != null) {
+            validateConditionAndValue(indexCriterion);
+            final String indexedExpressionOrColumn =
+                getIndexedExpressionOrColumn(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled, schemaValidator);
+            if (indexedExpressionOrColumn != null) {
+              // The generated column comparison already excludes rows where the aspect is NULL or
+              // soft-deleted (gma_deleted), because the generated column evaluates to NULL in both
+              // cases and NULL fails all SQL comparisons. No need for explicit aspect null/deleted checks.
+              hasEffectivePathFilter = true;
+              sqlFilters.add(parseSqlFilter(indexedExpressionOrColumn, pathParams.getCondition(), pathParams.getValue()));
+            } else {
+              log.warn("Skipping filter: Neither expression index nor virtual column found for Aspect '{}' and Path '{}' for Asset '{}'",
+                  aspect, pathParams.getPath(), entityType);
+            }
           }
 
-          sqlFilters.add(parseSqlFilter(indexedExpressionOrColumn, condition, pathParams.getValue()));
+          if (!hasEffectivePathFilter) {
+            // No effective path filter — need explicit aspect null and soft-deletion checks.
+            final String aspectColumn = getAspectColumnName(entityType, indexCriterion.getAspect());
+            sqlFilters.add(aspectColumn + " IS NOT NULL");
+            sqlFilters.add(String.format(SOFT_DELETED_CHECK, aspectColumn));
+          }
+        } else {
+          // URN criterion — only process path filter, no aspect null/deleted checks needed
+          if (pathParams != null) {
+            validateConditionAndValue(indexCriterion);
+            final String indexedExpressionOrColumn =
+                getIndexedExpressionOrColumn(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled, schemaValidator);
+            if (indexedExpressionOrColumn == null) {
+              log.warn("Skipping filter: Neither expression index nor virtual column found for Aspect '{}' and Path '{}' for Asset '{}'",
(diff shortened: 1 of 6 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
```diff
@@ -185,27 +185,46 @@ public static String parseIndexFilter(@Nonnull String entityType, @Nullable Inde
...
-                aspect, pathParams.getPath(), entityType);
-            continue;
+
+        if (!isUrn(aspect)) {
+          boolean hasEffectivePathFilter = false;
+
+          if (pathParams != null) {
+            validateConditionAndValue(indexCriterion);
+            final String indexedExpressionOrColumn =
+                getIndexedExpressionOrColumn(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled, schemaValidator);
```
mridul111998: This is duplicated in both the URN and non-URN branches, and only ever runs when pathParams != null. Could it be hoisted to a single call right after fetching pathParams?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
```diff
@@ -185,27 +185,46 @@ public static String parseIndexFilter(@Nonnull String entityType, @Nullable Inde
...
+        } else {
+          // URN criterion — only process path filter, no aspect null/deleted checks needed
+          if (pathParams != null) {
+            validateConditionAndValue(indexCriterion);
+            final String indexedExpressionOrColumn =
+                getIndexedExpressionOrColumn(entityType, aspect, pathParams.getPath(), nonDollarVirtualColumnsEnabled, schemaValidator);
+            if (indexedExpressionOrColumn == null) {
+              log.warn("Skipping filter: Neither expression index nor virtual column found for Aspect '{}' and Path '{}' for Asset '{}'",
+                  aspect, pathParams.getPath(), entityType);
+              continue;
```
mridul111998: this seems like an issue, this branch drops the entire criterion via continue when no virtual column exists. I feel we should explicitly throw exception rather than silently failing incase of misconfigured input. In this case a misconfigured input like "urn matches X" is sent by client instead of throwing exception we will end up resulting all the urn's.

rakhiagr: This is pre-existing behavior — the original code had the same `continue` when the generated column was missing for URN criteria.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
rakhiagr: Refactored.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLIndexFilterUtils.java
rakhiagr: I can fix it in follow up PR after CY.

mridul111998 (approved):

PR #615, opened 2026-04-27
perf(ebean-dao): share schema metadata cache per database to fix cold-start P99 spikes

ybz1013:
## Summary

- **Problem**: Each entity type on a shared database maintained its own `SchemaValidatorUtil` Caffeine cache for `information_schema` column/index metadata (10-min TTL). On cache expiry, the next request queries `information_schema` inline. For low-QPS entities (MlModel, DataPolicy, etc.) this single slow query dominates P95 and causes SLO breaches. Compounding factors: ~20 entity types on `metagalaxy` each had independent caches (20 redundant queries for the same data), all deployed together so all caches expired simultaneously (confirmed: 7+ hosts refreshing in the same second), and no pre-warming after deploy/restart.

- **Fix**: Introduce `SharedSchemaCache` — one cache instance per database URL (via static registry) instead of one per entity type. ~50 independent caches collapse to 4 (one per database). Each entity table is pre-warmed during `ensureSchemaUpToDate()` so the first post-deploy request is never slow. A background daemon thread refreshes all registered tables every 9 minutes with per-host random jitter (0–60s) to prevent thundering-herd expiry across the fleet.

- **Impact**: Reduces `information_schema` queries from ~150 to ~2 per database per refresh cycle per host. Zero API changes — `metadata-graph-assets`, `metadata-store`, and `metadata-graph-query` all benefit automatically.

### Files changed
- **NEW** `SharedSchemaCache.java` — per-database shared cache with background refresh + jitter
- **MODIFIED** `SchemaValidatorUtil.java` — new constructor accepting `SharedSchemaCache`; all public methods delegate to it when present; backward-compatible `EbeanServer` constructor unchanged for tests
- **MODIFIED** `EbeanLocalAccess.java` — resolves `SharedSchemaCache` by JDBC URL at construction; calls `registerAndPreWarm` in `ensureSchemaUpToDate()`

## Testing Done

- [x] Compiles cleanly with JDK 11 (`./gradlew :dao-impl:ebean-dao:compileJava`)
- [x] Existing `SchemaValidatorUtil(EbeanServer)` constructor unchanged — all existing tests remain compatible
- [x] `EbeanLocalAccessTest.resetValidatorInstance()` still works (injects old-style validator via reflection, bypassing shared cache — intentional for test isolation)
- [ ] Integration tests blocked by Apple Silicon MariaDB setup (`brew install mariadb` required per CLAUDE.md) — pre-existing environment issue unrelated to this change

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -111,8 +111,9 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
 
   public void ensureSchemaUpToDate() {
     _schemaEvolutionManager.ensureSchemaUpToDate();
-    // Pre-warm the shared cache for this entity's table so the first request is never slow.
+    // Pre-warm the shared cache for both the prod and test tables so the first request is never slow.
     validator.registerAndPreWarm(getTableName(_entityType));
+    validator.registerAndPreWarm(SQLSchemaUtils.getTestTableName(_entityType));
   }
 
   @Override
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -113,7 +113,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
     _schemaEvolutionManager.ensureSchemaUpToDate();
     // Pre-warm the shared cache for both the prod and test tables so the first request is never slow.
     validator.registerAndPreWarm(getTableName(_entityType));
-    validator.registerAndPreWarm(SQLSchemaUtils.getTestTableName(_entityType));
+    validator.registerAndPreWarm(getTestTableName(_entityType));
   }
 
   @Override
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -13,6 +13,7 @@ import com.linkedin.metadata.dao.utils.RecordUtils;
 import com.linkedin.metadata.dao.utils.SQLIndexFilterUtils;
 import com.linkedin.metadata.dao.utils.SQLStatementUtils;
 import com.linkedin.metadata.dao.utils.SchemaValidatorUtil;
+import com.linkedin.metadata.dao.utils.SharedSchemaCache;
 import com.linkedin.metadata.events.IngestionTrackingContext;
 import com.linkedin.metadata.query.ExtraInfo;
 import com.linkedin.metadata.query.ExtraInfoArray;
@@ -92,7 +93,16 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
     _entityType = ModelUtils.getEntityTypeFromUrnClass(_urnClass);
     _schemaEvolutionManager = createSchemaEvolutionManager(serverConfig);
     _nonDollarVirtualColumnsEnabled = nonDollarVirtualColumnsEnabled;
-    validator = new SchemaValidatorUtil(server);
+    validator = buildValidator(server, serverConfig);
+  }
+
+  private static SchemaValidatorUtil buildValidator(EbeanServer server, ServerConfig serverConfig) {
+    String dbUrl = serverConfig.getDataSourceConfig() != null
+        ? serverConfig.getDataSourceConfig().getUrl() : null;
+    if (dbUrl != null && !dbUrl.isEmpty()) {
+      return new SchemaValidatorUtil(SharedSchemaCache.getInstance(server, dbUrl));
+    }
+    return new SchemaValidatorUtil(server);
   }
 
   public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor) {
@@ -101,6 +111,8 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
 
   public void ensureSchemaUpToDate() {
     _schemaEvolutionManager.ensureSchemaUpToDate();
+    // Pre-warm the shared cache for this entity's table so the first request is never slow.
+    validator.registerAndPreWarm(getTableName(_entityType));
   }
 
   @Override
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -99,10 +99,11 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   private static SchemaValidatorUtil buildValidator(EbeanServer server, ServerConfig serverConfig) {
     String dbUrl = serverConfig.getDataSourceConfig() != null
         ? serverConfig.getDataSourceConfig().getUrl() : null;
-    if (dbUrl != null && !dbUrl.isEmpty()) {
-      return new SchemaValidatorUtil(SharedSchemaCache.getInstance(server, dbUrl));
+    if (dbUrl == null || dbUrl.isEmpty()) {
+      throw new IllegalStateException(
+          "ServerConfig must have a DataSourceConfig with a non-empty URL to use SharedSchemaCache");
     }
-    return new SchemaValidatorUtil(server);
+    return new SchemaValidatorUtil(SharedSchemaCache.getInstance(server, dbUrl));
   }
 
   public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor) {
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
@@ -102,10 +102,6 @@ public class SchemaValidatorUtil {
    */
   @VisibleForTesting
   void clearCaches() {
-    if (sharedCache != null) {
-      sharedCache.clearCaches();
-      return;
-    }
     indexCache.invalidateAll();
     columnCache.invalidateAll();
     indexExpressionCache.invalidateAll();
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
@@ -20,10 +20,18 @@ import lombok.extern.slf4j.Slf4j;
  * Utility class for validating presence of columns and indexes in MySQL tables
  * by querying information_schema. Uses Caffeine caches to reduce DB load and
  * support eventual consistency of schema evolution.
+ *
+ * <p>When constructed with a {@link SharedSchemaCache}, all cache operations delegate to that
+ * shared instance so that entity types on the same database share one set of caches instead of
+ * maintaining redundant per-entity copies.
  */
 @Slf4j
 public class SchemaValidatorUtil {
+  @Nullable
   private final EbeanServer server;
+
+  @Nullable
+  private final SharedSchemaCache sharedCache;
 
   // Cache: tableName → Set of index names
   // Configuration:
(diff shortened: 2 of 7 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,201 @@
...
+    for (String table : registeredTables) {
+      refreshTable(table);
+    }
+  }
+
+  private void refreshTable(String tableName) {
+    try {
+      columnCache.put(tableName, loadColumns(tableName));
+      indexCache.put(tableName, loadIndexes(tableName));
+      log.debug("Schema cache refreshed for table '{}'", tableName);
```
rakhiagr: We should refresh `indexExpressionCache` as well?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,202 @@
...
+  public String getIndexExpression(@Nonnull String tableName, @Nonnull String indexName) {
+    String lowerTable = tableName.toLowerCase();
+    String lowerIndex = indexName.toLowerCase();
+    try {
+      Map<String, String> indexes = indexExpressionCache.get(lowerTable, this::loadIndexesAndExpressions);
+      return SchemaValidatorUtil.cleanIndexExpression(
+          indexes != null ? indexes.getOrDefault(lowerIndex, null) : null);
+    } catch (Exception e) {
+      // MariaDB for local testing doesn't support EXPRESSION column
+      log.debug("Unable to load index expressions for table '{}': {}", lowerTable, e.getMessage());
```
rakhiagr: We don't have DEBUG level logs enabled I think. We should make it INFO level.

rakhiagr: And at all other places too

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -101,6 +111,8 @@ public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor)
 
   public void ensureSchemaUpToDate() {
     _schemaEvolutionManager.ensureSchemaUpToDate();
+    // Pre-warm the shared cache for this entity's table so the first request is never slow.
+    validator.registerAndPreWarm(getTableName(_entityType));
```
rakhiagr, line 116: Nit: this only pre-warms the prod tables. `batchGetUnion` calls `columnExists` with `getTestTableName(entityUrn)` when `isTestMode=true`. Should we do it for test tables also?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * <p>Compared to a per-entity-type {@link SchemaValidatorUtil}, this eliminates redundant
+ * {@code information_schema} queries for entity types that share the same database, pre-warms
+ * each table at startup so the first request is never slow, and refreshes in the background with
+ * random per-host jitter to prevent thundering-herd cache expiry across a fleet.
+ */
+@Slf4j
+public class SharedSchemaCache {
+
+  private static final long CACHE_TTL_MINUTES = 10;
```
rakhiagr, line 35: Nit: Lets add 1 line comment to each of these properties?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,202 @@
...
+    String lower = tableName.toLowerCase();
+    registeredTables.add(lower);
+    refreshTable(lower);
+  }
+
+  // ── cache access ─────────────────────────────────────────────────────────────
+
+  public boolean columnExists(@Nonnull String tableName, @Nonnull String columnName) {
+    Set<String> columns = columnCache.get(tableName.toLowerCase(), this::loadColumns);
+    return columns != null && columns.contains(columnName.toLowerCase());
```
rakhiagr: Nit: columns != null is dead code? Same for following methods?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,202 @@
...
+      log.info("Unable to load index expressions for table '{}': {}", lowerTable, e.getMessage());
+      return null;
+    }
+  }
+
+  // ── background refresh ────────────────────────────────────────────────────────
+
+  private void scheduleBackgroundRefresh() {
+    long jitterSeconds = ThreadLocalRandom.current().nextLong(0, JITTER_MAX_SECONDS + 1);
+    long initialDelay = REFRESH_INTERVAL_SECONDS + jitterSeconds;
```
rakhiagr: Can you explain why we did we add jitterSeconds here?

rakhiagr:
Few comments:
1. How are you planning to test whether these changes are helping with P99 spikes?
2. `EbeanLocalRelationshipQueryDAO` still uses new SchemaValidatorUtil(server). Should we migrate them too?
3. Lets add some unit tests for `SharedSchemaCache`?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
ybz1013: Good question! Without jitter, all ~7 metagalaxy hosts would start their background refresh at the same moment (since they all deploy together and `REFRESH_INTERVAL_SECONDS` is fixed). That would cause a thundering herd where all hosts simultaneously query `information_schema` every 9 minutes, amplifying DB load. Adding random jitter of 0–60s to the *initial* delay staggers the first refresh across hosts so subsequent cycles are naturally spread out.

ybz1013:
Thanks for the review @rakhiagr! Addressing all three:

1. **P99 testing plan**: The P99 improvement isn't easily unit-testable, but the mechanism is observable in two ways: (a) the `Schema cache refreshed for table '...'` INFO log emitted by `registerAndPreWarm` at startup confirms pre-warming happened before the first request; (b) in production, we can compare `information_schema` query counts before/after via MySQL slow-query logs or metrics — the cold-start queries that previously fired on the first entity request per pod restart should no longer appear. The background refresh log (`Background schema cache refresh: N tables`) also confirms the thundering-herd jitter is working across the fleet.

2. **`EbeanLocalRelationshipQueryDAO` migration**: Done — added a new `(EbeanServer, ServerConfig, EBeanDAOConfig)` constructor that delegates to `SharedSchemaCache` via the same `buildValidator()` pattern as `EbeanLocalAccess`. The existing two-arg constructors are kept for backward compatibility.

3. **Unit tests for `SharedSchemaCache`**: Added `SharedSchemaCacheTest` covering `columnExists`, `getColumns`, `indexExists`, `registerAndPreWarm` (verifies no DB call after warm via Mockito), `clearCaches`, singleton identity per URL, and `getIndexExpression` (mocked since MariaDB doesn't support the EXPRESSION column).

ybz1013:
[test-plan-mgcy26-shared-schema-cache.md](https://github.com/user-attachments/files/27251419/test-plan-mgcy26-shared-schema-cache.md)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
```diff
@@ -66,16 +74,41 @@ public class SchemaValidatorUtil {
...
   @VisibleForTesting
   void clearCaches() {
+    if (sharedCache != null) {
+      sharedCache.clearCaches();
+      return;
+    }
     indexCache.invalidateAll();
     columnCache.invalidateAll();
+    indexExpressionCache.invalidateAll();
   }
```
mridul111998, line 108: why is this change required, this function already seems redundant to me, Can we clean it up if not required

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -0,0 +1,202 @@
...
+  // Random per-host jitter (0-60s) added to the initial refresh delay so that a fleet of hosts
+  // (e.g. 7 metagalaxy pods all starting at deploy time) do not hammer information_schema simultaneously.
+  private static final long JITTER_MAX_SECONDS = 60;
+  // Upper bound on the number of tables tracked across all databases; sized generously for safety.
+  private static final int CACHE_MAX_SIZE = 1000;
+
+  private static final String SQL_GET_ALL_COLUMNS =
+      "SELECT COLUMN_NAME FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = database() AND TABLE_NAME = '%s'";
+  private static final String SQL_GET_ALL_INDEXES =
+      "SELECT DISTINCT INDEX_NAME FROM information_schema.STATISTICS WHERE TABLE_SCHEMA = database() AND TABLE_NAME = '%s'";
```
mridul111998, line 48: why do we need this query? we are getting index name already in `SQL_GET_ALL_INDEXES_WITH_EXPRESSIONS` query,

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SchemaValidatorUtil.java
ybz1013: Good catch — you're right that this change isn't strictly needed. `SchemaValidatorUtil.clearCaches()` is `@VisibleForTesting` and our tests actually call `SharedSchemaCache.clearRegistry()` / `SharedSchemaCache.clearCaches()` directly, so the delegation I added here is dead code. I'll remove it and leave the method body clearing only the per-instance Caffeine caches (which is the pre-existing behavior). If you'd prefer we remove the whole method since it's also not called anywhere, happy to do that too — let me know.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
ybz1013: The two queries need to stay separate for MariaDB compatibility. MariaDB (our embedded test DB) does not have the `EXPRESSION` column in `information_schema.STATISTICS`, so any query that selects it throws an exception. `SQL_GET_ALL_INDEXES` (no EXPRESSION column) is used by `indexExists()`, which must work reliably in all environments including local dev/test. `SQL_GET_ALL_INDEXES_WITH_EXPRESSIONS` is only used by `getIndexExpression()`, which wraps the call in a try/catch and gracefully returns `null` on failure (the MariaDB case). If we merged them into one query, `indexExists()` would throw in MariaDB and break all local tests.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -92,7 +93,16 @@ public EbeanLocalAccess(EbeanServer server, ServerConfig serverConfig, @Nonnull
...
+    validator = buildValidator(server, serverConfig);
+  }
+
+  private static SchemaValidatorUtil buildValidator(EbeanServer server, ServerConfig serverConfig) {
+    String dbUrl = serverConfig.getDataSourceConfig() != null
+        ? serverConfig.getDataSourceConfig().getUrl() : null;
+    if (dbUrl != null && !dbUrl.isEmpty()) {
+      return new SchemaValidatorUtil(SharedSchemaCache.getInstance(server, dbUrl));
+    }
+    return new SchemaValidatorUtil(server);
```
mridul111998: when can dburl be null or empty? I would recommend throwing an exception here and cleanup this path completely, and

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
rakhiagr: nit: flip `+ jitterSeconds` to `- jitterSeconds` — keeps the same fleet spread but always leaves a 60s buffer before TTL?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
ybz1013: Good point — in practice `dbUrl` should never be null since all production metagalaxy configs use URL-based datasource setup. The fallback was overly defensive. Done — `buildValidator()` now throws `IllegalStateException` if `dbUrl` is null or empty, and the silent fallback path is removed.

Also incorporated @rakagrawal's suggestion from Slack: added `log.info("Inline schema cache miss: loading ... for table '{}'", tbl)` inside all three `cache.get()` loaders in `SharedSchemaCache`, so operators can see unexpected cold loads in production (after this PR, these should only appear at startup before pre-warming completes, never mid-traffic).

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
rakhiagr: @ybz1013 Can you address this?

ybz1013: Good catch — done. With `+` the worst-case initial delay was 540+60 = 600s, which equals the TTL exactly and could let entries expire before the refresh fires. With `-` the initial delay is 480–540s, so the refresh always fires at least 2 minutes before the 10-minute TTL. Updated the constant comment to explain the new semantics.

rakhiagr (approved):
Thank you for addressing all the comments!

PR #616, opened 2026-05-04
feat(ebean-dao): add optional FORCE INDEX hint for filter query

NatalliaUlashchick:
## Summary

- Adds per-DAO `configureOptionalForceIndex(String indexName, Class<? extends RecordTemplate> requiredAspect)` that conditionally emits a MySQL `FORCE INDEX` hint on the offset-pagination `listUrns` filter query
- The hint only activates when the `IndexFilter` contains a criterion matching the required aspect class — other queries against the same entity table are unaffected
- At startup, validates the configured index exists via `SchemaValidatorUtil.indexExists()`; if missing, auto-disables the hint and logs an ERROR instead of letting queries fail

## Motivation

`MlModelInstanceAssetService.filter` intermittently times out because MySQL's `prefer_ordering_index=on` heuristic picks a PRIMARY key scan instead of the composite index `idx_urn$model_urn$status`. FORCE INDEX is the only structural fix that doesn't require DDL on the 117GB prod table or server-side optimizer changes. See `spec/force-index-mlmodelinstance-spec.md` for full investigation.

## Changes

| File | Change |
|---|---|
| `SQLStatementUtils.java` | New 5-arg `createFilterSql` overload with `forceIndexName`; existing 4-arg delegates with `null` |
| `IEbeanLocalAccess.java` | Added `configureOptionalForceIndex(String, Class<? extends RecordTemplate>)` |
| `EbeanLocalAccess.java` | Stores index name + required aspect FQCN; `resolveForceIndex()` gates on `IndexFilter` criteria; `validateForceIndex()` checks index existence at boot |
| `InstrumentedEbeanLocalAccess.java` | Delegates `configureOptionalForceIndex` to wrapped `_delegate` |
| `EbeanLocalDAO.java` | Public `configureOptionalForceIndex` that delegates to `_localAccess` |

## Test plan

- [x] Unit test: `createFilterSql` with null `forceIndexName` — identical to 4-arg (regression)
- [x] Unit test: `createFilterSql` with non-null `forceIndexName` — `FORCE INDEX` between table name and WHERE
- [x] Unit test: count-query `replaceFirst` rewrite preserves FORCE INDEX clause
- [x] Integration test: offset-pagination `listUrns` with FORCE INDEX executes against embedded MariaDB
- [x] Integration test: FORCE INDEX not applied when filter lacks the required aspect
- [x] Integration test: keyset-pagination path does NOT include FORCE INDEX
- [x] Integration test: null config produces identical results to default
- [x] Integration test: `validateForceIndex` auto-disables when index is missing
- [x] Integration test: `validateForceIndex` keeps hint when index exists (PRIMARY)

## Follow-up (metadata-graph-assets)

After this is released, MGA adds one line in `MlModelInstanceLocalDaoFactory.createInstance()`:
```java
dao.configureOptionalForceIndex("idx_urn$model_urn$status", MlModelInstanceStatus.class);
```

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -71,6 +71,7 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   private UrnPathExtractor<URN> _urnPathExtractor;
   private final SchemaEvolutionManager _schemaEvolutionManager;
   private final boolean _nonDollarVirtualColumnsEnabled;
+  private String _forceFilterIndexName;
 
   // TODO confirm if the default page size is 1000 in other code context.
   private static final int DEFAULT_PAGE_SIZE = 1000;
@@ -108,6 +109,11 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
 
   public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor) {
     _urnPathExtractor = urnPathExtractor;
+  }
+
+  @Override
+  public void setForceFilterIndexName(@Nullable String forceIndexName) {
+    _forceFilterIndexName = forceIndexName;
   }
 
   public void ensureSchemaUpToDate() {
@@ -376,7 +382,8 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   public ListResult<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion,
       int start, int pageSize) {
     // Run COUNT in a separate query/transaction so neither query exceeds the 5s kill threshold.
-    final String baseSql = SQLStatementUtils.createFilterSql(_entityType, indexFilter, _nonDollarVirtualColumnsEnabled, validator);
+    final String baseSql = SQLStatementUtils.createFilterSql(_entityType, indexFilter,
+        _nonDollarVirtualColumnsEnabled, validator, _forceFilterIndexName);
     final String countSql = baseSql.replaceFirst("SELECT urn", "SELECT COUNT(urn) AS _total_count");
     final SqlRow countRow = _server.createSqlQuery(countSql).findOne();
     final int totalCount = (countRow == null) ? 0 : countRow.getInteger("_total_count");
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -113,9 +113,10 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   }
 
   @Override
-  public void configureOptionalForceIndex(@Nullable String indexName, @Nullable String requiredAspectFqcn) {
+  public void configureOptionalForceIndex(@Nullable String indexName,
+      @Nullable Class<? extends RecordTemplate> requiredAspect) {
     _forceIndexName = indexName;
-    _forceIndexRequiredAspect = requiredAspectFqcn;
+    _forceIndexRequiredAspect = (requiredAspect != null) ? requiredAspect.getCanonicalName() : null;
   }
 
   public void ensureSchemaUpToDate() {
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -71,8 +71,8 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   private UrnPathExtractor<URN> _urnPathExtractor;
   private final SchemaEvolutionManager _schemaEvolutionManager;
   private final boolean _nonDollarVirtualColumnsEnabled;
-  private String _forceFilterIndexName;
-  private String _forceFilterIndexRequiredAspect;
+  private String _forceIndexName;
+  private String _forceIndexRequiredAspect;
 
   // TODO confirm if the default page size is 1000 in other code context.
   private static final int DEFAULT_PAGE_SIZE = 1000;
@@ -113,9 +113,9 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
   }
 
   @Override
-  public void setForceFilterIndex(@Nullable String indexName, @Nullable String requiredAspectFqcn) {
-    _forceFilterIndexName = indexName;
-    _forceFilterIndexRequiredAspect = requiredAspectFqcn;
+  public void configureOptionalForceIndex(@Nullable String indexName, @Nullable String requiredAspectFqcn) {
+    _forceIndexName = indexName;
+    _forceIndexRequiredAspect = requiredAspectFqcn;
   }
 
   public void ensureSchemaUpToDate() {
@@ -530,15 +530,15 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
    */
   @Nullable
   private String resolveForceIndex(@Nullable IndexFilter indexFilter) {
-    if (_forceFilterIndexName == null || _forceFilterIndexRequiredAspect == null) {
+    if (_forceIndexName == null || _forceIndexRequiredAspect == null) {
       return null;
     }
     if (indexFilter == null || !indexFilter.hasCriteria()) {
       return null;
     }
     boolean hasRequiredAspect = indexFilter.getCriteria().stream()
-        .anyMatch(c -> _forceFilterIndexRequiredAspect.equals(c.getAspect()));
-    return hasRequiredAspect ? _forceFilterIndexName : null;
+        .anyMatch(c -> _forceIndexRequiredAspect.equals(c.getAspect()));
+    return hasRequiredAspect ? _forceIndexName : null;
   }
 
   /**
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
@@ -113,6 +113,14 @@ public class EbeanLocalAccess<URN extends Urn> implements IEbeanLocalAccess<URN>
     _urnPathExtractor = urnPathExtractor;
   }
 
+  /**
+   * {@inheritDoc}
+   *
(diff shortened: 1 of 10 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -110,6 +112,12 @@ public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor)
     _urnPathExtractor = urnPathExtractor;
   }
 
+  @Override
+  public void setForceFilterIndex(@Nullable String indexName, @Nullable String requiredAspectFqcn) {
```
NatalliaUlashchick: the name doesn't reflect its signature

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java
```diff
@@ -389,9 +389,23 @@ public static <ASPECT extends RecordTemplate> String createAspectUpdateWithOptim
...
+   * @param forceIndexName MySQL index name for FORCE INDEX hint, or null to omit
+   * @return translated SQL where statement
+   */
+  public static String createFilterSql(String entityType, @Nullable IndexFilter indexFilter, boolean nonDollarVirtualColumnsEnabled,
+      @Nonnull SchemaValidatorUtil schemaValidator, @Nullable String forceIndexName) {
     final String tableName = getTableName(entityType);
-    String whereClause = parseIndexFilter(entityType, indexFilter, nonDollarVirtualColumnsEnabled, schemaValidator);
-    return "SELECT urn FROM " + tableName + "\n" + whereClause;
+    final String whereClause = parseIndexFilter(entityType, indexFilter, nonDollarVirtualColumnsEnabled, schemaValidator);
+    final String forceClause = (forceIndexName == null) ? "" : " FORCE INDEX (`" + forceIndexName + "`)";
```
NatalliaUlashchick: rename to forceIndex

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -376,7 +384,9 @@ public List<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSort
   public ListResult<URN> listUrns(@Nullable IndexFilter indexFilter, @Nullable IndexSortCriterion indexSortCriterion,
       int start, int pageSize) {
     // Run COUNT in a separate query/transaction so neither query exceeds the 5s kill threshold.
```
NatalliaUlashchick: move this comment to its actual line. Also add a comment explaining why we need force index here

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -130,7 +130,7 @@ public void testGetAspect() {
     EbeanMetadataAspect ebeanMetadataAspect = ebeanMetadataAspectList.get(0);
 
     // Expect: the content of aspect foo is returned
-    assertEquals(AspectFoo.class.getCanonicalName(), ebeanMetadataAspect.getKey().getAspect());
+    assertEquals(AspectFoo.class, ebeanMetadataAspect.getKey().getAspect());
```
NatalliaUlashchick: why did you change this?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/EbeanLocalAccessTest.java
```diff
@@ -614,7 +614,7 @@ public void testBatchUpsertSingleAspect() {
     assertEquals(1, results.size());
     assertEquals("{\"value\":\"single\"}", results.get(0).getMetadata());
     assertEquals(fooUrn.toString(), results.get(0).getKey().getUrn());
-    assertEquals(AspectFoo.class.getCanonicalName(), results.get(0).getKey().getAspect());
+    assertEquals(AspectFoo.class, results.get(0).getKey().getAspect());
```
NatalliaUlashchick: and this?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -110,11 +112,38 @@ public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor)
     _urnPathExtractor = urnPathExtractor;
   }
 
+  @Override
+  public void configureOptionalForceIndex(@Nullable String indexName,
+      @Nullable Class<? extends RecordTemplate> requiredAspect) {
+    _forceIndexName = indexName;
+    _forceIndexRequiredAspect = (requiredAspect != null) ? requiredAspect.getCanonicalName() : null;
```
ZihanLi58: should this be a list of aspects?

ZihanLi58: actually this should be a list of (aspect, path)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -514,6 +547,23 @@ public Map<String, Long> countAggregate(@Nullable IndexFilter indexFilter,
...
+   */
+  @Nullable
+  private String resolveForceIndex(@Nullable IndexFilter indexFilter) {
+    if (_forceIndexName == null || _forceIndexRequiredAspect == null) {
+      return null;
+    }
+    if (indexFilter == null || !indexFilter.hasCriteria()) {
+      return null;
+    }
+    boolean hasRequiredAspect = indexFilter.getCriteria().stream()
```
ZihanLi58: when resolve we need to make sure the filter contains all and only contains the aspect-path  we want to force on? i.e. only when filter is a_urn.model_urn and a_status.removed we will force the index, if it's only a_urn.model_urn or only a_status.removed or a_urn.model_urn + a_status.removed + other field, we should not use the force index?

NatalliaUlashchick: I am inclined towards the subset not the exact match. Reasons:
  - The framework always appends deleted_ts IS NULL and soft-delete checks — those aren't in IndexCriterion but are always in the WHERE clause. An exact match on IndexCriterion would still have "extra" WHERE conditions at the SQL level.
  - If someone adds an extra filter condition in the future, exact match would silently deactivate the force index, and the query goes back to timing out.                                                                                            
  - The composite index is still beneficial as long as both its key columns are present, regardless of extra WHERE conditions.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
NatalliaUlashchick: Agree, will add it in the next revision.

rakhiagr:
@NatalliaUlashchick QQ: Do we need `FORCE INDEX` on the COUNT query as well? The original issue seems specific to the data query with `ORDER BY/LIMIT/OFFSET`. Just want to confirm we’ve verified that forcing the same index on the COUNT path won’t  have any regression.

NatalliaUlashchick:
> @NatalliaUlashchick QQ: Do we need `FORCE INDEX` on the COUNT query as well? The original issue seems specific to the data query with `ORDER BY/LIMIT/OFFSET`. Just want to confirm we’ve verified that forcing the same index on the COUNT path won’t have any regression.

the COUNT query benefits from FORCE INDEX too. Both queries share the same WHERE clause (i_urn$model_urn IN (...) AND i_model_instance_status$status = 'ACTIVE' AND ...), and the composite index `idx_urn$model_urn$status ` covers exactly those columns. I verified that this index is correctly picked up even without using FORCE clause in prod db today.

The FORCE INDEX is derived from the same base SQL,  so the force carries over automatically. I don't see regression risk, since the composite index is the optimal access path for this WHERE clause shape regardless of whether  we're selecting rows or counting

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
ZihanLi58: this has a strong assumption that the forced index is the most correct index to use, what's the case that if forced index is status, while user actually want to do filter on status and ownership, in which case, ownership is a optimized index? 
Just want to call out the risk here, if we think the current approach is better, let's add comprehensive comments here to mention the side effects when you set the force index.

NatalliaUlashchick: This is a vaild concern. We should treat this as a surgical fix for one known-bad query shape, not a generic mechanism. If the query shape changes, the force index should deactivate and let MySQL decide. If MySQL still picks wrong for the new shape, we extend the solution then.                
 I refactored it to do the exact match.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -110,11 +113,43 @@ public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor)
     _urnPathExtractor = urnPathExtractor;
   }
 
+  @Override
+  public void configureOptionalForceIndex(@Nullable String indexName,
```
ZihanLi58, line 125: add comments here?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -110,11 +113,43 @@ public void setUrnPathExtractor(@Nonnull UrnPathExtractor<URN> urnPathExtractor)
...
+      @Nullable Map<Class<? extends RecordTemplate>, String> requiredCriteria) {
```
ZihanLi58, line 126: nit: do you want to support multi  paths for one same aspect? I think it's not needed for the current use case, so fine to not resolved it.

NatalliaUlashchick: Correct — Map<Class, String> limits to one path per aspect. Not needed for the current use case (each aspect in the mlmodelinstance filter has a single path). If a future use case needs multiple paths per aspect, we can change the value type to Set<String>. Leaving as-is to keep it simple.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
NatalliaUlashchick: copied the javadoc from IEbeanLocalAccess interface declaration.

ZihanLi58 (approved):

PR #628, opened 2026-07-03
fix(dao): use canonical DB-stored URN when emitting MAE events

mridul111998:
## :octocat: What is this change?
Use the canonical DB-stored URN (from `ExtraInfo.getUrn()`) when emitting MAE events in `BaseLocalDAO`, instead of the incoming MCE URN. This fix applies to **all entity types and all indexes** since it is in the core DAO layer.

## :point_right: Why is this required?
Two HDFS clusters (WAR and Holdem) emit MCEs with different casing for the same dataset path (e.g. `LixTrackingArchive` vs `lixTrackingArchive`). Because the Hosted Search document ID is derived from the URN, two different document IDs were created → duplicate browse results in Datahub.

The root cause: `addCommon()` was emitting MAE with the **incoming** MCE URN instead of the **canonical** URN stored in the DB. The DB (MySQL) normalises casing via CI collation, so there is only ever one row — but two different HS documents were being created.

## :white_check_mark: How was this tested?
- Added `testBothCaseVariantsProduceSameCanonicalUrnInMAE` in `BaseLocalDAOTest` using a string-keyed `BurgerUrn` to simulate two MCEs with different-case URNs arriving for the same entity. Verifies that both MAEs are emitted with the canonical (DB-stored) URN and the upper-case URN never appears.
- Updated `makeAspectEntry()` test helper in `BaseLocalDAOTest` and `BaseLocalDAOAspectVersionTest` to require an explicit `@Nullable Urn` parameter, mirroring production behaviour where `ExtraInfo` always carries the stored URN for existing entities. This eliminated `RequiredFieldNotPresentException` failures in existing tests.
- Consolidated duplicate `makeAspectEntry` helper into a single static method in `BaseLocalDAOTest`, shared via static import.
- Full `dao-api` test suite passes (244 tests).

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -126,11 +126,15 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     }
   }
 
-  @Value
+  @Getter
+  @AllArgsConstructor
   static class AddResult<ASPECT extends RecordTemplate> {
     ASPECT oldValue;
     ASPECT newValue;
     Class<ASPECT> klass;
+    // The canonical URN as stored in the DB (may differ in case from the incoming MCE URN).
+    // For new entities with no existing DB row, this is set to the incoming MCE URN.
+    @Nonnull Urn canonicalUrn;
   }
 
   @Value
@@ -184,7 +188,7 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   @Value
   @EqualsAndHashCode(callSuper = true)
   public static class AspectCreateLambda<ASPECT extends RecordTemplate> extends AspectUpdateLambda<ASPECT> {
-    
+
     public AspectCreateLambda(@Nonnull ASPECT value) {
       super(value); // Uses AspectUpdateLambda constructor that creates lambda: (ignored) -> value
     }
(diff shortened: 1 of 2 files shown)


dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -1043,21 +1057,26 @@ private <ASPECT extends RecordTemplate> ASPECT unwrapAddResult(URN urn, @Nonnull
...
+    // Falls back to the incoming urn for new entities — canonicalUrn is always non-null.
+    @SuppressWarnings("unchecked")
+    final URN maeUrn = (URN) result.getCanonicalUrn();
+
     // Invoke post-update hooks if there's any
     // Note that we do NOT support post-update (or pre-update) hooks for deletion operations (yet). However, since
     //   newValue can in theory be NULL outside of deletion operations, we need to check for that here.
     if (_aspectPostUpdateHooksMap.containsKey(aspectClass) && !isDeletion) {
-      _aspectPostUpdateHooksMap.get(aspectClass).forEach(hook -> hook.accept(urn, newValue));
+      _aspectPostUpdateHooksMap.get(aspectClass).forEach(hook -> hook.accept(maeUrn, newValue));
```
jphui, line 1069: IIUC this is the core change, rather than use the urn from the payload, use the urn from the DB

jphui (approved):
🚢 awesome fix

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
mridul111998: yes correct, and the main this is how we are passing this urn from all the places without adding a new extra DB call

yangyangv2 (commented):
Look good to me, thanks for making the fix!

Please fix the build failure and update the comments, then ship-it

PR #631, opened 2026-07-28
feat(dao-api): add pluggable usage events on DAO read, write, and delete paths

dannytrn:
## Summary

Adds an optional, pluggable usage-tracking layer to the Ebean DAO — one usage event per
successful read / write / delete — without the kernel depending on any concrete event schema.

- `BaseDaoUsageEmitter` (dao-api): model-agnostic interface (plain strings + `DaoUsageTarget`)
  with a `NoOpDaoUsageEmitter` default; contract is fire-and-forget (async, non-blocking, never
  throws).
- `UsageTrackingEbeanLocalAccess` (ebean-dao): decorator over `IEbeanLocalAccess` that delegates
  first and emits only after success; installed via `EbeanLocalDAO.setUsageEmitter` (mirrors
  `setBenchmarkMetrics`, no-op in `OLD_SCHEMA_ONLY`).
- `DaoReadContext`: thread-local marker so the internal read-before-write is not counted as a
  consumer read.

Purely additive: no signature changes, and no behavior change until an emitter is installed
(default no-op = zero overhead). Design doc: `docs/rfc/active/dao-usage-event-instrumentation.md`.

## Testing Done

- [x] Compile + Checkstyle (main & test) for `dao-api` and `ebean-dao`
- [x] Unit tests pass: `DaoReadContextTest` (3), `NoOpDaoUsageEmitterTest` (1), `UsageTrackingEbeanLocalAccessTest` (20)
- [x] `BaseLocalDAOTest` (68) passes — covers the modified batch read-before-write path
- [x] `./gradlew spotlessCheck`
- [x] Full `ebean-dao` DB-backed integration suite passes locally (embedded MariaDB per the Apple Silicon setup in `docs/developers.md`)

## Checklist

- [x] Docs related to the changes have been added/updated (RFC: `docs/rfc/active/dao-usage-event-instrumentation.md`)

### DIFF
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -2157,12 +2157,15 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
    * Similar to {@link #getWithExtraInfo(Set)} but only using only one {@link AspectKey}.
    */
   @Nonnull
+  @SuppressWarnings("unchecked")
   public <ASPECT extends RecordTemplate> Optional<AspectWithExtraInfo<ASPECT>> getWithExtraInfo(
       @Nonnull AspectKey<URN, ASPECT> key) {
-    if (getWithExtraInfo(Collections.singleton(key)).containsKey(key)) {
-      return Optional.of((AspectWithExtraInfo<ASPECT>) getWithExtraInfo(Collections.singleton(key)).get(key));
-    }
-    return Optional.empty();
+    // Query once and null-check the result. Calling getWithExtraInfo twice (once to test
+    // containsKey, once to read) issued a redundant round trip for every single-key read, and with
+    // usage tracking enabled it emitted two READ events for one logical read.
+    final AspectWithExtraInfo<? extends RecordTemplate> result =
+        getWithExtraInfo(Collections.singleton(key)).get(key);
+    return result == null ? Optional.empty() : Optional.of((AspectWithExtraInfo<ASPECT>) result);
   }
 
   /**
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -912,12 +912,9 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
 
     // Single batched query - uses existing infrastructure.
     // Mark as an internal read-before-write so usage instrumentation does not count it as a consumer read.
-    DaoReadContext.markInternalRead();
     final Map<AspectKey<URN, ? extends RecordTemplate>, AspectWithExtraInfo<? extends RecordTemplate>> results;
-    try {
+    try (DaoReadContext.Scope ignored = DaoReadContext.markInternalRead()) {
       results = getWithExtraInfo(keys);
-    } finally {
-      DaoReadContext.clear();
     }
 
     // Convert to class-based map for easier lookup
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -35,6 +35,7 @@ import com.linkedin.metadata.dao.retention.TimeBasedRetention;
 import com.linkedin.metadata.dao.retention.VersionBasedRetention;
 import com.linkedin.metadata.dao.storage.LocalDAOStorageConfig;
 import com.linkedin.metadata.dao.tracking.BaseTrackingManager;
+import com.linkedin.metadata.dao.tracking.DaoReadContext;
 import com.linkedin.metadata.dao.tracking.TrackingUtils;
 import com.linkedin.metadata.dao.urnpath.UrnPathExtractor;
 import com.linkedin.metadata.dao.utils.ModelUtils;
@@ -909,9 +910,15 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
         .map(lambda -> new AspectKey<>(lambda.getAspectClass(), urn, LATEST_VERSION))
         .collect(Collectors.toSet());
 
-    // Single batched query - uses existing infrastructure
-    Map<AspectKey<URN, ? extends RecordTemplate>, AspectWithExtraInfo<? extends RecordTemplate>> results =
-        getWithExtraInfo(keys);
+    // Single batched query - uses existing infrastructure.
+    // Mark as an internal read-before-write so usage instrumentation does not count it as a consumer read.
+    DaoReadContext.markInternalRead();
+    final Map<AspectKey<URN, ? extends RecordTemplate>, AspectWithExtraInfo<? extends RecordTemplate>> results;
+    try {
+      results = getWithExtraInfo(keys);
+    } finally {
+      DaoReadContext.clear();
+    }
 
     // Convert to class-based map for easier lookup
     Map<Class<? extends RecordTemplate>, AspectWithExtraInfo<RecordTemplate>> byClass = new HashMap<>();
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
@@ -1925,7 +1925,11 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   private <ASPECT extends RecordTemplate> Optional<ASPECT> backfill(@Nonnull BackfillMode mode,
       @Nonnull Class<ASPECT> aspectClass, @Nonnull URN urn) {
     checkValidAspect(aspectClass);
-    Optional<ASPECT> aspect = get(aspectClass, urn, LATEST_VERSION);
+    // Backfill is a system operation, not consumer traffic, so its read is not counted as usage.
+    final Optional<ASPECT> aspect;
+    try (DaoReadContext.Scope ignored = DaoReadContext.markInternalRead()) {
+      aspect = get(aspectClass, urn, LATEST_VERSION);
+    }
     aspect.ifPresent(value -> backfill(mode, value, urn));
     return aspect;
   }
@@ -1972,8 +1976,11 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
     Set<Class<? extends RecordTemplate>> aspectToBackfill =
         aspectClasses == null ? getValidAspectTypes(_aspectUnionClass) : aspectClasses;
     checkValidAspects(aspectToBackfill);
-    final Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> urnToAspects =
-        get(aspectToBackfill, urns);
+    // Backfill is a system operation, not consumer traffic, so its read is not counted as usage.
+    final Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> urnToAspects;
+    try (DaoReadContext.Scope ignored = DaoReadContext.markInternalRead()) {
+      urnToAspects = get(aspectToBackfill, urns);
+    }
     urnToAspects.forEach((urn, aspects) -> {
       aspects.forEach((aspectClass, aspect) -> aspect.ifPresent(value -> backfill(mode, value, urn)));
     });
@@ -2012,7 +2019,10 @@ public abstract class BaseLocalDAO<ASPECT_UNION extends UnionTemplate, URN exten
   public Map<URN, Map<Class<? extends RecordTemplate>, Optional<? extends RecordTemplate>>> backfillEntityTables(
       @Nonnull Set<Class<? extends RecordTemplate>> aspectClasses, @Nonnull Set<URN> urns) {
     urns.forEach(urn -> aspectClasses.forEach(aspect -> updateEntityTables(urn, aspect)));
-    return get(aspectClasses, urns);
+    // Backfill is a system operation, not consumer traffic, so its read is not counted as usage.
+    try (DaoReadContext.Scope ignored = DaoReadContext.markInternalRead()) {
+      return get(aspectClasses, urns);
+    }
   }
 
   /**
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
@@ -1,0 +1,46 @@
+package com.linkedin.metadata.dao.tracking;
+
+import java.util.List;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+
+/**
+ * Interface for emitting a Metadata Graph usage event once per successful DAO operation
+ * (read / write / delete). Emission is an observability side-channel and MUST be
+ * fire-and-forget: implementations are required to be asynchronous, non-blocking, and to
+ * never throw back into the DAO call path.
+ *
+ * <p>The kernel passes only neutral primitives ({@link String} and {@link DaoUsageTarget})
+ * so it does not depend on any Avro / event-schema module (keeps
+ * {@code checkGmaDoesNotDependOnModels} green). A no-op implementation
+ * ({@link NoOpDaoUsageEmitter}) lives in the kernel; the concrete producer-backed
+ * implementation lives in the service layer.
+ */
+public interface BaseDaoUsageEmitter {
+
+  /**
+   * Record a completed DAO usage operation. Implementations MUST NOT block and MUST NOT
+   * propagate exceptions to the caller.
+   *
+   * @param operationType    one of {@code "READ"}, {@code "WRITE"}, {@code "DELETE"},
+   *                         {@code "DELETE_ALL"}
+   * @param entityType       entity type derived from the URN class (e.g. {@code "dataset"})
+   * @param sourceOperation  pure DAO method name (e.g. {@code "batchGetUnion"})
+   * @param actorUrn         string form of the caller URN for writes; {@code null} for reads
+   *                         (no audit stamp is available on the read path in v1)
+   * @param impersonatorUrn  string form of the service-on-behalf-of URN, or {@code null}
+   * @param targets          per-URN targets ({@code {urn, aspects[]}}); {@code aspects} is
+   *                         empty for a whole-entity {@code DELETE_ALL}
+   */
+  void emit(@Nonnull String operationType, @Nonnull String entityType,
+      @Nonnull String sourceOperation, @Nullable String actorUrn,
+      @Nullable String impersonatorUrn, @Nonnull List<DaoUsageTarget> targets);
+
+  /**
+   * Whether usage emission is enabled. Callers short-circuit all instrumentation (building
+   * targets, deriving the caller, etc.) when this returns {@code false}, giving zero
+   * overhead in the default / disabled configuration.
+   */
+  boolean isEnabled();
+}
--- a/dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
+++ b/dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
@@ -29,17 +29,26 @@ public interface BaseDaoUsageEmitter {
    * @param actorUrn         string form of the caller URN for writes; {@code null} for reads
    *                         (no audit stamp is available on the read path in v1)
    * @param impersonatorUrn  string form of the service-on-behalf-of URN, or {@code null}
-   * @param targets          per-URN targets ({@code {urn, aspects[]}}); {@code aspects} is
-   *                         empty for a whole-entity {@code DELETE_ALL}
+   * @param targets          per-URN targets ({@code {urn, aspects[]}}); {@code aspects} may be
+   *                         empty (whole-entity {@code DELETE_ALL}, or a {@code create} with no
+   *                         aspects) -- use {@code operationType}, not an empty list, to identify
+   *                         the operation kind
    */
   void emit(@Nonnull String operationType, @Nonnull String entityType,
       @Nonnull String sourceOperation, @Nullable String actorUrn,
       @Nullable String impersonatorUrn, @Nonnull List<DaoUsageTarget> targets);
 
   /**
-   * Whether usage emission is enabled. Callers short-circuit all instrumentation (building
-   * targets, deriving the caller, etc.) when this returns {@code false}, giving zero
-   * overhead in the default / disabled configuration.
+   * Whether usage emission is enabled.
+   *
+   * <p>Invoked on every captured DAO operation, synchronously on the caller's thread and -- on
+   * write paths -- inside an open transaction. Implementations MUST therefore be a cheap local
+   * read, e.g. a {@code volatile boolean} field refreshed out of band. They MUST NOT perform
+   * I/O, acquire locks, or consult a remote config or feature-flag service on this call path.
+   *
+   * <p>Callers short-circuit all instrumentation (building targets, deriving the caller) when
+   * this returns {@code false}. The "zero overhead when disabled" property of the DAO depends
+   * on this method honoring the above.
    */
   boolean isEnabled();
 }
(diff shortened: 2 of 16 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
```diff
@@ -0,0 +1,340 @@
...
+
+  @Override
+  public <ASPECT_UNION extends RecordTemplate> int batchUpsert(@Nonnull URN urn,
+      @Nonnull List<BaseLocalDAO.AspectUpdateContext<RecordTemplate>> updateContexts,
+      @Nonnull AuditStamp auditStamp, @Nullable IngestionTrackingContext ingestionTrackingContext,
+      boolean isTestMode) {
+    final int result =
+        _delegate.batchUpsert(urn, updateContexts, auditStamp, ingestionTrackingContext, isTestMode);
+    if (result > 0 && _usageEmitter.isEnabled() && !isTestMode && !isBackfill(ingestionTrackingContext)) {
+      safeEmit(OP_WRITE, "batchUpsert", auditStamp,
```
NatalliaUlashchick: [EbeanLocalDAO.java#L797-L813 ](https://github.com/linkedin/datahub-gma/blob/a33aa45eef83cf7163ae1caf72c3f69bac16baf2/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java#L797-L813)— runInTransactionWithRetry at L799 wraps _localAccess.batchUpsert (L801, emits here), then handleRelationshipIngestion at L808 can throw and roll the whole transaction back. 

The decorator sits below the transaction boundary, so WRITE events are emitted for writes that later roll back, and re-emitted on every optimistic-locking retry.   

Same shape for other methods as well. Can we buffer targets per-transaction and flush after runInTransactionWithRetry ?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -825,7 +850,9 @@ public <ASPECT extends RecordTemplate> void updateEntityTables(@Nonnull URN urn,
       }
       AuditStamp auditStamp = makeAuditStamp(result);
       ASPECT aspect = toRecordTemplate(aspectClass, result).orElse(null);
-      _localAccess.add(urn, aspect, aspectClass, auditStamp, null, false);
+      // Mark as backfill so this old->new schema migration is not reported as organic usage.
```
NatalliaUlashchick, line 879: Could you implement the same strategy for backfil reads? 

[BaseLocalDAO.java#L1979 ](https://github.com/linkedin/datahub-gma/blob/a33aa45eef83cf7163ae1caf72c3f69bac16baf2/dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java#L1979)— backfill(...) calls get(aspectToBackfill, urns), the ordinary consumer read path, which flows into batchGetUnion and emits.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -590,6 +592,29 @@ public void setBenchmarkMetrics(@Nonnull BaseDaoBenchmarkMetrics metrics) {
...
+   * No-op when {@code _localAccess} is {@code null} (OLD_SCHEMA_ONLY mode).
+   *
+   * <p>Purely additive: when this is never called the DAO behaves exactly as before, and the
+   * emitter is fire-and-forget (never throws into the call path). Composes with
+   * {@link #setBenchmarkMetrics} -- both decorate {@code _localAccess}.
+   *
+   * @param usageEmitter the usage emitter implementation to use
+   */
+  public void setUsageEmitter(@Nonnull BaseDaoUsageEmitter usageEmitter) {
+    if (_localAccess instanceof UsageTrackingEbeanLocalAccess) {
```
NatalliaUlashchick: The instanceof guard inspects only the outermost decorator, but [setBenchmarkMetrics — L589-L593](https://github.com/linkedin/datahub-gma/blob/a33aa45eef83cf7163ae1caf72c3f69bac16baf2/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java#L589-L593) wraps the same field with InstrumentedEbeanLocalAccess, burying the usage decorator one layer down. 
so setUsageEmitter → setBenchmarkMetrics → setUsageEmitter double-wraps and emits every event twice.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
```diff
@@ -0,0 +1,340 @@
...
+   * @param delegate     the real local-access implementation to wrap
+   * @param usageEmitter the usage emitter (may be a no-op)
+   * @param urnClass     the URN class, used to derive the entity type name once at construction
+   */
+  public UsageTrackingEbeanLocalAccess(@Nonnull IEbeanLocalAccess<URN> delegate,
+      @Nonnull BaseDaoUsageEmitter usageEmitter, @Nonnull Class<URN> urnClass) {
+    _delegate = delegate;
+    _usageEmitter = usageEmitter;
+    // Same derivation as InstrumentedEbeanLocalAccess to keep the entity dimension consistent.
+    _entityType = urnClass.getSimpleName().replace("Urn", "").toLowerCase();
```
NatalliaUlashchick: let's use `urn.getEntityType()` instead

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/DaoReadContext.java
```diff
@@ -0,0 +1,46 @@
...
+  private static final ThreadLocal<Boolean> INTERNAL_READ = ThreadLocal.withInitial(() -> Boolean.FALSE);
+
+  private DaoReadContext() {
+  }
+
+  /**
+   * Marks the current thread as executing an internal read-before-write.
+   */
+  public static void markInternalRead() {
+    INTERNAL_READ.set(Boolean.TRUE);
```
NatalliaUlashchick: DaoReadContext  is not reentrant

 clear()  unconditionally calls  remove() . Today there is exactly one level of nesting at each of the two call sites (I traced  batchGetOldValuesWithExtraInfo → getWithExtraInfo → batchGet → batchGetHelper → _localAccess.batchGetUnion , and  getLatest → queryLatest , both leaf-synchronous), so it is correct right now.

But the invariant is: "no marked region may ever contain another marked region," enforced only by convention, across two modules ( dao-api  sets it,  ebean-dao  reads it), with  getWithExtraInfo  being a public overridable method. If anyone nests, the inner  clear()  silently unmarks the outer scope and internal reads start being billed as consumer reads — a silent data corruption with no test that would catch it.

Suggestion: make it save/restore, ideally as a scope:

```
public static Scope markInternalRead() {
  final boolean previous = isInternalRead();
  INTERNAL_READ.set(Boolean.TRUE);
  return () -> { if (previous) INTERNAL_READ.set(Boolean.TRUE); else INTERNAL_READ.remove(); };
}
public interface Scope extends AutoCloseable { @Override void close(); }
```

Call sites become  `try (DaoReadContext.Scope ignored = DaoReadContext.markInternalRead()) { ... } , `which also removes the mark/try/finally boilerplate.

Minor, same file:  `ThreadLocal.withInitial(...)`  installs a map entry on every  `get()` , so every request thread that ever performs a read permanently gains an entry it never had. Use a bare  `ThreadLocal<Boolean>`  and  `Boolean.TRUE.equals(INTERNAL_READ.get())`  so read-only threads never populate the map.

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/DaoUsageTarget.java
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * Simple names of the aspects touched for this URN. Empty for a whole-entity delete
+   * ({@code DELETE_ALL}) where no specific aspect applies.
+   */
+  @Nonnull
+  List<String> aspects;
+
+  public DaoUsageTarget(@Nonnull String urn, @Nonnull List<String> aspects) {
+    this.urn = urn;
+    this.aspects = Collections.unmodifiableList(aspects);
```
NatalliaUlashchick: `Collections.unmodifiableList(aspects)`  is a view, not a copy. Every current caller passes a freshly built list so it's safe today. `Collections.unmodifiableList(new ArrayList<>(aspects))`  would be a better  alternative

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
```diff
@@ -0,0 +1,340 @@
...
+  // ---------------------------------------------------------------------------------------
+
+  @Nonnull
+  @Override
+  public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnion(
+      @Nonnull List<AspectKey<URN, ? extends RecordTemplate>> keys, int keysCount, int position,
+      boolean includeSoftDeleted, boolean isTestMode) {
+    final List<EbeanMetadataAspect> result =
+        _delegate.batchGetUnion(keys, keysCount, position, includeSoftDeleted, isTestMode);
+    if (_usageEmitter.isEnabled() && !isTestMode && !DaoReadContext.isInternalRead()) {
```
NatalliaUlashchick, line 156: this method calls `DaoReadContext` but the others don't. Let's at least add a comment which explains this asymmetry in behaviour

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/DaoReadContext.java
dannytrn: Fixed this in [631f031](https://github.com/linkedin/datahub-gma/pull/631/commits/631f031ce913f3b28dbee8a1626b048ec2a8719c). 

You're right that this is not breaking today and nothing is nesting currently but definitely would've broken after I addressed the backfill comment. Suppressing backfill reads would introduce nesting and early clear().

I used your scope suggestion as-is, saving the previous value and restoring it on close instead of always clearing, so nesting is safe now. Also changed it to a plain `ThreadLocal` so null values don't get thrown onto the read path. Added a few tests as well.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
dannytrn: Fixed in [9016582](https://github.com/linkedin/datahub-gma/pull/631/commits/9016582b2b948986e6e690538848aa755749884a). You're right, the instanceof only ever sees the outer layer, so ssetUsageEmitter → setBenchmarkMetrics → setUsageEmitter stacked two decorators and doubled every event. I switched to a boolean flag on the DAO, which doesn't care about the layering order

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
dannytrn: done in [9016582](https://github.com/linkedin/datahub-gma/pull/631/commits/9016582b2b948986e6e690538848aa755749884a)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
dannytrn: Fixed [706a6f7](https://github.com/linkedin/datahub-gma/pull/631/commits/706a6f7d348c8a7caa2e8e77011a5d79c9941e7d). Backfill writes were already excluded via the tracking context but the reads weren't, so a backfill showed up as a spike of organic reads, which should be excluded.

Wrapped the read in four places rather than just the one you pointed at: both  `backfill`  overloads, `backfillEntityTables` , and  `backfillLocalRelationships`  in  `EbeanLocalDAO` . The rest (`backfillWithNewValue` ,  `backfillMAE` , the paged variant) all delegate into  `backfill(mode, Set, Set)` , so they're covered.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
dannytrn: I changed it to just have `list` check `DaoReadContext` as well, so read methods are symmetric

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
dannytrn: Yes - fixed this in [ab6cc3f](https://github.com/linkedin/datahub-gma/pull/631/commits/ab6cc3f7b9fda306a6591b43b1486d19a9ef483c). Writes now record their emission into a thread-local buffer that `runInTransactionWithRetry`  releases only after the commit, and each attempt drops what the previous one buffered. 

If the transaction rolls back, the buffer is thrown away and nothing is reported. Before each retry it clears what the previous attempt buffered, so a retried write is reported once instead of once per attempt. The buffer tracks depth and only flushes at the outermost commit.

dao-api/src/main/java/com/linkedin/metadata/dao/BaseLocalDAO.java
```diff
@@ -2160,12 +2167,15 @@ public abstract Map<String, Long> countAggregate(@Nullable IndexFilter indexFilt
...
   public <ASPECT extends RecordTemplate> Optional<AspectWithExtraInfo<ASPECT>> getWithExtraInfo(
       @Nonnull AspectKey<URN, ASPECT> key) {
-    if (getWithExtraInfo(Collections.singleton(key)).containsKey(key)) {
-      return Optional.of((AspectWithExtraInfo<ASPECT>) getWithExtraInfo(Collections.singleton(key)).get(key));
-    }
-    return Optional.empty();
+    // Query once and null-check the result. Calling getWithExtraInfo twice (once to test
+    // containsKey, once to read) issued a redundant round trip for every single-key read, and with
+    // usage tracking enabled it emitted two READ events for one logical read.
+    final AspectWithExtraInfo<? extends RecordTemplate> result =
```
NatalliaUlashchick, line 2176: Good catch on fixing extra DB call.

docs/rfc/active/dao-usage-event-instrumentation.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **Reads have no caller.** Writes carry an `AuditStamp`; reads do not, so every read event has `actorUrn = null`.
+- **Only new/dual-schema DAOs are instrumented.** The decorator wraps the new-schema access layer, so an
+  `OLD_SCHEMA_ONLY` DAO emits nothing, and historical (non-latest-version) reads served from the legacy path are not
+  emitted.
+- **Read counts are an upper bound on volume.** Internal read-before-writes are excluded via `DaoReadContext`, but the
+  uninstrumented legacy/historical paths above still make read counts incomplete — treat them as a lower bound on
+  distinct consumers.
+- **Writes reflect committed change, not attempts.** Emission is gated on affected-row count (`> 0`) and happens inside
+  the write transaction, so a rolled-back or retried write can over-count and a no-op upsert does not emit.
```
NatalliaUlashchick: I think this is already fixed. Let's fix the documentation

docs/rfc/active/dao-usage-event-instrumentation.md
```diff
@@ -0,0 +1,118 @@
+# DAO Usage Event Instrumentation
```
NatalliaUlashchick, line 1: Could you read proof the documentation agains the recent changes? Some points I noticed below:

• L44 still says entity type is "derived from the URN class simple name at construction (same as  InstrumentedEbeanLocalAccess )." Now false.
•  DaoUsageBuffer  is not mentioned anywhere — a new kernel class with savepoint-aware nesting semantics and a hard enter/exit pairing contract, absent from the design doc.
• The  DaoReadContext  section still describes the  clear() -in-a- finally  API rather than  Scope .
• Backfill read exclusion — a real behaviour change — isn't documented.
• The upper-bound/lower-bound contradiction at L98-100 is still there.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -605,24 +635,42 @@ public void ensureSchemaUpToDate() {
...
-        break;
-      } catch (RollbackException | DuplicateKeyException | OptimisticLockException exception) {
-        lastException = exception;
-      }
-    } while (++retryCount <= maxTransactionRetry);
+    boolean committed = false;
+    // Usage emissions from this transaction are buffered and released only after the outermost
+    // commit, so a rolled-back write is never reported and a retry is reported once.
+    // enter() and exit() must stay paired with nothing between them that can throw.
+    final int usageMark = DaoUsageBuffer.enter();
```
NatalliaUlashchick, line 648: "Zero overhead when disabled" is no longer strictly true (minor)

 `DaoUsageBuffer.enter()`  is unconditional. Every transactional write now costs two  ThreadLocal  operations plus a  Frame  +  ArrayList  allocation — even when no emitter was ever installed. The PR body claims "no behavior change until an emitter is installed (default no-op = zero overhead)."

In absolute terms this is tiny (two short-lived young-gen objects per transaction). I raise it because it contradicts the PR's own stated invariant, and that invariant is presumably what got this approved for a shared kernel. Cheap fixes: lazily allocate  pending  on first  record , or gate  `enter()`  on a static flag set by  `setUsageEmitter` .

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/DaoUsageBuffer.java
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * <p>Only the outermost frame releases anything: Ebean implements nested transactions with
+ * savepoints, so an inner commit is undone if the outer transaction rolls back. See
+ * {@code EbeanLocalDAO#runInTransactionWithRetry} for the enter/truncate/exit/flush sequence.
+ *
+ * <p>{@link #record} runs immediately when no frame is active, so non-transactional paths are
+ * unchanged. Reads are not buffered -- a read that was served happened regardless of whether a
+ * later write rolls back.
+ *
+ * <p>Callers MUST pair {@link #enter()} with {@link #exit} in a {@code finally}, with nothing that
```
NatalliaUlashchick, line 24: Please fix the java doc. It doesn't drop them —  `record()`  appends to  pending  (L109) and  depth  never returns to zero, so the list grows without bound on a pooled thread. That's a memory leak, not data loss.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccessTest.java
```diff
@@ -0,0 +1,517 @@
...
+import org.mockito.ArgumentCaptor;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static com.linkedin.testing.TestUtils.makeFooUrn;
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.*;
+
+
+/**
```
NatalliaUlashchick, line 32: Could you add a test to verify that  `result == 0`  suppresses emission — a documented behaviour on five methods, still uncovered.

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @param operationType    one of {@code "READ"}, {@code "WRITE"}, {@code "DELETE"},
+   *                         {@code "DELETE_ALL"}
+   * @param entityType       entity type derived from the URN class (e.g. {@code "dataset"})
+   * @param sourceOperation  pure DAO method name (e.g. {@code "batchGetUnion"})
+   * @param actorUrn         string form of the caller URN for writes; {@code null} for reads
+   *                         (no audit stamp is available on the read path in v1)
+   * @param impersonatorUrn  string form of the service-on-behalf-of URN, or {@code null}
+   * @param targets          per-URN targets ({@code {urn, aspects[]}}); {@code aspects} is
+   *                         empty for a whole-entity {@code DELETE_ALL}
```
NatalliaUlashchick: aspects  being empty isn't unique to  DELETE_ALL  —  create  with an empty  aspectValues  produces the same shape with  operationType = "WRITE" . It's reachable:  `createAspectsWithCallbacks`  has no empty-list guard (the validation loop just doesn't iterate) and  `prepareMultiColumnInsert`  handles an empty aspect list fine — it builds the base asset INSERT, returns 1, and clears the  result > 0  gate in  UsageTrackingEbeanLocalAccess#L215 .

No data problem today:  operationType  still distinguishes the two, and nothing consumes these events yet. But this wording invites a consumer to shape-infer with  if (aspects.isEmpty()) → whole-entity , which would misclassify a bare-asset create as an entity-level delete.

Suggest rewording to make  operationType  the authoritative discriminator:
 aspects  may be empty (whole-entity  DELETE_ALL , or a  create  with no aspects). Use  `operationType`  to identify the operation kind; do not infer it from an empty  aspects  list.
Same wording applies to  DaoUsageTarget.java:27-28  and the RFC at line 30

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  void emit(@Nonnull String operationType, @Nonnull String entityType,
+      @Nonnull String sourceOperation, @Nullable String actorUrn,
+      @Nullable String impersonatorUrn, @Nonnull List<DaoUsageTarget> targets);
+
+  /**
+   * Whether usage emission is enabled. Callers short-circuit all instrumentation (building
+   * targets, deriving the caller, etc.) when this returns {@code false}, giving zero
+   * overhead in the default / disabled configuration.
+   */
```
NatalliaUlashchick, line 52:
```suggestion
/**
 * Whether usage emission is enabled.
 *
 * <p>Invoked on every captured DAO operation, synchronously on the caller's thread and — on
 * write paths — inside an open transaction. Implementations MUST therefore be a cheap local
 * read, e.g. a {@code volatile boolean} field refreshed out of band. They MUST NOT perform
 * I/O, acquire locks, or consult a remote config or feature-flag service on this call path.
 *
 * <p>Callers short-circuit all instrumentation (building targets, deriving the caller) when
 * this returns {@code false}. The "zero overhead when disabled" property of the DAO depends
 * on this method honoring the above.
 */
```

docs/rfc/active/dao-usage-event-instrumentation.md
dannytrn: Fixed

docs/rfc/active/dao-usage-event-instrumentation.md
dannytrn: Fixed the entity-type line ( urn.getEntityType() ), rewrote the DaoReadContext section for the Scope API and added backfill-read exclusion, added a DaoUsageBuffer section

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
dannytrn: Added an armed flag set by setUsageEmitter ; enter() now returns before touching the thread-local or allocating a Frame until an emitter is installed, so the disabled path is zero-overhead again

dao-api/src/main/java/com/linkedin/metadata/dao/tracking/BaseDaoUsageEmitter.java
dannytrn: good catch, reworded so operationType is the authoritative discriminator and consumers shouldn't infer from an empty list. Applied the same to DaoUsageTarget and the RFC.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccessTest.java
dannytrn: Added new testZeroAffectedRowsSuppressesEmission that asserts no emit when the delegate returns 0, across all five gated methods (add, addWithOptimisticLocking, create, batchUpsert, softDeleteAsset)

NatalliaUlashchick (approved):

PR #632, opened 2026-07-29
feat(ebean-dao): add typed relationship keyset pagination

Ruturaj123:
## Summary

- Add typed keyset pagination for current relationship queries in `EbeanLocalRelationshipQueryDAO`.
- Bound each page by relationship ID using a fixed insertion high-water mark, stable ordering, and a positive page size.
- Enforce a maximum page size of 1,000 and support only new/dual schemas; old-schema callers fail explicitly.
- Preserve all existing offset-based APIs and centralize filters, schema naming, and index handling in the shared DAO.
- Document best-effort cross-page mutation consistency; rows inserted after page one are excluded, while concurrent updates or soft-deletes may affect later-page membership.

Related issue: [META-24159](https://linkedin.atlassian.net/browse/META-24159)

Follow-up changes are prepared separately:

1. V4 wrapper keyset pagination: `rgujar/add-v4-relationship-keyset-pagination`
2. Single-destination query optimization: `rgujar/optimize-single-destination-keyset`

## Testing Done

- `./gradlew :dao-impl:ebean-dao:test --tests com.linkedin.metadata.dao.localrelationship.EbeanLocalRelationshipQueryDAOTest`
- `./gradlew :dao-impl:ebean-dao:checkstyleMain`
- `./gradlew :dao-impl:ebean-dao:checkstyleTest`
- `./gradlew spotlessCheck`
- `EbeanLocalRelationshipQueryDAOTest`: 160 tests passed.
- Added TestNG coverage for empty/partial/exact/multi-page reads, insertion high-water behavior, concurrent replacement semantics, deleted-row gaps, typed entity and relationship filters, cursor validation, and defensive page behavior.

### Planned EI end-to-end validation

The production supernode from META-24159 does not exist in `ei-ltx1` or `ei4`, and no EI-origin equivalent was found. Its `DownstreamOf` degree is zero in EI.

After the V4 and GQS caller follow-ups are available, seed disposable data once in `ei-ltx1` using the synchronous metadata-store APIs:

1. Use timestamped EI-origin `external` dataset URNs:
   - `urn:li:dataset:(urn:li:dataPlatform:external,meta-24159.rgujar.<run-id>.hub,EI)`
   - `urn:li:dataset:(urn:li:dataPlatform:external,meta-24159.rgujar.<run-id>.child-0001,EI)`
2. Create the hub's default dataset-instance status with `DatasetInstances#setDatasetInstanceStatus(removed=false)`.
3. Create 1,001 children by calling `Datasets#setUpstreamLineage` once per child, with the hub as its single upstream dataset. This materializes 1,001 current `DownstreamOf` edges.
4. Write at no more than 5 requests/second with bounded concurrency and a persisted manifest of every test URN.
5. Query GQS with `stepMinDepth=1`, `stepMaxDepth=2`, `maxDepth=2`, and `maxNodes=1000`. Verify 1,000 unique results, bounded latency, two 500-row DAO pages, early termination, and no third page fetch.
6. Cleanup each child with `Datasets#overwriteUpstreamLineage(child, emptyLineage, false)`; a normal empty lineage write is not sufficient because actor-merge semantics can retain edges. Then mark child statuses removed and mark the hub removed last.
7. Verify the hub has zero active relationships through the DB-facing GQS API before completing cleanup.

This validates pagination and early-stop correctness, not 24.8M-edge production-scale performance.

### Rollout plan

This PR is additive: it introduces a new API and does not change any existing caller or offset-based query.

- V4 wrapper support and the single-destination query-plan optimization are separate follow-up changes.
- GQS does not currently have a keyset-path gate. Its caller PR must add a default-off CFG2 boolean that chooses exactly one path per request; do not shadow both paths because that would duplicate the expensive read.
- Add `pagesFetched`, `rowsFetched`, `pageSize`, and `stoppedEarly` logs/metrics in the GQS caller before EI validation.
- Validate QuickDeploy/D2 routing for the candidate build first, then run the synthetic EI test.
- Compare old and keyset paths through separate EI invocations on synthetic and known-low-degree queries, checking result sets and latency.
- Ramp the CFG2 gate to EI, then one production fabric/ring, while retaining the old path for immediate rollback.
- Ranked/non-current relationship queries continue using the existing API because their ranking semantics cannot be safely keyset-paged.
- Metadata-store adopts the typed API independently after the shared dependency is released.

## Checklist

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli)

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -56,6 +56,8 @@ public class EbeanLocalRelationshipQueryDAO {
   public static final String RELATIONSHIP_RETURN_TYPE = "relationship.return.type";
   public static final String MG_INTERNAL_ASSET_RELATIONSHIP_TYPE = "AssetRelationship.proto";
   private static final int FILTER_BATCH_SIZE = 200;
+  // Hard upper bound on keyset page size to keep per-page DB work bounded.
+  private static final int MAX_KEYSET_PAGE_SIZE = 1000;
   private static final String IDX_DESTINATION_DELETED_TS = "idx_destination_deleted_ts";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_destination_deleted_ts) ";
   private static final String DESTINATION_FIELD =  "destination";
@@ -280,6 +282,174 @@ public class EbeanLocalRelationshipQueryDAO {
         .map(row -> RecordUtils.toRecordTemplate(relationshipType, row.getString("metadata")))
         .collect(Collectors.toList());
   }
+
+  /**
+   * Keyset (seek) paginated, current-only ({@code deleted_ts IS NULL}) variant of
+   * {@link #findRelationships(Class, LocalRelationshipFilter, Class, LocalRelationshipFilter, Class,
+   * LocalRelationshipFilter, int, int)} that walks the matching set in bounded pages. Ranked/
+   * non-current pagination is unsupported because it could not bound the per-page DB work.
+   *
+   * <p>First page: pass {@code cursor = null}; the DAO captures {@code maxId}, the largest
+   * relationship row id when paging starts ({@code COALESCE(MAX(id), 0)}), and returns rows with
+   * {@code 0 < rt.id <= maxId} ordered by id, up to {@code pageSize}. Later inserts get larger ids
+   * and are excluded, keeping the scan finite. Continuation: pass the next cursor from the previous
+   * {@link RelationshipKeysetPage}. The combined pages are not a point-in-time snapshot: existing
+   * rows updated or soft-deleted between page calls can change which rows a later page returns (see
+   * {@link RelationshipKeysetCursor}).</p>
+   *
+   * <p>Supported only when {@code SchemaConfig} is {@code NEW_SCHEMA_ONLY} or {@code DUAL_SCHEMA};
+   * {@code OLD_SCHEMA_ONLY} throws {@link UnsupportedOperationException}.</p>
+   *
+   * @param sourceEntityClass the source entity class to query
+   * @param sourceEntityFilter the filter to apply to the source entity when querying
+   * @param destinationEntityClass the destination entity class
+   * @param destinationEntityFilter the filter to apply to the destination entity when querying
+   * @param relationshipType the type of relationship to query
+   * @param relationshipFilter the filter to apply to relationship when querying
+   * @param pageSize the maximum number of relationships to return per page. Must be between 1 and
+   *                 1000 inclusive.
+   * @param cursor the cursor from the previous page, or {@code null} for the first page.
+   * @return a page of relationship records plus a next cursor (null when the scan is exhausted).
+   * @throws UnsupportedOperationException when the DAO is in {@code OLD_SCHEMA_ONLY} mode.
+   */
+  @Nonnull
+  public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplate, RELATIONSHIP extends RecordTemplate>
+      RelationshipKeysetPage<RELATIONSHIP> findRelationshipsByKeyset(
+      @Nullable Class<SRC_SNAPSHOT> sourceEntityClass, @Nonnull LocalRelationshipFilter sourceEntityFilter,
+      @Nullable Class<DEST_SNAPSHOT> destinationEntityClass, @Nonnull LocalRelationshipFilter destinationEntityFilter,
+      @Nonnull Class<RELATIONSHIP> relationshipType, @Nonnull LocalRelationshipFilter relationshipFilter, int pageSize,
+      @Nullable RelationshipKeysetCursor cursor) {
+    validateEntityFilter(sourceEntityFilter, sourceEntityClass);
+    validateEntityFilter(destinationEntityFilter, destinationEntityClass);
+    validateRelationshipFilter(relationshipFilter, false);
+
+    String destTableName = null;
+    if (destinationEntityClass != null) {
+      destTableName = SQLSchemaUtils.getTableName(ModelUtils.getUrnTypeFromSnapshot(destinationEntityClass));
+    }
+
+    String sourceTableName = null;
+    if (sourceEntityClass != null) {
+      sourceTableName = SQLSchemaUtils.getTableName(ModelUtils.getUrnTypeFromSnapshot(sourceEntityClass));
+    }
+
+    final String relationshipTableName = SQLSchemaUtils.getRelationshipTableName(relationshipType);
+
+    final KeysetScanResult scan = findRelationshipsByKeysetCore(relationshipTableName, sourceTableName,
+        sourceEntityFilter, destTableName, destinationEntityFilter, relationshipFilter, pageSize, cursor);
+
+    final List<RELATIONSHIP> relationships = new ArrayList<>(scan.getRows().size());
+    for (SqlRow row : scan.getRows()) {
+      relationships.add(RecordUtils.toRecordTemplate(relationshipType, row.getString(METADATA)));
+    }
+
+    return new RelationshipKeysetPage<>(relationships, scan.getHighWaterId(), scan.getNextCursor());
+  }
+
+  /**
+   * Shared row-level keyset core for {@link #findRelationshipsByKeyset}: captures the first-page
+   * {@code maxId} (largest row id when paging starts), runs the current-only keyset SQL, validates
+   * strictly increasing id progress and computes the next cursor. Callers resolve their own table
+   * names/filters and map the returned rows. Supported only for {@code NEW_SCHEMA_ONLY} and
+   * {@code DUAL_SCHEMA}; {@code OLD_SCHEMA_ONLY} throws {@link UnsupportedOperationException} before
+   * any SQL is built or executed.
+   */
+  @Nonnull
+  private KeysetScanResult findRelationshipsByKeysetCore(@Nonnull final String relationshipTableName,
+      @Nullable final String sourceTableName, @Nullable final LocalRelationshipFilter sourceEntityFilter,
+      @Nullable final String destTableName, @Nullable final LocalRelationshipFilter destinationEntityFilter,
+      @Nonnull final LocalRelationshipFilter relationshipFilter, final int pageSize,
+      @Nullable final RelationshipKeysetCursor cursor) {
+    if (pageSize < 1 || pageSize > MAX_KEYSET_PAGE_SIZE) {
+      throw new IllegalArgumentException(
+          "pageSize must be between 1 and " + MAX_KEYSET_PAGE_SIZE + " but was " + pageSize);
+    }
+    if (_schemaConfig == EbeanLocalDAO.SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException(
+          "Keyset pagination is not supported in OLD_SCHEMA_ONLY mode; use NEW_SCHEMA_ONLY or DUAL_SCHEMA.");
+    }
+
+    final long lastId;
+    final long maxId;
+    if (cursor == null) {
+      lastId = 0L;
+      maxId = fetchHighWaterId(relationshipTableName);
+    } else {
+      lastId = cursor.getLastId();
+      maxId = cursor.getMaxId();
+    }
+
+    // Nothing left to scan (empty table, or the previous page reached maxId).
+    if (lastId >= maxId) {
+      return new KeysetScanResult(Collections.emptyList(), maxId, null);
+    }
+
+    final String sql = buildFindRelationshipKeysetSQL(relationshipTableName, relationshipFilter, sourceTableName,
+        sourceEntityFilter, destTableName, destinationEntityFilter, pageSize, lastId, maxId);
+
+    final List<SqlRow> rows = executeSqlWithIndexCheck(sql, relationshipTableName);
+
+    long previousId = lastId;
+    long lastRowId = lastId;
+    for (SqlRow row : rows) {
+      final long id = row.getLong("id");
+      if (id <= previousId) {
+        throw new IllegalStateException(String.format(
+            "Relationship ids must strictly increase during keyset pagination but saw id %d after %d in table '%s'",
+            id, previousId, relationshipTableName));
+      }
+      previousId = id;
+      lastRowId = id;
+    }
+
+    // A next cursor is only warranted when the page was full and did not reach maxId.
+    RelationshipKeysetCursor nextCursor = null;
+    if (rows.size() == pageSize && lastRowId < maxId) {
+      nextCursor = new RelationshipKeysetCursor(lastRowId, maxId);
+    }
+
+    return new KeysetScanResult(rows, maxId, nextCursor);
+  }
+
+  /**
+   * Immutable holder for one keyset page's raw rows, {@code maxId} and next cursor.
+   */
+  private static final class KeysetScanResult {
+    private final List<SqlRow> _rows;
+    private final long _highWaterId;
+    private final RelationshipKeysetCursor _nextCursor;
+
+    KeysetScanResult(@Nonnull List<SqlRow> rows, long highWaterId,
+        @Nullable RelationshipKeysetCursor nextCursor) {
+      _rows = rows;
+      _highWaterId = highWaterId;
+      _nextCursor = nextCursor;
+    }
+
+    @Nonnull
+    List<SqlRow> getRows() {
+      return _rows;
+    }
+
+    long getHighWaterId() {
+      return _highWaterId;
+    }
+
+    @Nullable
+    RelationshipKeysetCursor getNextCursor() {
+      return _nextCursor;
+    }
+  }
+
+  /**
+   * Returns COALESCE(MAX(id), 0), the largest relationship row id ({@code maxId}), for the table.
+   */
+  private long fetchHighWaterId(@Nonnull final String relationshipTableName) {
+    final String sql = "SELECT COALESCE(MAX(id), 0) AS max_id FROM " + relationshipTableName;
+    final List<SqlRow> rows = _server.createSqlQuery(sql).findList();
+    return rows.isEmpty() ? 0L : rows.get(0).getLong("max_id");
+  }
+
   /**
    * Finds a list of relationships of a specific type (Urn) based on the given filters if applicable.
    *
@@ -862,6 +1032,89 @@ public class EbeanLocalRelationshipQueryDAO {
     if (includeNonCurrentRelationships) {
       sqlBuilder.append(") ranked_rows WHERE row_num = 1");
     }
+
+    return sqlBuilder.toString();
+  }
+
+  /**
+   * Keyset (seek) pagination counterpart of {@link #buildFindRelationshipSQL}: same join/filter
+   * construction plus keyset bounds ({@code rt.id > lastId AND rt.id <= maxId}, {@code maxId} being
+   * the largest relationship row id when paging starts), ascending id order and
+   * {@code LIMIT pageSize}. Kept separate to leave {@link #buildFindRelationshipSQL} untouched.
+   *
+   * <p>Always current-only ({@code deleted_ts IS NULL}); ranked/non-current pagination is
+   * unsupported because it could not bound the per-page DB work.</p>
+   *
+   * <p>Supported only for {@code NEW_SCHEMA_ONLY} and {@code DUAL_SCHEMA}; {@code OLD_SCHEMA_ONLY}
+   * throws {@link UnsupportedOperationException}.</p>
+   */
+  @Nonnull
+  @VisibleForTesting
+  public String buildFindRelationshipKeysetSQL(@Nonnull final String relationshipTableName,
+      @Nonnull LocalRelationshipFilter relationshipFilter, @Nullable final String sourceTableName,
+      @Nullable LocalRelationshipFilter sourceEntityFilter, @Nullable final String destTableName,
+      @Nullable LocalRelationshipFilter destinationEntityFilter, int pageSize, long lastId, long maxId) {
+    if (pageSize < 1 || pageSize > MAX_KEYSET_PAGE_SIZE) {
+      throw new IllegalArgumentException(
+          "pageSize must be between 1 and " + MAX_KEYSET_PAGE_SIZE + " but was " + pageSize);
+    }
+    if (_schemaConfig == EbeanLocalDAO.SchemaConfig.OLD_SCHEMA_ONLY) {
+      throw new UnsupportedOperationException(
+          "Keyset pagination is not supported in OLD_SCHEMA_ONLY mode; use NEW_SCHEMA_ONLY or DUAL_SCHEMA.");
+    }
+
+    relationshipFilter = LogicalExpressionLocalRelationshipCriterionUtils.normalizeLocalRelationshipFilter(relationshipFilter);
+    sourceEntityFilter = LogicalExpressionLocalRelationshipCriterionUtils.normalizeLocalRelationshipFilter(sourceEntityFilter);
+    destinationEntityFilter = LogicalExpressionLocalRelationshipCriterionUtils.normalizeLocalRelationshipFilter(destinationEntityFilter);
+
+    // Cursor bounds are validated non-negative longs, so inlining them is safe and matches the
+    // existing flow that inlines filter values (including colon-bearing urns) without named params.
+    final String maxIdPredicate = "rt.id <= " + maxId;
+    final String lastIdPredicate = "rt.id > " + lastId;
+    final StringBuilder sqlBuilder = new StringBuilder();
+
+    sqlBuilder.append("SELECT rt.*");
+    sqlBuilder.append(" FROM ").append(relationshipTableName).append(" rt ");
+
+    final List<Triplet<LocalRelationshipFilter, String, String>> filters = new ArrayList<>();
+
+    if (_schemaConfig == EbeanLocalDAO.SchemaConfig.NEW_SCHEMA_ONLY || _schemaConfig == EbeanLocalDAO.SchemaConfig.DUAL_SCHEMA) {
+      if (destTableName != null) {
+        sqlBuilder.append("INNER JOIN ").append(destTableName).append(" dt ON dt.urn=rt.destination ");
+
+        if (destinationEntityFilter != null) {
+          filters.add(new Triplet<>(destinationEntityFilter, "dt", destTableName));
+        }
+      } else if (destinationEntityFilter != null) {
+        validateEntityFilterOnlyOneUrn(destinationEntityFilter);
+        filters.add(new Triplet<>(destinationEntityFilter, "rt", relationshipTableName));
+      }
+
+      if (sourceTableName != null) {
+        sqlBuilder.append("INNER JOIN ").append(sourceTableName).append(" st ON st.urn=rt.source ");
+
+        if (sourceEntityFilter != null) {
+          filters.add(new Triplet<>(sourceEntityFilter, "st", sourceTableName));
+        }
+      }
+
+      filters.add(new Triplet<>(relationshipFilter, "rt", relationshipTableName));
+
+      final String whereClause = SQLStatementUtils.whereClause(SUPPORTED_CONDITIONS,
+          _eBeanDAOConfig.isNonDollarVirtualColumnsEnabled(), _schemaValidatorUtil,
+          filters.toArray(new Triplet[filters.size()]));
+
+      sqlBuilder.append("WHERE rt.deleted_ts is NULL");
+      if (whereClause != null) {
+        sqlBuilder.append(" AND ").append(whereClause);
+      }
+      sqlBuilder.append(" AND ").append(lastIdPredicate).append(" AND ").append(maxIdPredicate);
+    } else {
+      // OLD_SCHEMA_ONLY is rejected above; only NEW_SCHEMA_ONLY and DUAL_SCHEMA are supported here.
+      throw new RuntimeException("The schema config must be set to DUAL_SCHEMA or NEW_SCHEMA_ONLY.");
+    }
+
+    sqlBuilder.append(" ORDER BY rt.id ASC LIMIT ").append(pageSize);
 
     return sqlBuilder.toString();
   }
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
@@ -1,0 +1,48 @@
+package com.linkedin.metadata.dao;
+
+import lombok.Getter;
+
+
+/**
+ * Immutable position of a keyset (seek) pagination scan over a single relationship table, used by
+ * {@link EbeanLocalRelationshipQueryDAO#findRelationshipsByKeyset}.
+ *
+ * <p>{@code lastId} is the id of the last row returned; the next page starts strictly after it
+ * ({@code rt.id > lastId}) and is {@code 0} for the first page. {@code maxId} is the largest
+ * relationship row id when paging starts, captured on the first page
+ * ({@code COALESCE(MAX(id), 0)}); every page is bounded by {@code rt.id <= maxId}. Later inserts get
+ * larger ids and are excluded, which keeps the scan finite. The combined pages are not a
+ * point-in-time snapshot: existing rows updated or soft-deleted between page calls can change which
+ * rows a later page returns, so callers must not assume they see every row that was current when
+ * paging started. Both values are non-negative and {@code lastId <= maxId} always holds.</p>
+ */
+public final class RelationshipKeysetCursor {
+
+  @Getter
+  private final long lastId;
+  @Getter
+  private final long maxId;
+
+  /**
+   * Creates a cursor.
+   *
+   * @param lastId id of the last row already returned; the next page starts strictly after it.
+   *               Must be non-negative.
+   * @param maxId largest relationship row id when paging starts; bounds the scan. Must be
+   *              non-negative and {@code >= lastId}.
+   */
+  public RelationshipKeysetCursor(long lastId, long maxId) {
+    if (lastId < 0) {
+      throw new IllegalArgumentException("lastId must be non-negative but was " + lastId);
+    }
+    if (maxId < 0) {
+      throw new IllegalArgumentException("maxId must be non-negative but was " + maxId);
+    }
+    if (lastId > maxId) {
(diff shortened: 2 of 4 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -280,6 +280,157 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+   * {@link #findRelationships(Class, LocalRelationshipFilter, Class, LocalRelationshipFilter, Class,
+   * LocalRelationshipFilter, int, int)} that walks the matching set in bounded pages. Ranked/
+   * non-current pagination is unsupported because it could not bound the per-page DB work.
+   *
+   * <p>First page: pass {@code cursor = null}; the DAO captures the insertion high-water id
+   * ({@code COALESCE(MAX(id), 0)}) and returns rows with {@code 0 < rt.id <= highWaterId} ordered by
+   * id, up to {@code pageSize}. Continuation: pass the {@link RelationshipKeysetPage#getNextCursor()
+   * next cursor}. Best-effort, not a snapshot: {@code maxId} excludes rows inserted after the first
+   * page, but concurrent updates/soft-deletes of rows with {@code id <= maxId} can change which rows
+   * a later page observes (see {@link RelationshipKeysetCursor}).</p>
```
jphui: To rephrase here for my understanding, this endpoint will not necessarily account for the case where deletions / insertions change the result set in which the pagination will be different? (ie "not a snapshot")

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -280,6 +280,157 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+      return _highWaterId;
+    }
+
+    @Nullable
+    RelationshipKeysetCursor getNextCursor() {
+      return _nextCursor;
+    }
+  }
+
+  /** Returns COALESCE(MAX(id), 0), the insertion high-water id, for the relationship table. */
```
jphui: I'm honestly really dumb, can you reword what this means? As a new developer with/without context of this repo, not sure what a insertion high-water id is perhaps in the context of the problem being solved here :) Thanks!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -866,6 +1017,100 @@ public String buildFindRelationshipSQL(@Nonnull final String relationshipTableNa
...
+      final String whereClause = SQLStatementUtils.whereClause(SUPPORTED_CONDITIONS,
+          _eBeanDAOConfig.isNonDollarVirtualColumnsEnabled(), _schemaValidatorUtil,
+          filters.toArray(new Triplet[filters.size()]));
+
+      sqlBuilder.append("WHERE rt.deleted_ts is NULL");
+      if (whereClause != null) {
+        sqlBuilder.append(" AND ").append(whereClause);
+      }
+      sqlBuilder.append(" AND ").append(lastIdPredicate).append(" AND ").append(maxIdPredicate);
+    } else if (_schemaConfig == EbeanLocalDAO.SchemaConfig.OLD_SCHEMA_ONLY) {
```
jphui: https://jarvis.corp.linkedin.com/codesearch/results?query=%22old_schema_only%22&repofacet=+linkedin-multiproduct/service-gms+&isFacet=true&current=1&nresults=10

This is the only case that even specifies OLD_SCHEMA_ONLY, I personally lean towards leaving this pathway un-implemented and just throwing here (AND add documentation).

To me the tradeoff is (yes) more compatibility with old systems but introduces more tech debt in the long run.

jphui: Good catch to ensure old pathways "should" be supported tjhough

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -280,6 +280,157 @@ public <SRC_SNAPSHOT extends RecordTemplate, DEST_SNAPSHOT extends RecordTemplat
...
+
+  /**
+   * Keyset (seek) paginated, current-only ({@code deleted_ts IS NULL}) variant of
+   * {@link #findRelationships(Class, LocalRelationshipFilter, Class, LocalRelationshipFilter, Class,
+   * LocalRelationshipFilter, int, int)} that walks the matching set in bounded pages. Ranked/
+   * non-current pagination is unsupported because it could not bound the per-page DB work.
+   *
+   * <p>First page: pass {@code cursor = null}; the DAO captures the insertion high-water id
+   * ({@code COALESCE(MAX(id), 0)}) and returns rows with {@code 0 < rt.id <= highWaterId} ordered by
+   * id, up to {@code pageSize}. Continuation: pass the {@link RelationshipKeysetPage#getNextCursor()
```
jphui: (AI found) No upper bound on  pageSize . A caller can pass Integer.MAX_VALUE  and defeat the bounding intent. Minor — consider a sane cap or a doc note.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
```diff
@@ -0,0 +1,95 @@
...
+      throw new IllegalArgumentException(
+          "highWaterId must be non-negative but was " + highWaterId);
+    }
+    if (nextCursor != null && nextCursor.getMaxId() != highWaterId) {
+      throw new IllegalArgumentException(
+          "nextCursor maxId (" + nextCursor.getMaxId() + ") must match highWaterId ("
+              + highWaterId + ")");
+    }
+    _relationships =
+        Collections.unmodifiableList(new java.util.ArrayList<>(relationships));
```
jphui: nit?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -866,6 +1017,100 @@ public String buildFindRelationshipSQL(@Nonnull final String relationshipTableNa
...
+   * construction plus keyset bounds ({@code rt.id > lastId AND rt.id <= maxId}, {@code maxId} being
+   * the insertion high-water mark), ascending id order and {@code LIMIT pageSize}. Kept separate to
+   * leave {@link #buildFindRelationshipSQL} untouched.
+   *
+   * <p>Always current-only ({@code deleted_ts IS NULL}); ranked/non-current pagination is
+   * unsupported because it could not bound the per-page DB work.</p>
+   */
+  @Nonnull
+  @VisibleForTesting
+  public String buildFindRelationshipKeysetSQL(@Nonnull final String relationshipTableName,
```
jphui, line 1053: (AI found) FORCE INDEX  dropped in the keyset builder. 

buildFindRelationshipSQL()  applies  FORCE INDEX (idx_destination_deleted_ts)  when filtering on  destination , and  executeSqlWithIndexCheck 's error message still references those index targets. The keyset builder never emits any FORCE INDEX. Combined with  ORDER BY rt.id ASC , a destination-filtered keyset scan may pick the PK order and filter-scan rows — the exact large-fan-out case pagination is meant to help. Consider mirroring the FORCE INDEX logic, or at least documenting why it's intentionally omitted.

jphui (commented):
=== 1 ===
Can you detail how you plan on testing this E2E? From the linked ticket it looks like there is mainly 1 (?) dataset where this is an issue?

Do you know if the shape of the problem is reproducible easily in EI? If so I'd recommend that since this could potentially break downstream consumers..

=== 2 ===
Same-ish question: how do you plan on ramping this?

Is the idea here that this new call replaces the old one? Will this be LIX'd then? Or are there special cases where this should be called instead of the old one, etc.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  public long getHighWaterId() {
+    return _highWaterId;
+  }
+
+  @Nullable
+  public RelationshipKeysetCursor getNextCursor() {
+    return _nextCursor;
+  }
```
jphui: I think these can be @Getter (?)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
```diff
@@ -0,0 +1,78 @@
...
+  public long getLastId() {
+    return _lastId;
+  }
+
+  public long getMaxId() {
+    return _maxId;
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
jphui: Where and how is this used in your logic?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public boolean hasNext() {
+    return _nextCursor != null;
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
jphui: ditto here... are you performing comparisons at the logic level?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(_lastId, _maxId);
+  }
+
+  @Nonnull
+  @Override
+  public String toString() {
```
jphui: same idea here... why is toString needed here?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(_relationships, _highWaterId, _nextCursor);
+  }
+
+  @Nonnull
+  @Override
+  public String toString() {
```
jphui: ditto

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+  public void testBuildFindRelationshipKeysetSQL() {
+    String sql = _localRelationshipQueryDAO.buildFindRelationshipKeysetSQL("relationship_table_name",
+        new LocalRelationshipFilter().setCriteria(new LocalRelationshipCriterionArray()).setDirection(RelationshipDirection.UNDIRECTED),
+        "source_table_name", null, "destination_table_name", null,
+        10, 5, 20);
+
+    assertEquals(sql,
+        "SELECT rt.* FROM relationship_table_name rt INNER JOIN destination_table_name dt ON dt.urn=rt.destination "
+            + "INNER JOIN source_table_name st ON st.urn=rt.source WHERE rt.deleted_ts is NULL "
+            + "AND rt.id > 5 AND rt.id <= 20 ORDER BY rt.id ASC LIMIT 10");
```
jphui, line 2248: I also hate to be pedantic here but can you add a small docstring that explains why this is the expected value in the context of your feature?

Especially highlight how this is DIFFERENT from the existing findRelationshipSql and what is new, effectively.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+
+    assertEquals(sql,
+        "SELECT rt.* FROM relationship_table_name rt INNER JOIN destination_table_name dt ON dt.urn=rt.destination "
+            + "INNER JOIN metadata_entity_foo st ON st.urn=rt.source WHERE rt.deleted_ts is NULL AND st.i_aspectfoo"
+            + (_eBeanDAOConfig.isNonDollarVirtualColumnsEnabled() ? "0" : "$") + "value='Alice' "
+            + "AND rt.id > 5 AND rt.id <= 20 ORDER BY rt.id ASC LIMIT 10");
+  }
+
+  @Test
+  public void testBuildFindRelationshipKeysetSQLRejectsNonPositivePageSize() {
```
jphui: nice edge case test 😎

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+    addReportsToChain(2, new FooUrn(1));
+    RelationshipKeysetPage<ReportsTo> page = keysetPage(5, null);
+    assertEquals(page.getRelationships().size(), 2);
+    assertFalse(page.hasNext());
+    assertNull(page.getNextCursor());
+    assertEquals(page.getHighWaterId(), 2L);
+  }
+
+  @Test
+  public void testFindRelationshipsByKeysetExactlyPageSize() throws URISyntaxException {
```
jphui, line 2329: good specificity in the testing 😎

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+    List<FooUrn> sources = addReportsToChain(3, dest);
+    // Re-adding source #2 and #3 soft-deletes ids 2 and 3 and inserts ids 4 and 5.
+    _localRelationshipWriterDAO.addRelationships(sources.get(1), AspectFoo.class,
+        Collections.singletonList(new ReportsTo().setSource(sources.get(1)).setDestination(new FooUrn(2))), false);
+    _localRelationshipWriterDAO.addRelationships(sources.get(2), AspectFoo.class,
+        Collections.singletonList(new ReportsTo().setSource(sources.get(2)).setDestination(new FooUrn(2))), false);
+
+    // Current rows toward dest #1: only id 1. High-water id is 5. A page size of 1 fills fully
+    // with id 1 (which is below maxId 5), so a next cursor is produced; the continuation query is
+    // empty because ids 2 and 3 are soft-deleted and 4/5 point elsewhere.
```
jphui: Just curious so if I am understanding the context of this setup properly...

relationships 1,2,3 exist

we delete 2 and 3 which cause 4 and 5 to be produced...

With your pagination logic, 4 and 5 won't show up with the "next cursor" since 2 and 3 are deleted (and are considered "next" semantically)...

if this is all correct, is this desired behavior given the context of the problem?

Why would we not want 4 and 5 to show up (since they are the next relationships after 1?

Is this what you meant by "best effort"?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+    assertEquals(first.getNextCursor().getMaxId(), 5L);
+
+    RelationshipKeysetPage<ReportsTo> second = _localRelationshipQueryDAO.findRelationshipsByKeyset(
+        null, emptyFilter(), null, emptyFilter(), ReportsTo.class, relationshipFilter, 1, first.getNextCursor());
+    assertTrue(second.getRelationships().isEmpty());
+    assertFalse(second.hasNext());
+  }
+
+  @Test
+  public void testFindRelationshipsByKeysetMultiPageOrderedComplete() throws URISyntaxException {
```
jphui, line 2374: How is this different than testFindRelationshipsByKeysetExactlyPageSize()?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+
+    // Capture the high-water id on the first page.
+    RelationshipKeysetCursor cursor = null;
+    List<ReportsTo> drained = new ArrayList<>();
+    RelationshipKeysetPage<ReportsTo> page = keysetPage(2, null);
+    drained.addAll(page.getRelationships());
+    cursor = page.getNextCursor();
+    assertEquals(page.getHighWaterId(), 5L);
+
+    // Insert more rows after the scan started; they must not be observed. Use distinct sources
```
jphui: "must not be observed" = best effort, right?

ie. "We'll do out best given the current state of doing the query, not necessarily aiming for consistency here"?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+      page = keysetPage(2, cursor);
+      drained.addAll(page.getRelationships());
+      cursor = page.getNextCursor();
+    }
+
+    assertEquals(drained.size(), 5);
+  }
+
+  @Test
+  public void testFindRelationshipsByKeysetSoftDeletedGaps() throws URISyntaxException {
```
jphui, line 2419: how is this different than testFindRelationshipsByKeysetFullPageBelowMaxIdYieldsFinalEmptyQuery()?

just like before there's an ID gap which you already tested?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+    RelationshipKeysetPage<ReportsTo> page = _localRelationshipQueryDAO.findRelationshipsByKeyset(
+        FooSnapshot.class, srcFilter, null, emptyFilter(), ReportsTo.class,
+        outgoingEmptyFilter(), 5, null);
+
+    assertEquals(page.getRelationships().size(), 1);
+    assertEquals(makeFooUrn(page.getRelationships().get(0).getSource().toString()), alice);
+  }
+
+  @Test
+  public void testFindRelationshipsByKeysetWithDestinationFilter() throws URISyntaxException {
```
jphui, line 2474: nice regression testing (I think that's what you're doing here?) 😎

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,388 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+  }
+
+  @Test
+  public void testRelationshipKeysetPageDefensiveAndValidation() throws URISyntaxException {
+    List<ReportsTo> source = new ArrayList<>();
+    source.add(new ReportsTo().setSource(new FooUrn(1)).setDestination(new FooUrn(2)));
+    RelationshipKeysetPage<ReportsTo> page =
+        new RelationshipKeysetPage<>(source, 10, new RelationshipKeysetCursor(5, 10));
+
+    // Defensive copy: mutating the source list does not affect the page.
```
jphui: is this just testing that mutating the input doesn't modify what was already returned by the API?

jphui: if so, clever!

jphui (commented):
Testing comments, some great edge case consideration here but some questions, too... I think I'm just OOTL on some of what I think are redundancies but not sure :/

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Correct. The first page records the largest existing row ID, excluding later inserts. Existing rows can still be updated or deleted between page calls, so later pages reflect those changes rather than a frozen point-in-time view

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Simplified the wording, it is just the largest relationship row ID when paging starts. Every page uses  id <= maxId , so rows inserted afterward are excluded and cannot extend the scan indefinitely

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Sure, makes sense. Updated keyset pagination to now explicitly reject `OLD_SCHEMA_ONLY` and documented support for only new/dual schemas. Existing non-keyset old-schema APIs remain unchanged.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Done, the API now accepts page sizes from 1 through 1,000

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
Ruturaj123: Thanks, updated!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: This was intentionally moved out of this PR to keep the foundation small. It is implemented in the separate follow-up branch `rgujar/optimize-single-destination-keyset`, with strict single-destination and index-existence guards

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
Ruturaj123: Updated

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
Ruturaj123: Good catch, its not being used currently. Removed!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
Ruturaj123: Removed

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
Ruturaj123: Removed

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java
Ruturaj123: Removed

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Added a comment. The new SQL adds lower/upper ID bounds, stable  ORDER BY id , and a finite  LIMIT

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Yes right

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Yes. It verifies that mutating the caller’s original list after constructing the page cannot change the page contents. It also verifies that the returned list itself is unmodifiable.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Yes, the interpretation is correct. The old row may disappear after being soft-deleted, while its replacement gets a new ID above the captured maximum and is excluded. 

Including them would allow continuous writes to keep extending the scan forever. The fixed maximum gives the traversal a stable endpoint. A later request will include those replacement rows.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: These cover different termination paths.

- `ExactlyPageSize`: three matching rows, page size three, and  maxId=3 . The last returned row reaches  maxId , so the DAO knows immediately there are no more rows and returns no cursor.
- `FullPageBelowMaxId`: one matching row fills the page, but unrelated rows make  maxId=5 . Because the last returned ID is below `maxId`, the DAO must return a cursor, the next empty page confirms no matching rows remain.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Rows inserted after paging begins are deterministically excluded, not best effort. Page one captures `maxId=5`, so every later page includes `id <= 5`, newly inserted IDs 6–8 cannot appear.

“Best effort” only applies to rows that already had IDs at or below `maxId`. Those existing rows may be updated or soft-deleted between page requests because each page is a separate SQL statement.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: The soft-deleted-gap test checks continuation through a gap in the middle. If current IDs are  `1, 3, 4` because ID `2` was deleted, pagination must return all three live rows without stopping at the missing ID.

The final-empty-page test checks a trailing filtered region. The page is full, but all remaining IDs up to `maxId` are deleted or do not match, so one additional empty query is required to establish completion

Ruturaj123:
> === 1 === Can you detail how you plan on testing this E2E? From the linked ticket it looks like there is mainly 1 (?) dataset where this is an issue?
> 
> Do you know if the shape of the problem is reproducible easily in EI? If so I'd recommend that since this could potentially break downstream consumers..
> 
> === 2 === Same-ish question: how do you plan on ramping this?
> 
> Is the idea here that this new call replaces the old one? Will this be LIX'd then? Or are there special cases where this should be called instead of the old one, etc.

Added details in PR description. But mainly:
The production supernode does not exist in EI, so we’ll programmatically create a disposable EI hub with 1,001 child relationships, validate the GQS caller through QD, and clean up afterward.

This PR only adds the DAO API. The separate GQS caller change will use a default-off CFG2 flag, keeping the old path available for rollback.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jphui: ah ok thanks!

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
jphui: ah ok cool, thanks for that!

jphui (approved):
🚢 , thanks for adding stuff esp the e2e testing methodology

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2190,4 +2192,414 @@ public void testFindRelationshipsWithSingleQuoteInUrn() throws Exception {
...
+    assertEquals(first.getHighWaterId(), 4L);
+    assertNotNull(first.getNextCursor());
+    List<ReportsTo> drained = new ArrayList<>(first.getRelationships());
+
+    // Between pages, "concurrently" replace c: this soft-deletes id 3 (which is <= maxId) and
+    // inserts a fresh current row id 5 (which is > maxId). The replacement row is excluded by the
+    // fixed maxId bound and the original id 3 is soft-deleted, so c may disappear from the scan even
+    // though it was current when the scan began. Because each page is a separate statement, maxId is
+    // only an insertion high-water mark, not a stable/complete snapshot: this is the documented
+    // best-effort membership behavior, chosen so the scan stays finite.
```
NatalliaUlashchick, line 2554: I'm slightly concerned about rows dropping out of the result set mid-scan. I understand we can't scan indefinitely and need some bound, but could we return a consistent snapshot as of a point in time instead?

The reason I think this will come up in practice: updates to relationship tables are done by soft-deleting the old row and inserting a new one. If that happens while a scan is in progress, the old row is excluded by  `deleted_ts IS NULL` , and its replacement gets an id above  maxId  so it's excluded by the ceiling. The relationship ends up in neither page, and the caller has no way to tell it was skipped.

It is documented in the javadoc, but I don't think consumers will reliably notice, and this feels like the kind of thing that surfaces much later as a hard-to-trace inconsistency.

Could we match rows that were alive when the scan started, rather than rows that are alive right now? That would mean storing the scan start time in the cursor alongside  lastId  and  maxId , and changing the filter:

from
```
WHERE rt.deleted_ts IS NULL
```

to
```
WHERE (rt.deleted_ts IS NULL OR rt.deleted_ts > '<scanStartTs>')
```

This works because relationship rows only ever go from alive to deleted, so  `deleted_ts > scanStartTs ` is a reliable "was alive at that point" test.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
```diff
@@ -0,0 +1,48 @@
+package com.linkedin.metadata.dao;
+
+import lombok.Getter;
+
+
+/**
+ * Immutable position of a keyset (seek) pagination scan over a single relationship table, used by
```
NatalliaUlashchick, line 7: The documentation says that it scans over a single relationship table, but  currently RelationshipKeysetCursor is a bare  (lastId, maxId) . With both  findRelationshipsByKeyset  and  findRelationshipsV4ByKeyset accepting it, mixing cursors between scans compiles, runs, and returns wrong results. Is it possible to  carry the table/type in the cursor and validate it ?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Makes sense, `deleted_ts > scanStartTs` is a good idea. Updated in https://github.com/linkedin/datahub-gma/pull/636/changes

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetCursor.java
Ruturaj123: Added https://github.com/linkedin/datahub-gma/pull/636/changes

PR #636, opened 2026-08-06
fix(ebean-dao): stop relationship keyset scans dropping rewritten rows

Ruturaj123:
## Summary

Follow-up to #632, addressing both post-merge review comments.

- Keyset scans now use scan-start membership: a row is returned if it was alive when the scan began. The cursor carries the scan start time, and each page merges two queries by id — current rows (`deleted_ts IS NULL`) and rows deleted after scan start (`deleted_ts > scanStartTime`).
- Relationship soft-deletes use `NOW(6)` instead of `NOW()`, in `DELETE_BY_SOURCE` and `DELETE_BY_SOURCE_AND_ASPECT` only. Asset soft-deletes are untouched.
- `RelationshipKeysetCursor` carries the relationship table name and rejects reuse against a different table. All four fields are required; the two-argument constructor is removed.
- Renames the "high-water" terminology to `maxId`, which the code already used for the same value.

`addRelationships` soft-deletes existing edges and then plain-`INSERT`s the replacements, so a relationship rewritten mid-scan failed `deleted_ts IS NULL` on its old row and `id <= maxId` on its new one, and appeared on no page at all.

`NOW()` truncates `deleted_ts` to the second while scan start is captured at full precision, so a delete at `10:00:00.500` recorded as `10:00:00.000000` looks earlier than a scan started at `10:00:00.123456` — the same vanishing-row bug inside a sub-second window.

Related issue: [META-24161](https://linkedin.atlassian.net/browse/META-24161)

### Why two queries instead of one `OR`

`(deleted_ts IS NULL OR deleted_ts > scanStartTime)` spans two disjoint `deleted_ts` ranges, so MySQL can no longer walk the index in `id` order and must sort, materializing every edge for that destination before `LIMIT` applies. On a 500k-row fixture with ~49.5k destinations and one 5,000-edge supernode: 1.59ms / 1,000 rows scanned for the current predicate, versus 8.19ms / 5,000 rows with the `OR`. The gap scales with supernode size, which is the case this work exists for.

Query A runs before Query B deliberately. A row soft-deleted between the two is returned by both, so the merge dedups equal primary-key ids; running B first would turn that race into a dropped row instead.

### Known limitations

- Not fully transactional. `deleted_ts` is statement time but rows become visible at commit, so a writer that deletes at T0, with a scan starting at T1 > T0 and committing at T2 > T1, still loses the edge. This exists on master today and is narrowed rather than closed here; fixing it needs stable logical relationship ids or a commit-ordered version marker.
- Cursor validation is table-scoped only. A cursor reused across the same table with different filters is still accepted. Happy to add a scan fingerprint if reviewers want that closed.
- The `NOW(6)` change is not yet covered by a test that fails without it; truncation causes under-inclusion, not the duplication the current test checks. Adding that next.

## Testing Done

- `./gradlew :dao-impl:ebean-dao:test --tests com.linkedin.metadata.dao.localrelationship.EbeanLocalRelationshipQueryDAOTest` — 176 tests passed
- `./gradlew :dao-impl:ebean-dao:compileJava :dao-impl:ebean-dao:compileTestJava`
- `git diff --check`
- Tests run on linux/amd64; the embedded MariaDB4j harness does not run on macOS hosts.
- New coverage for the vanishing-row bug, the Query A/B race (deterministic via a test seam rather than timing), rewrite-immediately-before-scan, cursor table scoping, and the generated SQL for both queries.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -349,7 +349,7 @@ public class EbeanLocalRelationshipQueryDAO {
       relationships.add(RecordUtils.toRecordTemplate(relationshipType, row.getString(METADATA)));
     }
 
-    return new RelationshipKeysetPage<>(relationships, scan.getHighWaterId(), scan.getNextCursor());
+    return new RelationshipKeysetPage<>(relationships, scan.getMaxId(), scan.getNextCursor());
   }
 
   /**
@@ -414,7 +414,7 @@ public class EbeanLocalRelationshipQueryDAO {
           relationshipType, assetRelationshipClass, row.getString(METADATA), row.getString(SOURCE), wrapOptions));
     }
 
-    return new RelationshipKeysetPage<>(relationships, scan.getHighWaterId(), scan.getNextCursor());
+    return new RelationshipKeysetPage<>(relationships, scan.getMaxId(), scan.getNextCursor());
   }
 
   /**
@@ -446,9 +446,9 @@ public class EbeanLocalRelationshipQueryDAO {
     final Timestamp scanStartTime;
     if (cursor == null) {
       lastId = 0L;
-      final KeysetHighWaterMark highWaterMark = fetchHighWaterMark(relationshipTableName);
-      maxId = highWaterMark.getMaxId();
-      scanStartTime = highWaterMark.getScanStartTime();
+      final KeysetScanStart scanStart = fetchScanStart(relationshipTableName);
+      maxId = scanStart.getMaxId();
+      scanStartTime = scanStart.getScanStartTime();
     } else {
       validateKeysetCursorTable(cursor, relationshipTableName);
       lastId = cursor.getLastId();
@@ -543,13 +543,13 @@ public class EbeanLocalRelationshipQueryDAO {
   }
 
   /**
-   * Immutable holder for one keyset scan's insertion high-water id and database start time.
+   * Immutable holder for one keyset scan's largest row id and database start time.
    */
-  private static final class KeysetHighWaterMark {
+  private static final class KeysetScanStart {
     private final long _maxId;
     private final Timestamp _scanStartTime;
 
-    KeysetHighWaterMark(long maxId, @Nonnull Timestamp scanStartTime) {
+    KeysetScanStart(long maxId, @Nonnull Timestamp scanStartTime) {
       _maxId = maxId;
       _scanStartTime = copyTimestamp(scanStartTime);
     }
@@ -569,13 +569,13 @@ public class EbeanLocalRelationshipQueryDAO {
    */
   private static final class KeysetScanResult {
     private final List<SqlRow> _rows;
-    private final long _highWaterId;
+    private final long _maxId;
     private final RelationshipKeysetCursor _nextCursor;
 
-    KeysetScanResult(@Nonnull List<SqlRow> rows, long highWaterId,
+    KeysetScanResult(@Nonnull List<SqlRow> rows, long maxId,
         @Nullable RelationshipKeysetCursor nextCursor) {
       _rows = rows;
-      _highWaterId = highWaterId;
+      _maxId = maxId;
       _nextCursor = nextCursor;
     }
 
@@ -584,8 +584,8 @@ public class EbeanLocalRelationshipQueryDAO {
       return _rows;
     }
 
-    long getHighWaterId() {
-      return _highWaterId;
+    long getMaxId() {
+      return _maxId;
     }
 
     @Nullable
@@ -596,18 +596,18 @@ public class EbeanLocalRelationshipQueryDAO {
 
   /**
    * Returns the largest relationship row id ({@code maxId}) and the database timestamp for the
-   * start of the scan. Both values come from one statement so the timestamp and high-water id are
+   * start of the scan. Both values come from one statement so the timestamp and max id are
    * captured against the same DB clock.
    */
-  private KeysetHighWaterMark fetchHighWaterMark(@Nonnull final String relationshipTableName) {
+  private KeysetScanStart fetchScanStart(@Nonnull final String relationshipTableName) {
     final String sql =
         "SELECT COALESCE(MAX(id), 0) AS max_id, DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.%f') AS scan_start_time FROM "
             + relationshipTableName;
     final List<SqlRow> rows = _server.createSqlQuery(sql).findList();
     if (rows.isEmpty()) {
-      throw new IllegalStateException("High-water query returned no rows for table " + relationshipTableName);
+      throw new IllegalStateException("Scan-start query returned no rows for table " + relationshipTableName);
     }
-    return new KeysetHighWaterMark(rows.get(0).getLong("max_id"),
+    return new KeysetScanStart(rows.get(0).getLong("max_id"),
         Timestamp.valueOf(rows.get(0).getString("scan_start_time")));
   }
 
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -25,8 +25,6 @@ import io.ebean.EbeanServer;
 import io.ebean.SqlQuery;
 import io.ebean.SqlRow;
 import java.sql.SQLException;
-import java.sql.Timestamp;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -64,8 +62,6 @@ public class EbeanLocalRelationshipQueryDAO {
   private static final String IDX_DESTINATION_DELETED_TS = "idx_destination_deleted_ts";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_destination_deleted_ts) ";
   private static final String DESTINATION_FIELD =  "destination";
-  private static final DateTimeFormatter MYSQL_TIMESTAMP_6_FORMATTER =
-      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
   private final EbeanServer _server;
   private final MultiHopsTraversalSqlGenerator _sqlGenerator;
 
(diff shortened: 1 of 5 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -436,23 +443,39 @@ private KeysetScanResult findRelationshipsByKeysetCore(@Nonnull final String rel
...
-        sourceEntityFilter, destTableName, destinationEntityFilter, pageSize, lastId, maxId);
-
-    final List<SqlRow> rows = executeSqlWithIndexCheck(sql, relationshipTableName);
+    final String currentSql = buildFindRelationshipKeysetCurrentSQL(relationshipTableName, relationshipFilter,
+        sourceTableName, sourceEntityFilter, destTableName, destinationEntityFilter, pageSize, lastId, maxId);
+
+    // Query current rows first. If a row is soft-deleted between the two reads, Query B may also see
+    // the same id; mergeKeysetRows dedups that id. Running B first could turn the race into a drop.
+    final List<SqlRow> currentRows = executeSqlWithIndexCheck(currentSql, relationshipTableName);
+    afterKeysetCurrentRowsFetched(relationshipTableName, currentRows);
```
NatalliaUlashchick, line 469: If there is no way to avoid having test seam in production api, could we at least add a short comment with the explanation here?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1220,19 +1346,46 @@ private List<SqlRow> executeSqlWithIndexCheck(String sql, String relationshipTab
...
+      SqlQuery query = _server.createSqlQuery(sql);
+      query.setParameter("scanStartTime", formatTimestamp6(scanStartTime));
+      return query.findList();
+    } catch (PersistenceException e) {
+      handleSqlExecutionException(e, relationshipTableName);
       throw new RuntimeException("Failed to execute SQL query for relationships", e);
     }
   }
 
+  private void handleSqlExecutionException(PersistenceException e, String relationshipTableName) {
```
NatalliaUlashchick: nit: rename to something like `throwIfMissingIndex` would give better clarity what the method does

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1220,19 +1346,46 @@ private List<SqlRow> executeSqlWithIndexCheck(String sql, String relationshipTab
...
+              + "Make sure FORCE INDEX targets like idx_destination_deleted_ts or idx_source_deleted_ts are created.",
+          relationshipTableName);
+      log.error(errorMsg);
+      throw new IllegalStateException(errorMsg, e);
+    }
+  }
+
+  @Nonnull
+  private static Timestamp copyTimestamp(@Nonnull Timestamp timestamp) {
+    Timestamp copy = new Timestamp(timestamp.getTime());
```
NatalliaUlashchick: duplicate. it already exists in `RelationshipKeysetCursor`

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -507,12 +595,20 @@ RelationshipKeysetCursor getNextCursor() {
...
+    final String sql =
+        "SELECT COALESCE(MAX(id), 0) AS max_id, DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.%f') AS scan_start_time FROM "
+            + relationshipTableName;
     final List<SqlRow> rows = _server.createSqlQuery(sql).findList();
-    return rows.isEmpty() ? 0L : rows.get(0).getLong("max_id");
+    if (rows.isEmpty()) {
+      throw new IllegalStateException("Scan-start query returned no rows for table " + relationshipTableName);
+    }
+    return new KeysetScanStart(rows.get(0).getLong("max_id"),
+        Timestamp.valueOf(rows.get(0).getString("scan_start_time")));
```
NatalliaUlashchick: we convert it to Timestamp on read, and later convert it back to string in `formatTimestamp6`.  Does it make sense to carrying it as a  String  end-to-end  to avoid error prone conversions?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -507,12 +595,20 @@ RelationshipKeysetCursor getNextCursor() {
...
+   * start of the scan. Both values come from one statement so the timestamp and max id are
+   * captured against the same DB clock.
    */
-  private long fetchHighWaterId(@Nonnull final String relationshipTableName) {
-    final String sql = "SELECT COALESCE(MAX(id), 0) AS max_id FROM " + relationshipTableName;
+  private KeysetScanStart fetchScanStart(@Nonnull final String relationshipTableName) {
+    final String sql =
+        "SELECT COALESCE(MAX(id), 0) AS max_id, DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.%f') AS scan_start_time FROM "
+            + relationshipTableName;
     final List<SqlRow> rows = _server.createSqlQuery(sql).findList();
```
NatalliaUlashchick: should we use `executeSqlWithIndexCheck` here as well ?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -507,12 +595,20 @@ RelationshipKeysetCursor getNextCursor() {
...
-    return rows.isEmpty() ? 0L : rows.get(0).getLong("max_id");
+    if (rows.isEmpty()) {
+      throw new IllegalStateException("Scan-start query returned no rows for table " + relationshipTableName);
```
NatalliaUlashchick: unreachable path. the query above never returns 0 rows

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
NatalliaUlashchick, line 1: Could you add test cases for merge's deferral path ?

Every new test has the merged result fitting inside one page. The inverted membership test does exercise interleaving (page 2 gets  c  from B and  d  from A), but never truncation — the case where A returns a full page, B contributes lower ids, and A's rows get pushed to the next page. That's precisely where  `deletedUpperId`  and the  `merged.size() < pageSize ` loop bound interact, and it's the part most likely to break under future edits.

Concrete scenario:

```
6 rows, ids 1..6, all alive. Start scan (maxId=6, scanStart=T).
Soft-delete ids 1 and 2 (deleted_ts > T).
Page 1, pageSize=2:
  A → {3,4}, size == pageSize, so deletedUpperId = 4
  B → {1,2}
  merged → {1,2}; ids 3,4 deferred
Page 2 → {3,4}
Page 3 → {5,6}
Assert: exactly {1..6}, no duplicates.
```

Without this, a regression in  deletedUpperId  or the merge bound would silently drop or duplicate rows and every existing test would still pass.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Added a comment

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Done

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Removed

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Makes sense, updated. Scan start is now a string from SQL through the cursor to the bound param, no conversions. Deleted `copyTimestamp`, `formatTimestamp6`, and the formatter.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: `executeSqlWithIndexCheck` is a wrapper which only helps in cases of missing index. This query has no WHERE and no FORCE INDEX. So the missing index error won't happen here.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Removed

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Added your scenario, plus a few nearby ones: B alone filling a page, deleted ids on both sides of A's last id, and a last page ending exactly at maxId. Each one drains every page and checks we got all rows back with no repeats.

While writing these I noticed `deletedUpperId` isn't actually needed for correctness. Query B already does `ORDER BY id LIMIT pageSize`, so it only ever returns its lowest ids. Capping it at A's last id just removes high ids that the merge would drop anyway. Relabelled the assertion so it doesn't read as a correctness guard.

NatalliaUlashchick (approved):
LGTM. Approving on the assumption the deferred build CI  check will pass.

PR #637, opened 2026-08-20
perf(ebean-dao): optimize single-destination keyset scans

Ruturaj123:
### Why

`buildFindRelationshipSQL` applies `FORCE INDEX (idx_destination_deleted_ts)` when the query filters on
`destination`. The keyset builders added for META-24161 never emitted it, so the keyset path lost a hint the
legacy path had.

It matters more on the keyset path than it did on the legacy one. Keyset paging adds `rt.id > ? AND rt.id <= ?`
plus `ORDER BY rt.id ASC`, which gives the optimizer a primary-key option the unbounded query never had.

Verified on the corp `metagalaxy` replica against `metadata_relationship_downstreamof`, first page
of a real reverse-lineage read:

|  | `key` | `rows` | `filtered` | `Extra` |
| --- | --- | --- | --- | --- |
| without hint | `PRIMARY` | 76,240,730 | 0 | Using where |
| with hint | `idx_destination_deleted_ts` | 39,597,970 | 100 | Using index condition |

Without the hint the optimizer walks the primary key in id order and expects roughly none of the rows it
visits to match the destination, so it cannot stop early. With the hint it seeks into the destination's index
entries, every row it touches is a match, and because InnoDB suffixes secondary indexes with the primary key
the index also satisfies `ORDER BY rt.id ASC`, so `LIMIT` stops the scan without a filesort.

Measured on the same replica, the plan the optimizer picks is not stable across destinations: three different
destinations produced a primary-key range scan, an index-merge intersect, and the destination index. Forcing
the index was the fastest of the three in every case and never slower.

### What

Emit the existing `FORCE_IDX_ON_DESTINATION` hint from the keyset builder, so it covers both statements each
page issues (current rows, and rows soft-deleted since scan start).

The hint is applied only when both hold:

- the destination is pinned to one urn by a direct, non-negated single-urn leaf (`EQUAL`, or `IN` with one
  value) on either the destination entity filter or the relationship filter, and
- `SchemaValidatorUtil.indexExists` reports the index is present on that relationship table.

Joins and filters are unchanged. Tables without the index, multi-urn filters and negated leaves are unaffected.

Bug: [META-24159](https://linkedin.atlassian.net/browse/META-24159)

## Testing Done

- New unit tests cover: `EQUAL` and single-value `IN` leaves emit the hint; multi-value `IN`, negated leaves and
  a missing index do not; the hint is placed immediately after the relationship table alias; and both the
  destination-entity-table and no-join shapes are covered. `SchemaValidatorUtil` is mocked so index presence is
  driven deterministically rather than depending on the embedded database.
- `./gradlew :dao-impl:ebean-dao:compileTestJava` passes.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -20,6 +20,7 @@ import io.ebean.config.ServerConfig;
 import com.linkedin.metadata.query.LocalRelationshipCriterion;
 import com.linkedin.metadata.query.LocalRelationshipCriterionArray;
 import com.linkedin.metadata.query.LocalRelationshipFilter;
+import com.linkedin.metadata.query.LocalRelationshipValue;
 import com.linkedin.metadata.query.RelationshipDirection;
 import io.ebean.EbeanServer;
 import io.ebean.SqlQuery;
@@ -62,6 +63,8 @@ public class EbeanLocalRelationshipQueryDAO {
   private static final String IDX_DESTINATION_DELETED_TS = "idx_destination_deleted_ts";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_destination_deleted_ts) ";
   private static final String DESTINATION_FIELD =  "destination";
+  // Default UrnField.name (see UrnField.pdl); a destination entity filter joined on dt.urn carries it.
+  private static final String URN_FIELD = "urn";
   private final EbeanServer _server;
   private final MultiHopsTraversalSqlGenerator _sqlGenerator;
 
@@ -1283,6 +1286,15 @@ public class EbeanLocalRelationshipQueryDAO {
     sqlBuilder.append("SELECT rt.*");
     sqlBuilder.append(" FROM ").append(relationshipTableName).append(" rt ");
 
+    // META-24159 (keyset builder only): when the destination filter pins exactly one urn (a direct,
+    // non-negated urn EQUAL or single-value IN leaf), force idx_destination_deleted_ts. InnoDB suffixes
+    // secondary indexes with the PK id, so (destination, deleted_ts) also satisfies ORDER BY rt.id ASC
+    // without a filesort. Joins/filters are always retained; the hint fires only when the index exists.
+    if (destTableName != null && isDirectSingleUrnLeaf(destinationEntityFilter, URN_FIELD)
+        && _schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
+      sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
+    }
+
     final List<Triplet<LocalRelationshipFilter, String, String>> filters = new ArrayList<>();
 
     if (_schemaConfig == EbeanLocalDAO.SchemaConfig.NEW_SCHEMA_ONLY) {
@@ -1295,6 +1307,11 @@ public class EbeanLocalRelationshipQueryDAO {
       } else if (destinationEntityFilter != null) {
         validateEntityFilterOnlyOneUrn(destinationEntityFilter);
         filters.add(new Triplet<>(destinationEntityFilter, "rt", relationshipTableName));
+      } else if (isDirectSingleUrnLeaf(relationshipFilter, DESTINATION_FIELD)
+          && _schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
+        // No destination entity table: a direct single-urn `destination` leaf pins rt.destination,
+        // so apply the same guarded idx_destination_deleted_ts hint.
+        sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
       }
 
       if (sourceTableName != null) {
@@ -1324,6 +1341,67 @@ public class EbeanLocalRelationshipQueryDAO {
     sqlBuilder.append(" ORDER BY rt.id ASC LIMIT ").append(pageSize);
 
     return sqlBuilder.toString();
+  }
+
+  /**
+   * Whether {@code filter} is a single, direct, non-negated urn leaf pinning exactly one value
+   * ({@code urn EQUAL '<value>'} as a scalar string, or {@code urn IN ('<value>')} as a one-element
+   * array), eligible for the META-24159 {@code idx_destination_deleted_ts} hint. Any {@code AND}/
+   * {@code OR}/{@code NOT} wrapper, extra criteria, a non-urn field, a scalar {@code IN}, or a multi-
+   * value (or non-scalar {@code EQUAL}) value is ineligible. The urn field must be named
+   * {@code expectedUrnFieldName} ({@code "destination"} for the relationship filter, else {@code "urn"}).
+   */
+  private boolean isDirectSingleUrnLeaf(@Nullable final LocalRelationshipFilter filter,
+      @Nonnull final String expectedUrnFieldName) {
+    final LocalRelationshipCriterion leaf = topLevelDirectCriterion(filter);
+    if (leaf == null || !leaf.hasField() || !leaf.getField().isUrnField()) {
+      return false;
+    }
+
+    if (!expectedUrnFieldName.equals(leaf.getField().getUrnField().getName())) {
+      return false;
+    }
+
+    final LocalRelationshipValue value = leaf.getValue();
+    if (value == null) {
+      return false;
+    }
+
+    switch (leaf.getCondition()) {
+      case EQUAL:
+        return value.isString();
+      case IN:
+        return value.isArray() && value.getArray().size() == 1;
+      default:
+        return false;
+    }
+  }
+
+  /**
+   * Returns the single top-level {@link LocalRelationshipCriterion} of {@code filter}, or
+   * {@code null} when it is empty, has more than one criterion, or wraps its criterion in any logical
+   * operation ({@code AND}/{@code OR}/{@code NOT}). The logical tree is intentionally not flattened,
+   * so negated or grouped filters stay ineligible for the single-destination hint.
+   */
+  @Nullable
+  private LocalRelationshipCriterion topLevelDirectCriterion(@Nullable final LocalRelationshipFilter filter) {
+    if (filter == null) {
+      return null;
+    }
+
+    if (filter.hasLogicalExpressionCriteria() && filter.getLogicalExpressionCriteria() != null) {
+      final LogicalExpressionLocalRelationshipCriterion node = filter.getLogicalExpressionCriteria();
+      if (node.hasExpr() && node.getExpr().isCriterion()) {
+        return node.getExpr().getCriterion();
+      }
+      return null;
+    }
+
+    if (filter.hasCriteria() && filter.getCriteria().size() == 1) {
+      return filter.getCriteria().get(0);
+    }
+
+    return null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -3071,4 +3072,210 @@ public void testFindRelationshipsV4ByKeysetRejectsCriteriaField() {
...
+  // A DAO whose SchemaValidatorUtil is a mock so tests drive indexExists deterministically (the embedded
+  // relationship tables carry no idx_destination_deleted_ts); injected via reflection to avoid a
+  // test-only setter on the production class.
+  private EbeanLocalRelationshipQueryDAO daoWithMockedIndex(boolean indexPresent) {
+    SchemaValidatorUtil mockValidator = mock(SchemaValidatorUtil.class);
+    when(mockValidator.indexExists(anyString(), anyString())).thenReturn(indexPresent);
+    EbeanLocalRelationshipQueryDAO dao = new EbeanLocalRelationshipQueryDAO(_server, _eBeanDAOConfig);
+    try {
+      java.lang.reflect.Field field = EbeanLocalRelationshipQueryDAO.class.getDeclaredField("_schemaValidatorUtil");
+      field.setAccessible(true);
```
NatalliaUlashchick: Reflecting into `_schemaValidatorUtil` leaves the DAO half-mocked: `_sqlGenerator` is final and keeps the *real* validator. Harmless here, but a trap for the next test that touches multi-hop traversal. It also fails at runtime rather than compile time on rename, and breaks outright if the field is ever made `final` — which it should be, since it's only written in constructors.

Suggest a `@VisibleForTesting` package-private constructor `(EbeanServer, EBeanDAOConfig, SchemaValidatorUtil)` that wires both, and making the field final. A package-private constructor is a legitimate seam; reflection isn't the safer alternative to a setter.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -3071,4 +3072,210 @@ public void testFindRelationshipsV4ByKeysetRejectsCriteriaField() {
...
   }
 
+  // Keyset (seek) pagination: META-24159 single-destination FORCE INDEX hint.
+  // A DAO whose SchemaValidatorUtil is a mock so tests drive indexExists deterministically (the embedded
+  // relationship tables carry no idx_destination_deleted_ts); injected via reflection to avoid a
+  // test-only setter on the production class.
+  private EbeanLocalRelationshipQueryDAO daoWithMockedIndex(boolean indexPresent) {
+    SchemaValidatorUtil mockValidator = mock(SchemaValidatorUtil.class);
+    when(mockValidator.indexExists(anyString(), anyString())).thenReturn(indexPresent);
+    EbeanLocalRelationshipQueryDAO dao = new EbeanLocalRelationshipQueryDAO(_server, _eBeanDAOConfig);
```
NatalliaUlashchick: `anyString(), anyString()` means a typo in `IDX_DESTINATION_DELETED_TS` passes every test in both data providers. Worth pinning:

```java
when(mockValidator.indexExists(eq("relationship_table_name"), eq(IDX_DESTINATION_DELETED_TS)))
    .thenReturn(indexPresent);
```

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1283,6 +1286,15 @@ private String buildFindRelationshipKeysetSQL(@Nonnull final String relationship
     sqlBuilder.append("SELECT rt.*");
     sqlBuilder.append(" FROM ").append(relationshipTableName).append(" rt ");
 
+    // META-24159 (keyset builder only): when the destination filter pins exactly one urn (a direct,
+    // non-negated urn EQUAL or single-value IN leaf), force idx_destination_deleted_ts. InnoDB suffixes
+    // secondary indexes with the PK id, so (destination, deleted_ts) also satisfies ORDER BY rt.id ASC
+    // without a filesort. Joins/filters are always retained; the hint fires only when the index exists.
+    if (destTableName != null && isDirectSingleUrnLeaf(destinationEntityFilter, URN_FIELD)
```
NatalliaUlashchick: This and the `else if` ~20 lines down are the same decision, split across two sites with opposite `destTableName` polarity. They're mutually exclusive, but the reader has to derive that. Consider computing one boolean before the `FROM` append and emitting the hint once.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1295,6 +1307,11 @@ private String buildFindRelationshipKeysetSQL(@Nonnull final String relationship
       } else if (destinationEntityFilter != null) {
         validateEntityFilterOnlyOneUrn(destinationEntityFilter);
         filters.add(new Triplet<>(destinationEntityFilter, "rt", relationshipTableName));
+      } else if (isDirectSingleUrnLeaf(relationshipFilter, DESTINATION_FIELD)
```
NatalliaUlashchick: Is skipping the branch directly above (`destTableName == null && destinationEntityFilter != null`) intentional?

`validateEntityFilterOnlyOneUrn` already guarantees exactly one urn criterion there, and the predicate lands straight on `rt` with no join and no equality propagation involved — arguably the cleanest of the three candidates, and it's the one that gets no hint. If unintentional:

```java
} else if (destinationEntityFilter != null) {
  validateEntityFilterOnlyOneUrn(destinationEntityFilter);
  if (isDirectSingleUrnLeaf(destinationEntityFilter, DESTINATION_FIELD)
      && _schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
    sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
  }
  filters.add(new Triplet<>(destinationEntityFilter, "rt", relationshipTableName));
}
```

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1326,6 +1343,67 @@ private String buildFindRelationshipKeysetSQL(@Nonnull final String relationship
...
+    if (filter.hasLogicalExpressionCriteria() && filter.getLogicalExpressionCriteria() != null) {
+      final LogicalExpressionLocalRelationshipCriterion node = filter.getLogicalExpressionCriteria();
+      if (node.hasExpr() && node.getExpr().isCriterion()) {
+        return node.getExpr().getCriterion();
+      }
+      return null;
+    }
+
+    if (filter.hasCriteria() && filter.getCriteria().size() == 1) {
+      return filter.getCriteria().get(0);
```
NatalliaUlashchick: This branch is unreachable from the only call site: `buildFindRelationshipKeysetSQL` runs `normalizeLocalRelationshipFilter` first, which either converts `criteria` into `logicalExpressionCriteria` or throws when both are set. When `criteria` is empty, `size() == 1` is false anyway.

Minimal safe edit is to delete just these four lines and return `null` — no need to restructure the method.

NatalliaUlashchick (commented):
A few small follow-ups on the hint guard and its tests. Nothing blocking — the eligibility check itself looks right, and I confirmed the conservative shapes it excludes are exactly the ones that degrade.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Agreed, replaced the reflection with a constructor that takes the validator and builds the SQL generator from the same instance, so the DAO can't be left half-configured, and made the field final.

One deviation: it's public rather than package-private. The test class is in `com.linkedin.metadata.dao.localrelationship` and the DAO is in `com.linkedin.metadata.dao`, so package-private isn't visible to it. Marked `@VisibleForTesting`.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Done. There's now one boolean computed before the `FROM` append and a single place that emits the hint.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Not intentional, fixed it.
The single boolean now covers all three shapes: entity-table `urn` leaf, `rt.destination` leaf on the entity filter, or `rt.destination` leaf on the relationship filter. Added a test for the empty-entity-filter case specifically.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Removed, left a comment saying why a populated `criteria` can't reach that point.

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Done, pinned to the exact table and index.

NatalliaUlashchick (approved):
All five review comments addressed and verified against the code at `488dc2d`, not just the replies:

- `_schemaValidatorUtil` is now `final`, and the new `@VisibleForTesting` constructor wires the validator and `_sqlGenerator` from the same instance — no more half-configured DAO, no reflection in the test. `public` instead of package-private is the right call given the test's package.
- The mock is pinned to the exact table and index, so a typo in either now fails.
- The hint is one boolean and one emission site.
- The dead `hasCriteria()` branch is gone, with a comment explaining why it's unreachable.

The destination-pinning fix went further than I asked, and correctly: hoisting the check above the render branches means a non-null-but-empty destination entity filter with a pinning relationship filter now gets the hint. Since that's how reverse-lineage reads arrive, the original version would have missed its own motivating case. Nice catch, and the new test pins it.

Also confirmed independently that the guard's conservatism is right: single-value `EQUAL` and single-value `IN` produce identical plans (MySQL folds a one-element `IN` to `Item_func_eq` at parse time, so it joins the multiple-equality set just like `=`), while the multi-value `IN` shape the guard excludes is exactly the one that picks up a filesort.

One non-blocking follow-up whenever convenient: the `"single-array destination EQUAL"` data-provider row asserts `rt.destination='('urn:li:foo:1')'` as expected output. That string isn't valid SQL — `SQLStatementUtils` guards `IN`-with-scalar but not `EQUAL`-with-array. The guard correctly refuses to hint that shape, so there's no production impact; it's just that pinning the malformed string will make a future fix look like a regression. Asserting only the absence of the hint would do.

PR #638, opened 2026-08-27
perf(ebean-dao): optimize keyset scans pinned to a single urn

Ruturaj123:
## Summary

Fixes two gaps in index-hint eligibility for keyset relationship scans, both of which leave a query
on a full `PRIMARY` scan:

1. **Source-pinned queries got no hint at all.** Adds `FORCE INDEX (idx_source_deleted_ts)` for
   queries pinning `rt.source` to a single urn, mirroring what #637 did for `rt.destination`.
2. **A urn pinned inside an `AND` group was not recognised.** Eligibility required a bare urn leaf,
   so any filter carrying a second predicate was skipped. This affected the destination side too,
   so #637's hint was silently missing on a common shape.

`LineageAccessorImpl` adds a second criterion whenever the caller passes `lineageType`, and a third
for `actorName`, so a plain lineage read lands in exactly that `AND` shape. Reproduced on
`corp-ltx1`, where it runs until the connection drops:

```
GET /datasets/{key}/lineages?downstream=true&lineageType=DIRECT_COPY

HTTP 500 ... Communications link failure
Query was: SELECT rt.* FROM metadata_relationship_downstreamof rt
  WHERE rt.deleted_ts is NULL
    AND (rt.destination='urn:li:dataset:(...)' AND rt.metadata$type='COPY')
    AND rt.id > 0 AND rt.id <= 516230316
  ORDER BY rt.id ASC LIMIT 1000
```

No `FORCE INDEX`. All three uncovered shapes (`source = X`, `(destination = X AND type = Y)` and
`(source = X AND type = Y)`) were confirmed present on the long-query audit and are closed by this
change.

Screenshot 2026-08-28 at 4 48 07 PM

**Why descending an `AND` is sound.** Every conjunct must hold, so one conjunct pinning the urn pins
it for the whole expression; the rest only narrow the row set. `OR`, `NOT` and unspecified operators
are *not* descended into: a row can satisfy an `OR` through its other branch, and a negated predicate
pins nothing, so in both cases the index would no longer cover the query.

Per-leaf rules are unchanged from #637: a urn field of the expected name, `EQUAL` on a scalar or `IN`
on a one-element array. Pinning to exactly one value is what lets the composite index also satisfy
`ORDER BY rt.id ASC`, since InnoDB's id suffix is ordered only within a single leading-column value.

`idx_source_deleted_ts` already exists on the relationship tables, so no migration is added. Each
hint is emitted only when `indexExists` confirms it.

**A dropped filter fixed along the way.** When no source entity class is passed, the source entity
filter was not rendered at all. The destination has an `else if` that moves its filter onto `rt` for
that case; the source had none, so a caller filtering on a non-MG source got back rows it had asked
to exclude. `validateEntityTypeAndFilter` only rejects a blank entity type, so a non-MG type such as
`NON_MG_ASSET` passes validation and then yields a null table name, which is how the filter went
missing. Both builders now give the source the same treatment, and eligibility follows it, so the
source is pinnable through that filter as the destination already was.

The new arm is gated on non-empty criteria rather than non-null, unlike the destination.
`validateEntityFilterOnlyOneUrn` reads `criteria.get(0)` without a size check, so an empty
logical-expression filter would fail there, and an empty filter contributes nothing to the `WHERE`
clause either way. The destination arm predates this and is left alone.

**Deliberate asymmetry.** Only one `FORCE INDEX` can be emitted, so destination wins when both sides
are pinned, keeping every currently hinted query byte-identical.

**Known limitation, unchanged here.** When an entity join is present, neither side considers the
relationship filter as a pinning source. That is pre-existing and symmetric across both sides.
Production lineage reads use the keyset builder.

## Testing Done

```
./gradlew :dao-impl:ebean-dao:compileJava :dao-impl:ebean-dao:compileTestJava   BUILD SUCCESSFUL
./gradlew :dao-impl:ebean-dao:checkstyleMain :dao-impl:ebean-dao:checkstyleTest BUILD SUCCESSFUL
./gradlew :dao-impl:ebean-dao:test    2241 tests, 11 failed, 2070 skipped
```

The 11 failures are pre-existing on this machine: `EmbeddedMariaInstance` will not start, so every
DB-backed class fails at `@BeforeClass init`. Verified identical counts on unmodified
`origin/master` (2241 / 11 / 2070). The new tests could not run locally for that reason and rely on
CI.

The embedded test schema declares neither `idx_source_deleted_ts` nor `idx_destination_deleted_ts`,
so `indexExists` is false throughout the DB-backed suite and no hint is emitted there. Broadening
eligibility to `AND` groups therefore cannot change the SQL those tests assert; the hint tests drive
`indexExists` through a mocked validator instead.

Tests added or updated:

- `relationshipFilterSourceCases`: eligibility matrix over `EQUAL`, single-value `IN`, `NOT`/`OR`
  groups, multi-value `IN`, array-valued `EQUAL`, a wrong field name, and the index-missing case.
- `testKeysetSqlHintsUrnPinnedInsideAndGroup`: the production shape, on both sides.
- `testKeysetSqlHintsUrnPinnedInNestedAndGroup` and
  `testKeysetSqlNoHintWhenUrnIsOnlyPinnedInsideNestedOr`: an `AND` beneath an `AND` pins; an
  `OR` beneath an `AND` does not.
- `testKeysetSqlDestinationWinsWhenBothSidesPinned`: now asserts the destination hint is emitted for
  a both-pinned `AND` group rather than no hint at all.
- `testKeysetSqlSourceEntityFilterJoinAndHint` and
  `testKeysetSqlFallsThroughToSourceWhenDestinationIndexMissing`: join retention and hint ordering.
- `testKeysetSqlRendersAndHintsSourceEntityFilterWhenSourceTableIsAbsent` and
  `testKeysetSqlIgnoresEmptySourceEntityFilterWhenSourceTableIsAbsent`: the dropped-filter fix, and
  that an empty filter is skipped rather than validated.
- `testKeysetSqlHintsUrnPinnedDeepInsideAndChain`: a urn pinned deep in an AND chain is still pinned,
  since the walk has no depth limit.
- `testLegacySqlNoSourceHintForShapesThatDoNotPinOneValue`: the legacy source arm refuses negated and
  multi-value leaves, and still hints the eligible shape.
- Legacy-builder equivalents for the destination-first chain, the source fallback, and a non-urn
  criterion being stepped over while locating a urn leaf.

Verified on `corp-ltx1` with `EXPLAIN ANALYZE`, on the join-free shape metadata-store issues.

Un-hinted, `rt.source` pinned:

```
-> Limit: 1000 row(s)
  -> Sort: rt.id, limit input to 1000 row(s) per chunk
    -> Filter: ((rt.source = '...') and (rt.deleted_ts is null) and (rt.id > 0) and (rt.id <= ...))
      -> Intersect rows sorted by row ID
        -> Index range scan on rt using idx_source_deleted_ts over (source = '...' AND deleted_ts = NULL AND 0 < id <= ...)
```

Hinted:

```
-> Limit: 1000 row(s)
  -> Index range scan on rt using idx_source_deleted_ts over (source = '...' AND deleted_ts = NULL AND 0 < id <= ...),
     with index condition: ((rt.source = '...') and (rt.deleted_ts is null) and (rt.id > 0) and (rt.id <= ...))
```

The intersect does not emit in `rt.id` order, so `ORDER BY rt.id` adds a sort, and a sort has to read
every matching row before `LIMIT` can apply. That is the unbounded per-page work keyset pagination
exists to prevent. The hint removes both the intersect and the sort, leaving a range scan that stops
at 1000.

The `AND` shape this PR adds behaves the same way. Hinted, `metadata$type` becomes a residual filter
over the same range scan:

```
-> Limit: 1000 row(s)
  -> Filter: (rt.metadata$type = 'COPY')
    -> Index range scan on rt using idx_source_deleted_ts over (source = '...' AND deleted_ts = NULL AND 0 < id <= ...),
       with index condition: (...)
```

while un-hinted it goes back through `Intersect rows sorted by row ID` and `Sort: rt.id`.

The joined shape shows the same transition, `index_merge` + `Using filesort` becoming `range` +
`Using index condition`.

**Where the hint helps, and where it does not.** The gain depends on how the pinned urn's rows are
distributed across the id space, not on fan-out alone.

Measured on a deliberately unfavourable subject, a urn holding roughly half the table (index
estimate 41.6M rows against a table estimate of 80.0M), the un-hinted plan is slightly *faster*,
4.31ms against 5.70ms for 1000 rows. With the urn that dominant a sequential `PRIMARY` scan finds
1000 matches after reading only 1349 rows, while the hinted plan pays 1000 secondary-index entries
plus 1000 bookmark lookups back to the primary key. That is the best case for scanning and the worst
case for the hint.

The case this targets is the opposite: a urn whose rows are sparse in id order. There a `PRIMARY`
scan reads far more rows than it returns: the `filtered=1` behaviour #637 was built on, and the
intersect-plus-sort plans shown above. The evidence that this is the shape that matters in production
is that destination-pinned queries disappeared from the long-query audit once #637 reached the
fabric, and that the queries above ran until the connection dropped.

Absolute magnitude on a representative subject is not quantified.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -1447,8 +1447,13 @@ public class EbeanLocalRelationshipQueryDAO {
       return isSingleUrnLeafCriterion(expr.getCriterion(), expectedUrnFieldName);
     }
 
-    // The expr union has exactly two members, so anything that is not a criterion is a logical
-    // operation. Flattening and where-clause construction make the same assumption.
+    // An expr union with neither member set is representable, and getLogical() returns null for it,
+    // so this cannot be inferred from isCriterion() alone. Flattening skips such a node rather than
+    // failing, so the filter reaches here and must be treated as pinning nothing.
+    if (!expr.isLogical()) {
+      return false;
+    }
+
     final LogicalOperation operation = expr.getLogical();
     if (operation.getOp() != Operator.AND) {
       return false;
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -1430,8 +1430,7 @@ public class EbeanLocalRelationshipQueryDAO {
    */
   private boolean pinsUrnFieldToOneValue(@Nullable final LocalRelationshipFilter filter,
       @Nonnull final String expectedUrnFieldName) {
-    if (filter == null || !filter.hasLogicalExpressionCriteria()
-        || filter.getLogicalExpressionCriteria() == null) {
+    if (filter == null || !filter.hasLogicalExpressionCriteria()) {
       return false;
     }
     return pinsUrnFieldToOneValue(filter.getLogicalExpressionCriteria(), expectedUrnFieldName, 0);
@@ -1448,12 +1447,10 @@ public class EbeanLocalRelationshipQueryDAO {
       return isSingleUrnLeafCriterion(expr.getCriterion(), expectedUrnFieldName);
     }
 
-    if (!expr.isLogical()) {
-      return false;
-    }
-
+    // The expr union has exactly two members, so anything that is not a criterion is a logical
+    // operation. Flattening and where-clause construction make the same assumption.
     final LogicalOperation operation = expr.getLogical();
-    if (operation.getOp() != Operator.AND || operation.getExpressions() == null) {
+    if (operation.getOp() != Operator.AND) {
       return false;
     }
 
@@ -1473,9 +1470,9 @@ public class EbeanLocalRelationshipQueryDAO {
    * composite index also satisfy {@code ORDER BY rt.id ASC}, since the id suffix is ordered only within a
    * single leading-column value.
    */
-  private boolean isSingleUrnLeafCriterion(@Nullable final LocalRelationshipCriterion leaf,
+  private boolean isSingleUrnLeafCriterion(@Nonnull final LocalRelationshipCriterion leaf,
       @Nonnull final String expectedUrnFieldName) {
-    if (leaf == null || !leaf.hasField() || !leaf.getField().isUrnField()) {
+    if (!leaf.getField().isUrnField()) {
       return false;
     }
 
@@ -1484,10 +1481,6 @@ public class EbeanLocalRelationshipQueryDAO {
     }
 
     final LocalRelationshipValue value = leaf.getValue();
-    if (value == null) {
-      return false;
-    }
-
     switch (leaf.getCondition()) {
       case EQUAL:
         return value.isString();
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -63,6 +63,11 @@ public class EbeanLocalRelationshipQueryDAO {
   private static final String IDX_DESTINATION_DELETED_TS = "idx_destination_deleted_ts";
   private static final String FORCE_IDX_ON_DESTINATION = " FORCE INDEX (idx_destination_deleted_ts) ";
   private static final String DESTINATION_FIELD =  "destination";
+  private static final String IDX_SOURCE_DELETED_TS = "idx_source_deleted_ts";
+  private static final String FORCE_IDX_ON_SOURCE = " FORCE INDEX (idx_source_deleted_ts) ";
+  // The UrnField.name a source-pinning leaf carries. Deliberately separate from the public SOURCE
+  // constant above, which names a result-set column and a DataMap key rather than a filter field.
+  private static final String SOURCE_FIELD = "source";
   // Default UrnField.name (see UrnField.pdl); a destination entity filter joined on dt.urn carries it.
   private static final String URN_FIELD = "urn";
   private final EbeanServer _server;
@@ -1146,18 +1151,17 @@ public class EbeanLocalRelationshipQueryDAO {
         // non-mg entity case, applying dest filter on relationship table
         filters.add(new Triplet<>(destinationEntityFilter, "rt", relationshipTableName));
       } else if (filterHasNonEmptyCriteria(relationshipFilter)) {
-        // Apply FORCE INDEX if destination field is being filtered, and the index exists
+        // Apply FORCE INDEX if the destination or source field is being filtered, and the index exists.
+        // This branch is reached only when no destination entity join was appended, so the hint still lands
+        // directly after "FROM <table> rt" and ahead of the source join added below.
         final LocalRelationshipCriterionArray relationshipCriteria =
             flattenLogicalExpressionLocalRelationshipCriterion(relationshipFilter.getLogicalExpressionCriteria());
-        for (LocalRelationshipCriterion criterion : relationshipCriteria) {
-          LocalRelationshipCriterion.Field field = criterion.getField();
-          if (field.getUrnField() != null && DESTINATION_FIELD.equals(field.getUrnField().getName())) {
-            // Check if index exists on 'destination' before applying FORCE INDEX
-            if (_schemaValidatorUtil.indexExists(relationshipTableName, IDX_DESTINATION_DELETED_TS)) {
-              sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
-            }
-            break;
-          }
+        // Destination is checked first and short-circuits, so a destination-filtered query keeps exactly the
+        // plan it had before META-24386 added the source arm.
+        if (!appendUrnFieldIndexHint(sqlBuilder, relationshipCriteria, relationshipTableName, DESTINATION_FIELD,
+            IDX_DESTINATION_DELETED_TS, FORCE_IDX_ON_DESTINATION)) {
+          appendUrnFieldIndexHint(sqlBuilder, relationshipCriteria, relationshipTableName, SOURCE_FIELD,
+              IDX_SOURCE_DELETED_TS, FORCE_IDX_ON_SOURCE);
         }
       }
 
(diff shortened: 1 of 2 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1146,18 +1154,18 @@ public String buildFindRelationshipSQL(@Nonnull final String relationshipTableNa
...
-              sqlBuilder.append(FORCE_IDX_ON_DESTINATION);
-            }
-            break;
-          }
+        // Destination takes precedence, so a query pinning both sides is hinted on the destination.
+        // The source arm is reached when the destination is not pinned, or is pinned but its index is
+        // absent, matching the keyset builder's else-if.
+        if (!appendUrnFieldIndexHint(sqlBuilder, relationshipCriteria, relationshipTableName, DESTINATION_FIELD,
+            IDX_DESTINATION_DELETED_TS, FORCE_IDX_ON_DESTINATION)) {
+          appendUrnFieldIndexHint(sqlBuilder, relationshipCriteria, relationshipTableName, SOURCE_FIELD,
```
NatalliaUlashchick: The source arm is new, so it doesn't have to inherit the destination arm's looseness. [`appendUrnFieldIndexHint`](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L1397-L1412) matches on field name alone, so a `NOT`-wrapped or multi-value-`IN` source leaf gets forced onto `idx_source_deleted_ts` — the two shapes your own `relationshipFilterSourceCases` matrix asserts must not be hinted on the keyset path.

Can you have the new source arm use the same eligibility rule [`pinsUrnFieldToOneValue`](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L1431-L1437) applies?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1362,62 +1387,113 @@ private String buildFindRelationshipKeysetSQL(@Nonnull final String relationship
...
+    if (filter == null || !filter.hasLogicalExpressionCriteria()) {
       return false;
     }
+    return pinsUrnFieldToOneValue(filter.getLogicalExpressionCriteria(), expectedUrnFieldName, 0);
+  }
 
-    if (!expectedUrnFieldName.equals(leaf.getField().getUrnField().getName())) {
+  private boolean pinsUrnFieldToOneValue(@Nonnull final LogicalExpressionLocalRelationshipCriterion node,
+      @Nonnull final String expectedUrnFieldName, final int depth) {
+    if (depth > MAX_HINT_FILTER_DEPTH || !node.hasExpr()) {
```
NatalliaUlashchick: This cap only bounds this walk. [`buildSQLQueryFromLogicalExpression`](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SQLStatementUtils.java#L581-L583) recurses over the same tree unbounded when the WHERE clause gets built forty lines later, so anything deep enough to blow the stack here blows it there anyway. What the cap does add is a silent cliff where a valid 17-level filter quietly loses its hint.

Can you either drop it, or reword [the comment on the constant](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L73-L75) so it doesn't claim protection it isn't providing?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -1299,23 +1307,40 @@ private String buildFindRelationshipKeysetSQL(@Nonnull final String relationship
...
-        : isDirectSingleUrnLeaf(destinationEntityFilter, DESTINATION_FIELD)
-            || isDirectSingleUrnLeaf(relationshipFilter, DESTINATION_FIELD);
+        ? pinsUrnFieldToOneValue(destinationEntityFilter, URN_FIELD)
+        : pinsUrnFieldToOneValue(destinationEntityFilter, DESTINATION_FIELD)
+            || pinsUrnFieldToOneValue(relationshipFilter, DESTINATION_FIELD);
 
+    // META-24386: the mirror image on the source side. Forward-lineage reads pin rt.source, and the source
+    // index keeps them off the PRIMARY scan this hint exists to avoid.
+    //
+    // Two shapes can pin the source, not the three the destination has. When sourceTableName is null the
```
NatalliaUlashchick: Agreed on not hinting here, but the premise is a bug rather than a rule. `findRelationshipsV4ByKeyset` with a non-MG source entity type gives `sourceTableName == null` while [`validateEntityTypeAndFilter`](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L885-L898) still accepts a non-empty source filter, so [the filter is dropped](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L1360-L1366) and the query returns rows the caller asked to exclude. The destination has [an `else if`](https://github.com/linkedin/datahub-gma/blob/bc26dc1f9f45217e5546d8549059b47c3c3998c5/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L1355-L1358) that renders its filter on `rt` for exactly this case; the source has no equivalent.

Can you give the source the same treatment so the filter reaches the WHERE clause? That also makes the source pinnable in the null-table case, so the eligibility ternary just above needs to follow it.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Agreed, changed the source arm to use  `pinsUrnFieldToOneValue`  so it applies the same rule as the keyset path. Added a UT for it as well.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Makes sense, dropped it.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Done, added the  `else if`  to both builders so the filter renders against  `rt` , and updated the eligibility ternary to follow it.

NatalliaUlashchick (approved):

PR #639, opened 2026-09-01
perf(ebean-dao): read a single-page relationship result in one statement

Ruturaj123:
## Summary

Reads a relationship result that fits inside one page with a single statement instead of paging it.

Keyset pagination (META-24161) replaced an unpaged read costing **one** statement with a paged read
costing **three** on the first page: scan start, current rows, and rows deleted since scan start.
Most reads on this table return far less than a page, so nearly every caller now pays three reads for
what used to take one. That is part of the latency regression in META-24388.

The first page is now read with one statement. A short page proves the whole result was returned, so
the other two are unnecessary: the scan-start bound only excludes rows inserted mid-scan, and there
is no second page to stay consistent with; the deleted-since-scan-start query only reconciles a soft
delete landing between two statements, and one statement reads one snapshot. The result matches what
the unpaged read produced before pagination.

A full page is treated as **inconclusive**, not complete. The row count cannot distinguish a result
that is exactly one page long from a longer one, and reading `pageSize + 1` would exceed the page
size bound the SQL builder enforces. Guessing wrong would silently drop rows, so that case pages as
before, costing one extra read on the first page only.

Only the first page is affected. Calls that pass a cursor are untouched.

## Why this is worth it

Counting rows per destination inside an id slice understates the total, because a destination's rows
are spread fairly evenly across the id range rather than clustered. True totals for the largest
destinations in a slice were 976, 990, 1027 and 1057 rows, against roughly four rows apiece inside a
0.4% slice, so a slice count near four corresponds to about a full page overall. That is the
threshold below.

| bucket | destinations | rows |
|---|---|---|
| likely fits one page | **62,915** (92.7%) | 74,531 |
| likely larger than one page | 4,922 (7.3%) | 233,823 |

At that hit rate roughly 1.8 of the 3 statements are saved per call. The 7.3% that fall back are the
row-heavy destinations, where one extra read against a scan that always spanned several pages is
marginal. The supernode in this table, at tens of millions of rows, stays on the paged path.

Two caveats. The threshold is derived from the even-spread observation rather than measured per
destination, so it is an estimate. And this counts destinations, not requests: if traffic
concentrates on larger datasets the hit rate is lower. The fast path is fail-safe either way, since a
wrong guess falls back and costs one extra read.

<details>
<summary>Queries behind these numbers (run on prod-shaped <code>metagalaxy</code>)</summary>

```sql
-- Bucket destinations by slice count. A slice count near 4 corresponds to about
-- one page overall; see the true-totals query below for where that comes from.
SELECT CASE WHEN c <= 4 THEN 'likely <=1000 total' ELSE 'likely >1000 total' END AS bucket,
       COUNT(*) AS destinations,
       SUM(c)   AS rows_in_bucket
FROM (
  SELECT destination, COUNT(*) AS c
  FROM metadata_relationship_downstreamof
  WHERE deleted_ts IS NULL
    AND id BETWEEN 518000000 AND 520054379   -- ~2M ids, about 0.4% of the table
  GROUP BY destination
) t
GROUP BY bucket;

-- True totals, no id filter, for the largest destinations in a slice. Forcing the
-- index keeps this to a few range scans. Returned 976, 990, 1027 and 1057.
SELECT destination, COUNT(*) AS total_rows
FROM metadata_relationship_downstreamof FORCE INDEX (idx_destination_deleted_ts)
WHERE deleted_ts IS NULL
  AND destination IN ('<urn>', '<urn>', '<urn>', '<urn>')
GROUP BY destination;
```

Pick the urns for the second query from the first slice, largest first. Excluding the supernode
keeps the query under the cost guard.

</details>

## Testing Done

```
./gradlew :dao-impl:ebean-dao:compileJava :dao-impl:ebean-dao:compileTestJava   BUILD SUCCESSFUL
./gradlew :dao-impl:ebean-dao:checkstyleMain :dao-impl:ebean-dao:checkstyleTest BUILD SUCCESSFUL
```

DB-backed tests cannot run on this machine (`EmbeddedMariaInstance` does not start, so those classes
fail at `@BeforeClass init` on master too) and rely on CI.

Tests added:

- `testFindRelationshipsByKeysetReadsShortResultInOneStatement`: counts statements rather than rows,
  since the saving is the point and a row-count assertion would not notice it.
- `testFindRelationshipsByKeysetFullFirstPageStillPages`: a full first page pages normally and still
  reports the scan's `maxId`.
- `testFindRelationshipsByKeysetMultiPageStillCompleteWithFastPath`: a multi-page result is returned
  in full, so the added read cannot drop rows on the path it does not short-circuit.

One existing test changed.
`testFindRelationshipsByKeysetDedupsRowDeletedBetweenCurrentAndDeletedQueries` used a page size of 10
on a single-row table, which now takes the single-statement path where the race it covers cannot
occur. Its page size is now 1 so it still exercises the two-statement path it was written for;
assertions are unchanged.

## Notes for reviewers

`getMaxId()` is reported differently on the fast path: the paged path reports the largest id in the
table when paging started, the single-statement path reports the largest id it returned. The value is
unused by any consumer in this repo, metadata-store or metadata-graph-query, and
`RelationshipKeysetPage` only constrains it when a cursor is present, which the fast path never
produces.

The single statement is the ordinary current-rows query, so it keeps whatever index hint the builder
would emit for a paged read, including `FORCE INDEX (idx_destination_deleted_ts)` from #637.

## Checklist

### DIFF
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -306,8 +306,10 @@ public class EbeanLocalRelationshipQueryDAO {
    * LocalRelationshipFilter, int, int)} that walks the matching set in bounded pages. Ranked/
    * non-current pagination is unsupported because it could not bound the per-page DB work.
    *
-   * <p>First page: pass {@code cursor = null}; the DAO captures {@code maxId}, the largest
-   * relationship row id when paging starts ({@code COALESCE(MAX(id), 0)}), and returns rows with
+   * <p>First page: pass {@code cursor = null}. A result that fits inside one page is read with a
+   * single statement and reports {@code maxId} as the largest id it returned. Otherwise the DAO
+   * captures {@code maxId}, the largest relationship row id when paging starts
+   * ({@code COALESCE(MAX(id), 0)}), and returns rows with
    * {@code 0 < rt.id <= maxId} ordered by id, up to {@code pageSize}. Later inserts get larger ids
    * and are excluded, keeping the scan finite. Continuation: pass the next cursor from the previous
    * {@link RelationshipKeysetPage}. The cursor also carries a database scan-start timestamp. Later
@@ -372,8 +374,10 @@ public class EbeanLocalRelationshipQueryDAO {
    * Validates V4 logical-expression filters and the {@code wrapOptions} contract. Ranked/non-current
    * pagination is unsupported because it could not bound the per-page DB work.
    *
-   * <p>First page: pass {@code cursor = null}; the DAO captures {@code maxId}, the largest
-   * relationship row id when paging starts ({@code COALESCE(MAX(id), 0)}), and returns rows with
+   * <p>First page: pass {@code cursor = null}. A result that fits inside one page is read with a
+   * single statement and reports {@code maxId} as the largest id it returned. Otherwise the DAO
+   * captures {@code maxId}, the largest relationship row id when paging starts
+   * ({@code COALESCE(MAX(id), 0)}), and returns rows with
    * {@code 0 < rt.id <= maxId} ordered by id, up to {@code pageSize}. Continuation: pass the next
    * cursor from the previous {@link RelationshipKeysetPage}. The cursor also carries a database
    * scan-start timestamp. Later pages include rows that were current at scan start even if they were
@@ -586,9 +590,11 @@ public class EbeanLocalRelationshipQueryDAO {
    * returned before keyset pagination was introduced.</p>
    *
    * <p>A full page is treated as inconclusive rather than complete. The row count alone cannot
-   * distinguish a result that is exactly one page long from one that is longer, and reading
-   * {@code pageSize + 1} rows to tell them apart would exceed the page-size bound the SQL builder
-   * enforces. Guessing wrong here would silently drop rows, so the ambiguous case pages instead.</p>
+   * distinguish a result that is exactly one page long from one that is longer. Reading
+   * {@code pageSize + 1} rows would tell them apart, but only below the maximum page size, since at
+   * the maximum it exceeds the bound the SQL builder enforces. Rather than have the fast path work
+   * for some page sizes and not others, the ambiguous case pages instead. Guessing wrong here would
+   * silently drop rows, and an exactly-full page is the rare case.</p>
    *
    * @return the complete result, or {@code null} when it did not fit in one page
    */
--- a/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
+++ b/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
@@ -453,6 +453,20 @@ public class EbeanLocalRelationshipQueryDAO {
           "Keyset pagination is only supported in NEW_SCHEMA_ONLY mode; OLD_SCHEMA_ONLY and DUAL_SCHEMA "
               + "(deprecated) are rejected.");
     }
+
+    // META-24388: most reads return far less than one page, and paging them costs three statements
+    // (scan start, current rows, rows deleted since scan start) where an unpaged read cost one.
+    // Probe with a single statement first. A short page proves the whole result was returned, so
+    // there is no second page to stay consistent with and the other two statements are unnecessary.
+    // A full page is inconclusive (there may or may not be more), so fall through and page properly.
+    if (cursor == null) {
+      final List<SqlRow> singleStatementRows = readWholeResultInOneStatement(relationshipTableName,
+          relationshipFilter, sourceTableName, sourceEntityFilter, destTableName, destinationEntityFilter, pageSize);
+      if (singleStatementRows != null) {
+        return new KeysetScanResult(singleStatementRows, largestIdOrZero(singleStatementRows), null);
+      }
+    }
+
     final long lastId;
     final long maxId;
     final String scanStartTime;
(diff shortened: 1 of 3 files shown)


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -560,6 +574,43 @@ protected void afterKeysetCurrentRowsFetched(@Nonnull String relationshipTableNa
...
+   * <p>Bounding by {@code Long.MAX_VALUE} rather than the scan's real {@code maxId} is what avoids
+   * the scan-start query: that bound exists only to keep multiple pages consistent with each other by
+   * excluding rows inserted mid-scan, and a single statement has no second page to be consistent
+   * with. The companion query for rows soft-deleted since scan start is unnecessary for the same
+   * reason, since one statement reads one snapshot. The result is the same shape the unpaged read
+   * returned before keyset pagination was introduced.</p>
+   *
+   * <p>A full page is treated as inconclusive rather than complete. The row count alone cannot
+   * distinguish a result that is exactly one page long from one that is longer, and reading
+   * {@code pageSize + 1} rows to tell them apart would exceed the page-size bound the SQL builder
```
NatalliaUlashchick: Only true when `pageSize` is 1000 — below that, `pageSize + 1` is within the builder's bound. Could you please fix the comment?

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
```diff
@@ -2694,8 +2784,10 @@ protected void afterKeysetCurrentRowsFetched(String relationshipTableName,
       }
     };
 
+    // Page size 1 rather than something larger: this test needs the two-statement path, and a
+    // result that fits inside one page is read in a single statement where no such race exists.
     RelationshipKeysetPage<ReportsTo> page = racyQueryDAO.findRelationshipsByKeyset(
-        null, emptyFilter(), null, emptyFilter(), ReportsTo.class, outgoingEmptyFilter(), 10, null);
+        null, emptyFilter(), null, emptyFilter(), ReportsTo.class, outgoingEmptyFilter(), 1, null);
```
NatalliaUlashchick, line 2770: [`...DoesNotDuplicateRelationshipRewrittenImmediatelyBeforeScan`](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java#L2799-L2820) just below needs the same fix. It drains with page size 10 against a single live row, so it takes [the fast path](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L462-L468) and Query B never runs. It still passes, but only because `deleted_ts IS NULL` already excludes the old row.

Nothing else covers this boundary — the [other merge tests](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java#L2871-L2934) all soft-delete *after* scan start. Can you put it back on the two-query path? Otherwise the `NOW(6)` fix from #636 loses its guard.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
```diff
@@ -453,6 +453,20 @@ private KeysetScanResult findRelationshipsByKeysetCore(@Nonnull final String rel
...
+    // META-24388: most reads return far less than one page, and paging them costs three statements
+    // (scan start, current rows, rows deleted since scan start) where an unpaged read cost one.
+    // Probe with a single statement first. A short page proves the whole result was returned, so
+    // there is no second page to stay consistent with and the other two statements are unnecessary.
+    // A full page is inconclusive (there may or may not be more), so fall through and page properly.
+    if (cursor == null) {
+      final List<SqlRow> singleStatementRows = readWholeResultInOneStatement(relationshipTableName,
+          relationshipFilter, sourceTableName, sourceEntityFilter, destTableName, destinationEntityFilter, pageSize);
+      if (singleStatementRows != null) {
+        return new KeysetScanResult(singleStatementRows, largestIdOrZero(singleStatementRows), null);
```
NatalliaUlashchick, line 470: This reports a different `maxId` than the paged path, which you documented in [`RelationshipKeysetPage`](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/RelationshipKeysetPage.java#L34-L39) — but [`findRelationshipsByKeyset`](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L309-L311) and [`findRelationshipsV4ByKeyset`](https://github.com/linkedin/datahub-gma/blob/52adf99577aa7233ef22488a0814750ea3f3f731/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java#L375-L377) still say the first page captures it as `COALESCE(MAX(id), 0)`. Mind updating those two javadocs as well?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Updated

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalRelationshipQueryDAO.java
Ruturaj123: Updated

dao-impl/ebean-dao/src/test/java/com/linkedin/metadata/dao/localrelationship/EbeanLocalRelationshipQueryDAOTest.java
Ruturaj123: Updated

NatalliaUlashchick (approved):

PR #622, opened 2026-05-11
perf(ebean-dao): collapse N per-aspect SQL queries into 1 in batchGetUnion

rakhiagr:
## Summary
- **Collapse N per-aspect SQL queries into 1 in `batchGetUnion()`** — previously generated 1 SQL per aspect class (e.g., 73 queries for a Dataset get with all aspects). Now issues a single SELECT with all requested `a_*` columns.
- **URN chunking at MAX_URNS_PER_QUERY=100** to keep IN clause safe for MySQL. Old implicit limit was 50 URNs/IN clause (keysCount=50, 1 aspect per key).
- **gma_deleted filtering moves from SQL to Java** — no `JSON_EXTRACT` in the query. Soft-deleted aspects are returned with their marker; callers (`EbeanLocalDAO.toRecordTemplate`) filter them out via `isSoftDeletedAspect()`.

### Changes

  | File | Change |
  |------|--------|
  | `SQLStatementUtils.java` | New `createMultiAspectReadSql()` — single SELECT with explicit aspect columns, entity-type validation guard |
  | `EBeanDAOUtils.java` | New `readMultiAspectSqlRows()` — parses multi-column rows, delegates to existing `readSqlRow()` |
  | `EbeanLocalAccess.java` | Rewrote `batchGetUnion()` to collect all columns + URNs, issue chunked queries (URN-count based) |
  | `EbeanLocalDAO.java` | Added `position>0` short-circuit for NEW_SCHEMA_ONLY/DUAL_SCHEMA to avoid duplicate results from outer pagination
  loop |
  | `IEbeanLocalAccess.java` | Updated Javadoc to describe new behavior |

  ### Edge Cases Handled

  - `a_urn` column excluded (not an aspect)
  - Column-doesn't-exist check per aspect (`validator.columnExists`)
  - NULL aspect columns (never written) — skipped in parser
  - Soft-deleted aspects (`gma_deleted`) — returned as markers, filtered by callers
  - Asset-level deletion (`deleted_ts`) — correctly applied per row
  - Empty URN set after column filtering — returns empty list
  - Mixed entity types in URN set — throws IllegalArgumentException
  - Pagination duplicates — `position>0` returns empty for NEW_SCHEMA_ONLY
  - Large URN batches — chunked into 100-URN queries

 ### Motivation

[...]

PR #640, opened 2026-09-10
test(ebean-dao): baseline benchmark for pre-#622 multi-aspect read [META-24100]

darshilbhayani92:
## Summary

Adds a standalone benchmark that measures the current cost of a multi-aspect entity read on the existing code: for an entity with many JSON aspect columns, the read issues one SELECT per aspect column, so a 73-aspect entity costs 73 round-trips per logical read. This PR captures that baseline (DB SELECT count + latency) as a reproducible, gated test.

## What this PR adds

| File | Change |
|------|--------|
| `dao-impl/ebean-dao/src/test/java/.../CurrentSchemaReadBenchmarkTest.java` | Gated baseline benchmark (`-Dgma.benchmark=true`), skipped in the normal suite/CI |
| `build.gradle` | Guarded systemProperty passthrough for `gma.benchmark` (no-op unless the flag is set) |
| `dao-impl/ebean-dao/docs/current-master-read-benchmark.md` | Results + reproducible run steps |

## Methodology

Provisions a synthetic entity table (`metadata_entity_foo`) with 73 JSON aspect columns and 50 seeded URN rows, then issues the current read shape: one SELECT per aspect column with the `JSON_EXTRACT(col,'$.gma_deleted') IS NULL` soft-delete filter => 73 round-trips per logical read.

- DB SELECT count — measured via MariaDB `Com_select` session-status delta (connection pinned in a txn).
- Latency — p50 / p90 / max over 200 iterations (20 warmup).

## Results (EmbeddedMariaDB, 50 URNs x 73 aspects = 3650 keys)

| Path | DB SELECTs / read | p50 | p90 | max |
|------|-------------------|-----|-----|-----|
| current | 73 | 28.671 ms | 32.456 ms | 42.806 ms |

Absolute latency is not production-representative (in-process DB, no network); the meaningful takeaways are the per-read query count (73 SELECTs for a 73-aspect entity) and its latency profile.

## How to run

```bash
export JAVA_HOME=$(/usr/libexec/java_home -v 11)   # Gradle 6.9.4 needs Java 11
./gradlew :dao-impl:ebean-dao:test --tests '*CurrentSchemaReadBenchmarkTest*' -Dgma.benchmark=true
```

Apple Silicon: `brew install mariadb` and enable the 3 documented M-series lines in `EmbeddedMariaInstance.java` (points MariaDB4j at the Homebrew install; local-only, not committed) — the bundled x86_64 MariaDB 10.2.11 needs OpenSSL 1.0.

## Testing done

- Ran locally (numbers above); reproduced across runs.
- Benchmark is gated off by default (`-Dgma.benchmark` unset => skipped), so CI is unaffected.


ZihanLi58 (commented):
Since you already do benchmark and have a mock table, can you also add the number for the benchmark result of the multiple aspect read? And I want to see the improvement for get, batchGet and filter

darshilbhayani92:
> can you also add the number for the benchmark result of the multiple aspect read? And I want to see the improvement for get, batchGet and filter

I am thinking to break down the benchmarking and PR changes in following way 
current benchmarking with multiple aspect read with separate calls (#640)
Interface changes to support the dual read (#641)
core logic to add the multiple read support 
final benchmark with multiple aspect (validate with baseline)

so, in this case, this PR will not have multiple aspect read support with a single call.

ZihanLi58 (approved):

PR #641, opened 2026-09-11
perf(ebean-dao): add multi-aspect batch read API surface + routing

darshilbhayani92:
## Summary

Introduces the **multi-aspect batch read API surface and strategy routing** for the read-latency optimization (META-24100), as a **pure no-op**: the default read path is `PER_ASPECT` and is completely unchanged (zero runtime impact). The optimized read is declared but intentionally **not implemented** here — it throws `UnsupportedOperationException` until the follow-up PR fills it in.

This is the **base PR** of a stacked series:
- **#641 (this PR)** — API surface + routing skeleton + stub tests
- **#642** — multi-aspect `batchGet` implementation + tests (stacked on this branch)
- **PR 3 (follow-up)** — multi-aspect **`filter`** (`list` / `listUrns`) support, reusing the same `readSqlRowForAspect` per-aspect row mapping introduced here. Kept as a separate PR so each read path (`batchGet`, then `filter`) can roll out and be benchmarked independently.

## What's included

| File | Change |
|---|---|
| `IEbeanLocalAccess.java` | default `batchGetUnionMultiAspect(...)` — **throws** (optional capability contract) |
| `EbeanLocalAccess.java` | `batchGetUnionMultiAspect(...)` override — **stubbed (throws)** |
| `InstrumentedEbeanLocalAccess.java`, `UsageTrackingEbeanLocalAccess.java` | decorator overrides — **stubbed (throws)**, declared for a consistent surface across the access layer |
| `SQLStatementUtils.java` | `createMultiAspectReadSql(...)` — signature + javadoc, **stubbed (throws)** |
| `EBeanDAOUtils.java` | `readSqlRowForAspect(...)` — public per-aspect row-mapping wrapper over private `readSqlRow` (the one real, reachable helper); reused by both the `batchGet` (#642) and `filter` (PR 3) multi-aspect paths |
| `EbeanLocalDAO.java` | `AspectReadStrategy` enum (`PER_ASPECT` default / `MULTI_ASPECT` / `DUAL`) + getter/setter, and the `newSchemaBatchGet` **routing seam** that switches on the strategy |

## Behavior

- **Default (`PER_ASPECT`) is unchanged** — `newSchemaBatchGet` delegates to the existing `batchGetUnion` per-aspect path. The only behavioral change vs `master` is a pure extraction of the existing call into the `newSchemaBatchGet` seam.
- `MULTI_ASPECT` / `DUAL` invoke the stubbed read, so they **throw `UnsupportedOperationException`** until #642 lands. Nothing enables them by default.

## Tests

Stub-level tests that assert the surface is declared but not yet implemented (they are updated to real assertions in #642):

- `SQLStatementUtilsTest.testCreateMultiAspectReadSql` — expects `UnsupportedOperationException`, using **real production dataset URNs** (from the PR #640 read benchmark) to mirror real-world URN shapes.
- `EbeanLocalAccessTest.testBatchGetUnionMultiAspectMatchesPerAspect` — expects `UnsupportedOperationException` from the stubbed access-layer read.

Existing `ebean-dao` suite stays green (the only local failure is the pre-existing macOS-only `testParseLocalRelationshipField` flake).

## Follow-up

- **#642** (stacked on this branch) implements `createMultiAspectReadSql`, the real `batchGetUnionMultiAspect` (group keys by table → one bundled read per table → per-`(urn, aspect)` mapping with soft-delete + missing-column parity), the decorator delegation, and upgrades the two stub tests to full assertions.
- **PR 3** adds multi-aspect **filter** (`list` / `listUrns`) support — the other major single-aspect-read hotspot called out in review — reusing the `readSqlRowForAspect` mapping added here. Tracked by the TODO on `readSqlRowForAspect`.


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
```diff
@@ -1477,6 +1506,29 @@ List<EbeanMetadataAspect> batchGetHelper(@Nonnull List<AspectKey<URN, ? extends
...
+   * Executes a new-schema batch read using the configured {@link AspectReadStrategy}: per-aspect
+   * (one SELECT per aspect), multi-aspect (one bundled SELECT per table), or dual (run both and compare,
+   * returning the per-aspect result as the safe source of truth).
+   */
+  private List<EbeanMetadataAspect> newSchemaBatchGet(@Nonnull List<AspectKey<URN, ? extends RecordTemplate>> keys,
+      int keysCount, int position) {
+    switch (_aspectReadStrategy) {
+      case MULTI_ASPECT:
+        return _localAccess.batchGetUnionMultiAspect(keys, keysCount, position, false, false);
+      case DUAL:
```
ZihanLi58, line 1519: should we add todo to remove dual and per aspect after we finish roll out? for code clean up purpose.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
```diff
@@ -363,6 +363,20 @@ static EntityDeletionInfo toEntityDeletionInfo(@Nonnull SqlRow row, @Nonnull Str
         .build();
   }
 
+  /**
```
ZihanLi58, line 366: Can we also add support for filter? which is also the major use case we are affected by the single aspect read.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalDAO.java
darshilbhayani92: Good call — added a TODO on `newSchemaBatchGet`'s javadoc to remove the PER_ASPECT and DUAL strategies (and the `AspectReadStrategy` switch) once MULTI_ASPECT is fully rolled out and validated, so multi-aspect becomes the only read path. Done in 9e8df276f6866d0e6fe60e114cde9afaca65990f.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/EBeanDAOUtils.java
darshilbhayani92: Agreed that filter (`listUrns`/`list` with filters) is the other major single-aspect-read hotspot. It's a separate read path from batchGet, so to keep this stack focused and let each optimization roll out and be benchmarked independently, I'll do it as a dedicated follow-up PR that reuses this same `readSqlRowForAspect` multi-aspect mapping. Added a TODO here pointing at that follow-up (9e8df276f6866d0e6fe60e114cde9afaca65990f).

darshilbhayani92: On the estimate: filter (`list` / `listUrns`) support is planned as the next stacked PR — **PR 3**, opened once #642 (multi-aspect `batchGet`) merges. It reuses the `readSqlRowForAspect` per-aspect mapping added in this PR, so it's a contained change (SQL builder + the access-layer filter path + tests). Keeping it separate lets `batchGet` and `filter` each roll out and be benchmarked independently. The TODO on `readSqlRowForAspect` tracks it.

ZihanLi58 (approved):

ZihanLi58 (approved):

PR #642, opened 2026-09-11
perf(ebean-dao): implement config-controlled multi-aspect batch read

darshilbhayani92:
## Summary

Implements the **config-controlled multi-aspect batch read** on top of the API surface + routing introduced in the base PR (#641). This collapses N per-aspect `SELECT`s into a single bundled `SELECT` of all requested aspect columns per entity table — the read-latency optimization for META-24100.

**Stacked on #641** (base branch = `dbhayani/multi-aspect-read-primitive`). It only makes the `MULTI_ASPECT` / `DUAL` paths functional; the default strategy is still `PER_ASPECT`, so **behavior is unchanged unless explicitly opted in**.

> Review #641 first — this PR's diff is only the implementation delta on top of it.

## What's included

| File | Change |
|---|---|
| `SQLStatementUtils.java` | `createMultiAspectReadSql(...)` — build one bundled `SELECT` of all aspect columns per table; urns escaped via `escapeReservedCharInUrn`; row-level `deleted_ts IS NULL` filter; also selects `deleted_ts` when `includeSoftDeleted` |
| `EbeanLocalAccess.java` | `batchGetUnionMultiAspect(...)` — group keys by entity table, run **one query per table**, then **row-driven mapping** back to aspects (see below) |
| `InstrumentedEbeanLocalAccess.java`, `UsageTrackingEbeanLocalAccess.java` | implement the stubbed decorator overrides to **delegate** the new read (metrics instrumentation + usage emission), required now that the routing seam actually invokes it |

## Mapping semantics (row-driven, matches the per-aspect path)

Rather than reverse-looking-up rows by the requested urn string, the mapping iterates the **returned rows** (like the per-aspect path) and emits one aspect per requested `(urn, aspect)` pair:

- **Case-insensitive urn matching** — requested pairs are keyed on a lower-cased urn (`matchKey`), so an aspect is not silently dropped when the stored urn differs from the requested urn by canonical/case. The returned aspect's urn is derived from the row itself, matching per-aspect output.
- **Duplicate keys collapse** — requested pairs live in a `Set` (`requestedPairs`); `requestedPairs.remove(...)` both restricts output to requested pairs and deduplicates repeated `(urn, aspect)` keys, matching the per-aspect `Set<Urn>` grouping.
- **Skip semantics preserved** — missing column → skip (single `validator.columnExists` check during key collection), no row / null cell → skip.

## Soft-delete semantics

A bundled query can't `AND` per-column JSON soft-delete checks without dropping the whole row when any single aspect is soft-deleted. So soft-delete is handled in two layers:
- **row level** in SQL (`deleted_ts IS NULL`, dropped when `includeSoftDeleted`), and
- **per-cell** during mapping via `EBeanDAOUtils.isSoftDeletedAspect`.

This preserves parity with the per-aspect path.

## Validation in prod without member impact (`DUAL`)

The `DUAL` strategy (in #641's `newSchemaBatchGet`) reads **both** paths, compares them via `EBeanDAOUtils.compareResults` (order-insensitive; logs a **WARN** on mismatch, never throws), and **returns the per-aspect (primary) result only**. So multi-aspect can be shadow-validated in production with zero member-facing risk before flipping to `MULTI_ASPECT`. Comparison is by aspect identity `(urn, aspect, version)`, which catches the real read-path risks (missing / extra / duplicate aspects, count and soft-delete-filtering differences).

## Tests

The two stub tests from #641 are upgraded to full assertions, plus new coverage for the review feedback:

- `SQLStatementUtilsTest.testCreateMultiAspectReadSql` — asserts the exact bundled SQL for `includeSoftDeleted` false/true, using **real production dataset URNs** (from the PR #640 read benchmark).
- `EbeanLocalAccessTest.testBatchGetUnionMultiAspectMatchesPerAspect` — per-aspect vs multi-aspect **result parity**, including missing-column and non-existent-urn skip.
- `EbeanLocalAccessTest.testBatchGetUnionMultiAspectExcludesSoftDeletedAspect` — **soft-delete behavior** (requested in review): a soft-deleted aspect is excluded when `includeSoftDeleted=false` and returned/flagged when `true`, with per-aspect parity.
- `EbeanLocalAccessTest.testBatchGetUnionMultiAspectDeduplicatesRepeatedKeys` — duplicate `(urn, aspect)` keys collapse to one result, matching the per-aspect path.

All pass locally (only the unrelated pre-existing macOS-only `testParseLocalRelationshipField` flake fails).

## Review feedback addressed

- Case-insensitive / canonical urn matching (row-driven mapping).
- Duplicate `(urn, aspect)` key parity (+ test).
- Removed redundant second `validator.columnExists` pass.
- Soft-delete behavior test (carried over from the reference PR #622 review).

## Notes / follow-up

- URN chunking for very large `IN (...)` lists is intentionally **not** included here and can be added in a later change.
- Multi-aspect **filter** (`list` / `listUrns`) support is a separate follow-up (PR 3), reusing `readSqlRowForAspect`.
- Rollout stays gated behind `AspectReadStrategy` (`PER_ASPECT` default → `DUAL` compare → `MULTI_ASPECT`).


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
```diff
@@ -165,7 +165,13 @@ public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnion(
   public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnionMultiAspect(
       @Nonnull List<AspectKey<URN, ? extends RecordTemplate>> keys, int keysCount, int position,
       boolean includeSoftDeleted, boolean isTestMode) {
-    throw new UnsupportedOperationException("batchGetUnionMultiAspect is not implemented yet");
+    final List<EbeanMetadataAspect> result =
```
ZihanLi58, line 168: Just curious, which of the decorator we are using now? we have instrumental one and the usage tracking ones..

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -365,7 +366,55 @@ public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnion(
...
+    }
+
+    // Map each requested (urn, aspect) key against its row, preserving per-aspect read semantics.
+    final List<EbeanMetadataAspect> results = new ArrayList<>();
+    for (int index = position; index < end; index++) {
+      final Urn entityUrn = aspectKeys.get(index).getUrn();
+      final Class<ASPECT> aspectClass = (Class<ASPECT>) aspectKeys.get(index).getAspectClass();
+      final String tableName = isTestMode ? getTestTableName(entityUrn) : getTableName(entityUrn);
+      final String columnName = getAspectColumnName(entityUrn.getEntityType(), aspectClass);
+      if (!validator.columnExists(tableName, columnName)) {
```
ZihanLi58, line 405: Is this check needed? or we can do it by check the sqlRow directly?

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -365,7 +366,55 @@ public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnion(
...
+        continue;
+      }
+      final SqlRow sqlRow = urnToRow.get(entityUrn.toString());
```
ZihanLi58, line 408: sqlRow.getString("urn") might not match entityUrn.toString(), as we have canical urn stuff, you might want to make it case insensitive.

ZihanLi58: check with @mridul111998 for more context on this

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
```diff
@@ -365,7 +366,55 @@ public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnion(
   public <ASPECT extends RecordTemplate> List<EbeanMetadataAspect> batchGetUnionMultiAspect(
       @Nonnull List<AspectKey<URN, ? extends RecordTemplate>> aspectKeys, int keysCount, int position,
       boolean includeSoftDeleted, boolean isTestMode) {
-    throw new UnsupportedOperationException("batchGetUnionMultiAspect is not implemented yet");
```
ZihanLi58, line 368: Small edge-case:  If a direct  IEbeanLocalAccess  caller passes duplicate  (urn, aspect)  keys, the multi-aspect path returns duplicates while the per-aspect path returns one result because it groups into  Set<Urn>. Not an issue for  EbeanLocalDAO.get(...)  if all callers go through the  Set<AspectKey>  API, but worth either matching the per-aspect behavior or documenting/testing the expected direct-access behavior.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
darshilbhayani92: Good catch — fixed. The old code did an exact-string `urnToRow.get(entityUrn.toString())` lookup, which would silently drop an aspect if the stored urn differed from the requested urn by canonical/case. I reworked the mapping to be **row-driven** (like the per-aspect path): it now iterates the returned rows and matches each against the requested `(urn, aspect)` pairs via a case-insensitive key (`matchKey` lower-cases the urn). The returned aspect's urn is derived from the row itself, so it matches per-aspect output exactly. (cb39303b190e170f68e8df745a681d420421a1a3)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
darshilbhayani92: Fixed as part of the same row-driven rewrite. Requested `(urn, aspect)` pairs now go into a `requestedPairs` **Set**, and the mapping loop uses `requestedPairs.remove(...)` — which both restricts output to requested pairs and collapses duplicate keys to a single result, matching the per-aspect path's `Set<Urn>` grouping. Added `testBatchGetUnionMultiAspectDeduplicatesRepeatedKeys` asserting parity when the same key is passed twice. (cb39303b190e170f68e8df745a681d420421a1a3)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
darshilbhayani92: Cleaned up — the previous version called `validator.columnExists(...)` twice (once when grouping keys, once again in the mapping pass). The rewrite computes it **once** during key collection; the mapping pass is now purely row-driven and reads columns straight off the `SqlRow` (skipping null cells via `sqlRow.get(columnName) == null`), so the redundant second check is gone. (cb39303b190e170f68e8df745a681d420421a1a3)

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
darshilbhayani92: Both decorators wrap the same `IEbeanLocalAccess`, layered by `EbeanLocalDAO`: `UsageTrackingEbeanLocalAccess` is applied when usage/read-tracking is enabled, and `InstrumentedEbeanLocalAccess` when benchmark metrics are enabled (via `setBenchmarkMetrics`). Both simply delegate `batchGetUnionMultiAspect` to the underlying access layer (usage-tracking adds the read emission, instrumented adds timing), so the new multi-aspect path is observed by whichever decorator(s) are active — no behavior change to which one is used.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
darshilbhayani92: Thanks — will loop in @mridul111998 for the canonical-urn context. As a defensive measure in the meantime, the row-to-key matching is now case-insensitive (`matchKey` lower-cases the urn), so a canonical/case difference between the stored urn and the requested urn no longer drops the aspect. Happy to tighten the normalization further if mridul flags additional canonicalization rules.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/UsageTrackingEbeanLocalAccess.java
darshilbhayani92: It's host-configured and the two compose in `EbeanLocalDAO` — base `EbeanLocalAccess`, optionally wrapped by `InstrumentedEbeanLocalAccess` (`setBenchmarkMetrics`) and/or `UsageTrackingEbeanLocalAccess` (`setUsageEmitter`, one-shot, skipped for `NoOpDaoUsageEmitter`), in either order. So at runtime it can be none / either / both.

The important guarantee is that the new read behaves identically under any of those configs. I verified `batchGetUnionMultiAspect` has exact parity with `batchGetUnion` in both decorators:
- **UsageTracking**: delegates + `emitRead("batchGetUnionMultiAspect", ...)` with the same `emissionEnabled() && !isTestMode && !DaoReadContext.isInternalRead()` gating.
- **Instrumented**: `instrument("batchGetUnionMultiAspect", ..., bucketCount(keys.size()), ...)` — same pattern.

So usage events still fire and benchmark timings are still recorded for the multi-aspect path regardless of which decorator(s) are active — no coverage gap. No code change needed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
darshilbhayani92: Good prompt — I split this into the two phases it actually spans:

1. **SQL-building phase** — the `validator.columnExists(tableName, columnName)` check is required here and stays. It runs *before* the query, so there's no `sqlRow` to inspect yet; and if a requested aspect column isn't in that table's schema, putting it in the `SELECT` fails with `Unknown column`. This also matches the per-aspect `batchGetUnion`, which does the same `columnExists` in its grouping loop.

2. **Mapping phase** — this is where your suggestion applies, and it's now done: the old redundant second `columnExists` call is gone, replaced by inspecting the row directly (`sqlRow.get(columnName) == null`). The columns iterated here are already schema-valid (they came from the phase-1 map), so a direct row check is all that's needed.

Net: one necessary schema check for building the statement + a direct `sqlRow` check for mapping — redundant double-check removed.

dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java
chmeng0: the fix looks clean.
not blocking: might worth noting that it relies on the urn column being case-insensitively collated so WHERE urn IN (...) returns case-variant rows for matchKey - mysql's default utf8mb4 collation is _ci, which is fine, if we ever need to use this with TiDB (default utf8mb4_bin, case sensitive), might need to explicitly pin ci collation

chmeng0 (approved):
LGTM

PR #643, opened 2026-09-16
fix(ebean-dao): emit the index hint for joined and entity-pinned queries, and warn when index metadata is missing

Ruturaj123:
## Summary

Two related changes to the ebean relationship DAO. Both came out of investigating slow lineage reads where `FORCE INDEX` was expected but absent.

### 1. The index hint never fired for joined or entity-pinned queries

Two independent defects kept `FORCE INDEX` off query shapes that should have had it.

**The legacy builder's hint block was unreachable behind a destination join.** `buildFindRelationshipSQL` placed the entire block in the third arm of the destination `if`/`else if` chain:

```java
if (destTableName != null)                { append dt JOIN; }
else if (destinationEntityFilter != null) { ... }
else if (filterHasNonEmptyCriteria(...))  { /* the entire hint block */ }
```

The first arm fires whenever a destination entity table is joined, so **any query joining `dt` got no hint at all**, on either side. The hint is now decided immediately after `FROM <table> rt` and before any join, which is where the clause has to land anyway.

**Both builders decided source pinning from too narrow a set of filters.** A caller can pin the source urn in three different places, and each builder only looked at some of them. In particular, a caller that joins a source entity table in order to filter on its other columns may still pin the source urn through the *relationship* filter, which is rendered against `rt` regardless of the join.

Both builders now share a single `pinsSourceToOneUrn` helper that documents the three shapes and consults the relationship filter in all of them, so the two cannot drift apart again.

Observed impact before the fix: across the full 15-day log retention for a service running this code, `idx_source_deleted_ts` appears **zero** times while `idx_destination_deleted_ts` appears 4,519 times, despite that service routinely issuing source-pinned queries of the form `INNER JOIN metadata_entity_dataset st ON st.urn=rt.source WHERE ... AND (st.urn='urn:li:dataset:(...)')`.

**Behaviour change, called out deliberately.** Queries that newly qualify will now carry a hint and may get a different plan. They are all queries that were already pinning a urn to exactly one value and were silently missing the hint, which is the case the hint exists for. The loose destination rule for relationship-filter shapes is left exactly as it was, so queries hinted today keep the same plan and byte-identical SQL. Reviewers who want to limit blast radius should focus here.

### 2. An empty index metadata result is now logged

`SchemaValidatorUtil` cannot tell "this table has no such index" from "this connection cannot read index metadata for this table". Both produce an empty `Set` from `loadIndexes`, and an empty result set from `information_schema.STATISTICS` is a *successful* query, so there is no exception to catch and nothing is logged. When it happens, `indexExists` returns false for every index on the table and each `FORCE INDEX` is dropped. Nothing fails: the query still returns correct rows, just without the index the hint was meant to pin. The only symptom is a slow query with no error and no log entry pointing at the cause.

Every existing InnoDB table reports at least a `PRIMARY` entry in `STATISTICS`, so an empty result is never legitimate for a table that exists. That invariant is what makes this safe to warn on unconditionally rather than a source of noise.

The warning is emitted from the cache loader rather than from `indexExists`, so its volume is bounded by cache misses (one per table per cache expiry) rather than by query rate. Both cache implementations, `SchemaValidatorUtil` and `SharedSchemaCache`, share the check.

An INFO-level echo of every relationship statement was considered instead and rejected: the one service that does this today spends about 10% of its entire log volume on it, and the busier consumer serves roughly 7.7x the request rate, so the same approach there would produce hundreds of millions of lines per day.

`log4j-core` is added as a **test-only** dependency so the warning can be asserted. It is already present transitively at runtime via `log4j-slf4j-impl`; this only exposes it to the test compile classpath.

## Testing Done

New `RelationshipIndexHintTest` covers both builders directly with a mocked `SchemaValidatorUtil`, so it needs no database and runs locally as well as in CI:

- Legacy builder: destination join hinted, source join hinted from the entity filter, and source join hinted from the **relationship** filter (the shape that regressed mid-review, caught by the existing `testLegacySqlSourceHintPrecedesSourceJoin`).
- Keyset builder: the same shapes, so the shared helper is exercised from both callers.
- Destination precedence on both builders.
- Pre-existing relationship-filter destination and source shapes, pinned so the refactor cannot regress them.
- Nothing-pinned and index-absent cases, confirming the gate still governs.
- Assertions check hint **position**, not just presence: appended after a join it would be syntactically invalid SQL.

New `SchemaValidatorUtilWarningTest`:

- Asserts exactly one WARN for an empty index set, naming both the table (what an operator greps for) and `FORCE INDEX` (the consequence, so the log explains why queries got slower rather than only what was missing).
- Asserts silence for a non-empty set, pinning the "not noisy" half of the contract.
- Log4j2 defaults to ERROR when no configuration file is present, which would have filtered the warning, so the test sets the logger level explicitly and restores it.

The existing `EbeanLocalRelationshipQueryDAOTest` assertions on exact SQL text, plus `SchemaValidatorUtilTest` and `SharedSchemaCacheTest`, cover the unchanged paths in CI.

## Checklist


dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/utils/SharedSchemaCache.java
```diff
@@ -185,6 +185,7 @@ private Set<String> loadIndexes(String tableName) {
     for (SqlRow row : rows) {
       indexes.add(row.getString("INDEX_NAME").toLowerCase());
     }
+    SchemaValidatorUtil.warnIfNoIndexMetadata(tableName, indexes);
```
NatalliaUlashchick: The [constructor pre-warms metadata](https://github.com/linkedin/datahub-gma/blob/f45925d69387e5309c70445ca31f60396075c929/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java#L95-L102) before [schema evolution runs](https://github.com/linkedin/datahub-gma/blob/f45925d69387e5309c70445ca31f60396075c929/dao-impl/ebean-dao/src/main/java/com/linkedin/metadata/dao/EbeanLocalAccess.java#L139-L145), so a fresh database now gets this warning during normal startup. Can you keep expected pre-migration warmup quiet while preserving the warning for unexpected metadata loss?

Ruturaj123: Done

NatalliaUlashchick (approved):

PR #644, opened 2026-09-18
test(ebean-dao): benchmark multi-aspect single-query read improvement [META-24100]

darshilbhayani92:
## What

Adds `MultiAspectReadBenchmarkTest` — an **old-vs-new latency + query-count benchmark** that quantifies how much the bundled multi-aspect read from #642 improves over master's current per-aspect read shape.

Stacked on **#642** (base branch) so the benchmark mirrors the exact multi-aspect SQL shape that `SQLStatementUtils.createMultiAspectReadSql` emits there. Companion "how much do we gain" number to the current-state baseline in #640 (which is left untouched).

## How it measures

Both read shapes run **back-to-back on the same seeded synthetic table** (`metadata_entity_foo`, 73 JSON aspect columns, one row per URN) so latencies are directly comparable:

- **current** (master today) — one `SELECT` per aspect column with `JSON_EXTRACT(col, '$.gma_deleted') IS NULL` => **73 round-trips** per logical read.
- **multi-aspect** (#642) — a single bundled `SELECT` of all aspect columns per table with a row-level `deleted_ts IS NULL` filter => **1 round-trip** per logical read.

Metrics per path: DB `Com_select` delta (connection pinned in a txn) + p50/p90/p99/max latency over 200 iterations (20 warmup). Real dataset URN shapes loaded from `benchmark-urns.txt` (editable without recompiling).

## Results (local in-process MariaDB, 73-aspect entity, 30 urns, 200 iters)

```
path         | DB SELECTs/read  | p50 ms    | p90 ms    | p99 ms    | max ms
current      | 73               | 26.985    | 30.969    | 34.415    | 37.135
multi-aspect | 1                | 1.139     | 1.216     | 1.375     | 4.180
improvement: DB SELECTs/read 73 -> 1 (73.0x fewer)   p50 latency 95.8% lower (23.70x faster)
```

| Path                | DB SELECTs / read | p50       | p90       | p99       | max       |
| ------------------- | ----------------- | --------- | --------- | --------- | --------- |
| current (master)    | **73**            | 26.985 ms | 30.969 ms | 34.415 ms | 37.135 ms |
| multi-aspect (#642) | **1**             | 1.139 ms  | 1.216 ms  | 1.375 ms  | 4.180 ms  |

**73× fewer round-trips; p50 latency 95.8% lower (~23.7× faster); p99 latency 34.415 ms → 1.375 ms (~25× faster)** for a 73-aspect entity.

> Absolute latency is not prod-representative (in-process DB, no network); the meaningful takeaway is the per-read query-count collapse (73 → 1) and the relative speedup.

## Notes

- Gated behind `-Dgma.benchmark=true` (skipped in the normal suite/CI).
- No production/source code changes — test + docs + benchmark harness plumbing only.
- Full details in `dao-impl/ebean-dao/docs/multi-aspect-read-benchmark.md`.

## How to run

```bash
export JAVA_HOME=$(/usr/libexec/java_home -v 11.0.21)
./gradlew :dao-impl:ebean-dao:test --tests '*MultiAspectReadBenchmarkTest*' -Dgma.benchmark=true
```


chmeng0 (approved):
LGTM

